PR #3, opened 2024-06-01
Use root level .gitignore file

Fuzzbender:
There is currently only one `.gitignore` on tools/app.

Git ignore file should be on root level so that it affects the whole repository.

FYI: @samilempinen @henryhaverinen


henryhaverinen (approved):
Looks good to me

Fuzzbender:
@henryhaverinen you need to "Approve" so that this can be merged.

PR #4, opened 2024-06-01
INTDEV-333 - Improve calculations generation perf as cards are added

Fuzzbender:
Improve calculation generation performance, when new cards are added to the project (via `create cards <template>`).

Change the implementation so that the file including the card specific logic programs, is written once; after all of the individual card specific calculation files are generated. Additionally, always generate the whole tree, instead of trying to generate card-by-card subtrees. The latter caused that cards sharing same subtrees would generate their logic files repeatedly (as many times as there were cards sharing the same ancestor).

This will improve the performance drastically:
- with some thousands of cards, the generation takes less than second, instead of minutes
- with 20000 cards, generation takes roughly two seconds, instead of hours
These values are from MacBook Pro M2.

There is still room to improve; we could just find a common ancestor for the cards, and if it is not "root", then generate card specific logic programs only for that subtree. If it is "root", then generate all of it.

### DIFF
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -1,6 +1,7 @@
 // node
 import { basename, join, sep } from 'node:path';
-import { mkdir, readFile, writeFile } from 'node:fs/promises';
+import { Dirent } from 'node:fs';
+import { mkdir, opendir, readFile, writeFile } from 'node:fs/promises';
 import { spawnSync } from 'node:child_process';
 
 // ismo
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (approved):
Looks good to me

tools/data-handler/src/calculate.ts
```diff
@@ -106,16 +108,27 @@ fieldtype(X, Field, "cardkeys") :- field(X, Field, _), card(Value) : field(X, Fi
...
+            const name = dirent.name;
+            const index = name.lastIndexOf(".");
+            return (index === -1) ? name : name.substring(0, index);
         }
-        promiseContainer.push(writeFile(destinationFile, cardTreeContent, { encoding: 'utf-8', flag: 'w' }));
 
-        await Promise.all(promiseContainer);
+        const files = await opendir(destinationFileBase);
+        let cardTreeContent: string = '';
+        for await (const file of files) {
```
Fuzzbender, line 128: I think the here is `await` is a leftover from some previous implementation that had parallel writing. I will remove it.

PR #5, opened 2024-06-01
INTDEV-308 Add workflow state validation for cards

Fuzzbender:
First part of validating card specific content.

Validating that each card's workflow state matches the card's workflow. 

These specific validation failures are detected:
- card has no metadata at all
- card's card-type is incorrect, or does not exist in the current project
- card's workflow is missing
- card's workflow is incorrect, or does not exist in the current project
- card' current workflow state is not correct


Fuzzbender:
When reviewing; you can mostly ignore the "test" folder changes; most of it is related to having invalid test data for unit tests.

tools/data-handler/src/validate.ts
```diff
@@ -237,6 +261,58 @@ export class Validate {
...
+                statusCode: 500,
+                message: `Card '${card.key}' has no metadata. Card object needs to be instantiated with '{metadata: true}'`
+            };
+        }
+
+        const cardType = await project.cardType(card.metadata?.cardtype);
+        if (!cardType) {
+            return {
+                statusCode: 500,
+                message: `Card '${card.key}' has invalid cardtype '${card.metadata?.cardtype}'`
```
samutoljamo-pareto, line 286: Do you think it would be beneficial to use `errorFunction` consistantly?

samutoljamo-pareto (approved):

tools/data-handler/src/validate.ts
Fuzzbender: It would.

PR #8, opened 2024-06-03
INTDEV-339 add i18 localization

samutoljamo-pareto:
Added i18n-module and translations for english


tools/app/.gitignore
Fuzzbender, line 1: Does the "app" need its own `.gitignore`, or could we have one shared one in the repository?

samutoljamo-pareto: Not really, I think it automatically created when nextjs-project was originally setup. These changes were done in the old repo so I had to squash/merge the old-repo to the this one. That's why it appeared again, but I'll remove it

PR #9, opened 2024-06-05
INTDEV-308 Validate 'customFields' in cards

Fuzzbender:
Continuation of card data validation. 

Card's custom fields are validated against types defined by "field types".

**Note** that `null` is acceptable value for all; `person` also allows for empty string. 

Additionally, fixed an issue with card creation where custom fields without values from template cards were not properly initialised. Now, they are set to `null` unless a template card has a value for them. This needs to be improved later with potential default values (from JSON schema `default`), or from custom field's default values (INTDEV-167). Potentially, also basic types might have default values apart from `null`. But `null` will suffice for now.

When "isms" project is validated with these changes the following validation errors are notified:
Screenshot 2024-06-05 at 21 39 32
When this PR is merged to "main" that project needs to be fixed.

### DIFF
--- a/tools/data-handler/src/validate.ts
+++ b/tools/data-handler/src/validate.ts
@@ -14,9 +14,16 @@ import { readJsonFile, readJsonFileSync } from './utils/json.js';
 import { requestStatus } from './interfaces/request-status-interfaces.js';
 import { pathExists } from './utils/file-utils.js';
 import { Project } from './containers/project.js';
-import { card } from './interfaces/project-interfaces.js';
+import { card, fieldtype } from './interfaces/project-interfaces.js';
+
+import * as EmailValidator from 'email-validator';
 
 const baseDir = dirname(fileURLToPath(import.meta.url));
+
+export interface LengthProvider {
+    length: number;
+}
+
 
 export class Validate {
 
@@ -41,6 +48,11 @@ export class Validate {
         this.directoryValidator = new DirectoryValidator();
         this.parentSchema = readJsonFileSync(Validate.parentSchemaFile);
         this.addChildSchemas();
+    }
+
+    // Helper to get length from types when needed.
+    private length<T extends LengthProvider>(item: T): number {
+        return item.length;
     }
 
     // Loads child schemas to validator.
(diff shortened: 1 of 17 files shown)


tools/data-handler/test/test-data/valid/decision-records/cardroot/decision_5/c/decision_6/index.json
```diff
@@ -2,5 +2,14 @@
...
-    "obsoletedBy": null
+    "obsoletedBy": null,
+    "admins": null,
+    "commit-description": null,
+    "finished": null,
+    "last-changed-date": null,
+    "last-changed-timestamp": null,
+    "number-of-commits": null,
+    "percentage-ready": null,
+    "commit-description": null,
```
Fuzzbender, line 13: "commit-description" is twice in the test card "_6"

tools/data-handler/src/validate.ts
```diff
@@ -140,6 +152,55 @@ export class Validate {
...
     }
 
+    // Validates that card's dataType can be used with JS types.
+    private validType<T>(value: T, field: string): boolean {
+        const typeOfValue = typeof value;
+
+        if (typeOfValue === 'object' && value === null) {
+            return true;
+        }
+
```
samutoljamo-pareto, line 162: I think it would be beneficial to add some kind of documentation on datatypes / accepted values on docs. I will implement validation also on the frontend so if we don't have documentation, we end up with two implementations of validation, which might conflict some time in the future.

samutoljamo-pareto:
I think it would be beneficial to add validation also to the [updateCardMetadata](https://github.com/CyberismoCom/cyberismo/blob/44596c528b3ba2b1d4ce3f15061f553d3b04876f/tools/data-handler/src/containers/project.ts#L786)-function

Fuzzbender:
> I think it would be beneficial to add validation also to the [updateCardMetadata](https://github.com/CyberismoCom/cyberismo/blob/44596c528b3ba2b1d4ce3f15061f553d3b04876f/tools/data-handler/src/containers/project.ts#L786)-function

True. I'll do that. It just basically needs to validate the card; the project structure is already valid.

tools/data-handler/src/validate.ts
Fuzzbender: As long as we have mono-repo, we should have just one validation for content. Let's share the validator then to both d-h and front end. We could re-write this validation once we start using one.

There is no current documentation on datatypes, apart from what is defined in the schemas. Needs to be done; I'll add ticket about it.

samutoljamo-pareto: Ok, I'll leave validation out of the edit metadata feature(ui) for now. If the content is validated when editing the metadata on the server side, it shouldn't produce invalid projects

Fuzzbender: Documentation ticket: https://cyberismo.atlassian.net/browse/INTDEV-351

samutoljamo-pareto (approved):

tools/data-handler/src/validate.ts
Fuzzbender: I'll create another ticket for PoC:ing if we could integrate "yup" and throw away own implementation for content validation (we are already using npm modules for file structure validation).

Fuzzbender: Yup ticket: https://cyberismo.atlassian.net/browse/INTDEV-352?atlOrigin=eyJpIjoiY2Q2YTg1YWNhNGMzNDY4YmFlOGVhMGY1NzRhMWYyNGEiLCJwIjoiaiJ9

Fuzzbender:
> I think it would be beneficial to add validation also to the [updateCardMetadata](https://github.com/CyberismoCom/cyberismo/blob/44596c528b3ba2b1d4ce3f15061f553d3b04876f/tools/data-handler/src/containers/project.ts#L786)-function

This is now fixed within the new commit.

PR #10, opened 2024-06-06
INTDEV-308 Validate enum values

Fuzzbender:
Final part of card content validations: validating that the enum values are correct.
Noticed that the interface for the custom types was incorrect; fixed it in this PR as well.

Additionally, I cleaned up the output from the validations:
Screenshot 2024-06-06 at 11 15 25


samutoljamo-pareto (approved):
Looks. Sidenote: Frontend uses PascalCase for interfaces and as far as I know, it is more common. Also noticed there is some mismatch between eslint/prettier config between dh and the next app. I'll add a low priority ticket about these

PR #13, opened 2024-06-06
INTDEV-315 Edit metadata in edit view and view metadata

samutoljamo-pareto:
### Description of changes
* Metadata is visible when viewing a card. "key" and "cardtype" are always visible and so are fields specified by "alwaysVisibleFields".
* Metadata is editable in edit view. In edit view, it is minimized by default. 
* react-hook-form module was added and used for the form.(It is a small widely used no-dependency package which removes boilerplate code around forms, no effect on the UI itself)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13632,6 +13632,21 @@
         "react": "^18.3.1"
       }
     },
+    "node_modules/react-hook-form": {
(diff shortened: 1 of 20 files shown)


Fuzzbender (approved):
Looks good to me

PR #17, opened 2024-06-08
Code scanning alert fix: Use arguments array when using 'execSync'

Fuzzbender:
Replace `execSync` with `execSyncFile` and put arguments into an array.


samutoljamo-pareto (approved):
Works on windows too

PR #19, opened 2024-06-11
INTDEV-361 Make 'createProject()' use its own sanity checks

Fuzzbender:
Currently, it is impossible to create a new project. This is due to that the shared validations for input parameters check that they can find a project folder. 

Creating project should have its own checks. This is different from other commands, which have always 'project' folder to find.

Added also unit test to cover the case.

### DIFF
--- a/tools/data-handler/test/command-handler.test.ts
+++ b/tools/data-handler/test/command-handler.test.ts
@@ -10,7 +10,7 @@ import { dirname, join } from 'node:path';
 
 // ismo
 import { CardsOptions, Cmd, Commands } from '../src/command-handler.js';
-import { copyDir, resolveTilde } from '../src/utils/file-utils.js'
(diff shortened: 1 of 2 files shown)


tools/data-handler/test/command-handler.test.ts
```diff
@@ -592,16 +592,20 @@ describe('create command', () => {
...
-        const projectDir = join(testDir, 'project-name');
         const prefix = 'proj';
         const name = 'test-project';
+        const projectDir = join(testDir, name);
         const testOptions: CardsOptions = { projectPath: projectDir };
         const result = await commandHandler.command(Cmd.create, ['project', prefix, name ], testOptions);
         try {
+            console.log(projectDir)
             await access(projectDir, fsConstants.R_OK);
+            console.log(2)
```
samutoljamo-pareto, line 603: You forgot some debug prints

samutoljamo-pareto (approved):

PR #25, opened 2024-06-13
INTDEV-361 Make 'createProject()' use its own sanity checks

Fuzzbender:
Re-do this PR again. It was lost in recent merge.


samutoljamo-pareto (approved):
Looks good to me

PR #26, opened 2024-06-13
INTDEV-364 Template is created to wrong path

Fuzzbender:
Template creation used wrong path if there were modules imported.
There was a clear programming error, as it was looking for "modules" path to exist, when it should have tried to find "modules/templates/<templateName>"

### DIFF
--- a/tools/data-handler/src/containers/template.ts
+++ b/tools/data-handler/src/containers/template.ts
@@ -152,6 +152,22 @@ export class Template extends CardContainer {
         return createdCards;
     }
 
+    // fetches path to module.
+    private moduleTemplatePath(templateName: string): string {
+        if (!pathExists(this.project.modulesFolder)) {
+            return '';
+        }
+        const files = readdirSync(this.project.modulesFolder, { withFileTypes: true });
+        const modules = files.filter(item => item.isDirectory());
+        for (const module of modules) {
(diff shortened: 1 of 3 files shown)


tools/data-handler/src/containers/template.ts
```diff
@@ -172,8 +172,10 @@ export class Template extends CardContainer {
             for (const directory of directories) {
                 const dirPath = join(directory.path, directory.name);
                 if (pathExists(resolve(templatePath))) {
-                    templatePath = join(dirPath, 'templates', templateName);
-                    break;
+                    if (pathExists(join(dirPath, 'templates', templateName))) {
```
samutoljamo-pareto, line 175: Did you mean to remove the first if-statement?

Fuzzbender: No, it is now checking if the path exists and then use the path. 
--> It is template from a module
If this if fails, then it cannot be module template, but a local one.

Fuzzbender: The previous `if` is probably not needed at all (row 174).

Fuzzbender: I think I will re-write this whole method; it is a mess.

Fuzzbender: I think it is now a lot clearer.

samutoljamo-pareto (approved):

PR #27, opened 2024-06-14
INTDEV-366 'create project' command parameters are in wrong order

Fuzzbender:
There was regression when `CommandHandler` interface was modified: `create project` parameters were flipped in the receiving end.


samutoljamo-pareto (approved):
Looks good to me

PR #28, opened 2024-06-14
INTDEV-365 Ensure that card array is returned when creating cards

Fuzzbender:
The return type of card creation was changed from `card[]` to `string[]` when `requestStatus` was dropped. 
This causes issues with emitting the changed cards for calculations. 

Switching the return value back to `card[]`.


samutoljamo-pareto (approved):
Changes look good but refactoring doCreateCards sometime wouldn't hurt. It's already over 100 lines

PR #30, opened 2024-06-14
Show fieldTypes in module details

Fuzzbender:
Currently `show module` shows cards, card-types, calculations, templates, workflows and basic info on module.
Field-types are not shown.  Add these to the command output.


samutoljamo-pareto (approved):
Looks good to me

PR #32, opened 2024-06-17
Feature/samutoljamo/intdev 309

samutoljamo-pareto:
* Replaced mui elements with Joy UI side by side
* Added delete functionality to UI

Note: In the last commit I replaced the emit-command in remove by using a direct function call. This was done because remove failed for nested structures as emit is not awaitable. I don't think emit is currently used by anything else so I moved to be after the actual removal.


tools/data-handler/package.json
```diff
@@ -39,6 +39,7 @@
     "typescript-eslint": "^7.10.0"
   },
   "dependencies": {
+    "@mui/joy": "^5.0.0-beta.36",
```
Fuzzbender, line 42: Why this is in d-h's package.json?

Shouldn't this be in app's?

Fuzzbender (approved):

tools/data-handler/src/remove.ts
```diff
@@ -45,22 +42,26 @@ export class Remove extends EventEmitter {
...
+        throw new Error(`Card '${cardKey}' not found`);
+      }
+
+      // Imported templates cannot be modified.
+      if (cardFolder.includes(`${sep}modules${sep}`)) {
+        throw new Error(`Cannot modify imported module`);
+      }
+
+      // Calculations need to be updated before card is removed.
+      const card = await Remove.project.findSpecificCard(cardKey);
```
Fuzzbender, line 56: This is fine for now; I think I will at some stage make the `handleDeleteCard` to call 
```
        await this.generateCardTreeContent(undefined);
        await this.genereteCardTree();
```
 (like in adding new cards case)
 It is almost as fast to generate the whole calculation files from scratch than to edit the existing ones (tested with ~22000 cards). Then we don't need to call `getCards(...)` in `handleDeletedCard` which is the cause of the issue.

tools/data-handler/package.json
samutoljamo-pareto: Yes it should

PR #33, opened 2024-06-17
INTDEV-324 Modules are imported with their own prefix

Fuzzbender:
Import modules by using their own project prefix.


samutoljamo-pareto (approved):

Fuzzbender:
I will not merge this; there are still some issues with projects that are used both locally and as modules.

Fuzzbender:
Now that this possibly works as it should, it would be beneficial to re-review this.
@samilempinen  @samutoljamo-pareto  @henryhaverinen 

As the change is significant towards the content repos; I'd like to have as many concerns and view points now, to avoid multiple updates to the content repos.

henryhaverinen (approved):
As far as I can understand the change, it looks great and just as we have agreed.

PR #35, opened 2024-06-19
INTDEV-360 Move prettier to top level

samutoljamo-pareto:
* Used new flatconfig
* Added config only on toplevel


samutoljamo-pareto:
I'll merge this when we have less PRs waiting so we can format the whole project easily

arime-pareto (approved):

PR #38, opened 2024-06-25
Move functionality

samutoljamo-pareto:
* Added move functionality
* Introduced redux
* Refactored code to use new hooks to make them simpler


tools/app/app/lib/hooks/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { usePathname } from 'next/navigation'
+import { useCallback, useEffect, useMemo, useRef } from 'react'
+import { findParentCard } from '../utils'
+import { useProject } from '../api'
+
+export function useCardKey() {
```
arime-pareto, line 6: In my opinion, if we are using the app router correctly, the page or layout should retrieve the card key from the url parameters and provide it to its child components. Otherwise there might be something wrong with the logical structure of the app. Also, we happen to change the routing, this hook will break.

tools/app/app/cards/layout.tsx
```diff
@@ -24,18 +25,28 @@ import {
...
+import {
+  closeNotification,
+  removeNotification,
+} from '../lib/slices/notifications'
 import { useRouter } from 'next/navigation'
-import { useTranslation } from 'react-i18next'
 
 function AppLayout({ children }: Readonly<{ children: React.ReactNode }>) {
   // Last URL parameter after /cards base is the card key
   const urlCardKey = useCardKey()
```
arime-pareto, line 40: Should be able to get card key from params.key if it is available in the route.

https://nextjs.org/docs/app/api-reference/file-conventions/layout#params-optional

tools/app/app/components/modals/DeleteModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const handleDelete = useCallback(async () => {
+    try {
+      await deleteCard()
+      dispatch(
+        addNotification({
+          message: t('deleteCardModal.success', { card: cardKey }),
+          type: 'success',
+        })
+      )
+      if (isMounted()) {
```
arime-pareto, line 80: Why do you need to check if component is mounted?

arime-pareto (changes requested):
In general looks good! There were a few parts I left some comments and questions about, please check those. :)

tools/app/app/cards/layout.tsx
arime-pareto, line 40: You can use useParams to retrieve card key from route parameters:

https://nextjs.org/docs/app/api-reference/functions/use-params

tools/app/app/components/modals/DeleteModal.tsx
samutoljamo-pareto: deleteCard-function might take a long time and onClose could be unsafe to call if the component is unmounted during that time. Might happen if connection is slow and user navigates before update is through.

tools/app/app/lib/utils.ts
```diff
@@ -233,6 +313,27 @@ export function countChildren(card: Card): number {
...
 }
 
+/**
+ * Return true if card is children of the other card
+ * Could be much more efficient
+ * @param card parent card to check if the other card is a child of
+ * @param key key of the card to check if it is a child of the parent card
+ * @returns
+ */
+export function isChildOf(card: Card, key: string): boolean {
```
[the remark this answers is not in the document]
samutoljamo-pareto: It was not intentional so I deleted it.

tools/app/app/lib/utils.ts
```diff
@@ -223,6 +215,94 @@ export function findCard(cards: Card[], key: string): Card | null {
...
+    if (cards[i].children) {
+      cards[i].children = editCard(cards[i].children || [], card)
+    }
+  }
+  return cards
+}
+
+/**
+ * Edits a card in a tree of cards
+ * This function converts CardDetails to Card before editing
```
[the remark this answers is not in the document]
samutoljamo-pareto: Good point, I rephrased it

tools/app/app/lib/utils.ts
```diff
@@ -223,6 +215,94 @@ export function findCard(cards: Card[], key: string): Card | null {
...
+        return found
+      }
+    }
+  }
+  return null
+}
+
+/**
+ * Edits a card in a tree of cards
+ * @param card
```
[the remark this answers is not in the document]
samutoljamo-pareto: I updated all documentation strings

tools/app/app/cards/layout.tsx
samutoljamo-pareto: I removed useCardKey and replaced each one with use-params

tools/app/app/components/modals/DeleteModal.tsx
arime-pareto: OK, got it! 👍

arime-pareto (approved):

PR #39, opened 2024-06-28
Fix editing

samutoljamo-pareto:
Refactored edit-component to handle null values correctly. Previously null values were given to UI components, which do not accept them. Values are stored using the correct types, but they are converted at render to a string. In case of a null value, an empty string is given.

Note: This means that if a user doesn't touch a field that is null value, it will be null even after saving. However, if user types something and then clears the input box, the value is updated to an empty string. 

This was built on top of the "move card"-feature, so I will rebase this after it has been merged to main.


samutoljamo-pareto:
I didn't rebase this, because it would've been much harder, since main contains the squashed version of #38 . Instead, I rebased this to #38 and merged other changes from main to this branch. Doesn't really matter since this will also be squashed before merging

tools/app/app/cards/[key]/edit/page.tsx
```diff
@@ -106,121 +62,117 @@ export default function Page({ params }: { params: { key: string } }) {
...
+        summary: __title__,
+        ...metadata,
       },
+      content: __content__,
     }
-  }, [card, getValues])
+  }, [preview, card])
 
-  if (!isReady) return <CircularProgress />
+  useEffect(() => {
```
Comment (arime-pareto) [inline tools/app/app/cards/[key]/edit/page.tsx:80]: This seems like a development-time debug stuff that should be removed?

tools/app/app/components/MetadataView.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const [expanded, setExpanded] = useState(initialExpanded)
+
+  const { cardType } = useCardType(metadata?.cardtype ?? null)
+
+  const { fieldTypes } = useFieldTypes()
+
+  const context = useFormContext() // must be inside a <FormProvider>
+  useEffect(() => {
+    setExpanded(initialExpanded ?? false)
+  }, [initialExpanded])
```
arime-pareto, line 43: The prop name "initialExpanded" seems to refer to something that should not change. The view is initially expanded or not and that's it. If it should be used to expand-collapse the view, I think the name should be just "expanded".

arime-pareto (changes requested):
Generally well done refactoring! Just a couple of minor notes given.

tools/app/app/components/MetadataView.tsx
samutoljamo-pareto: I removed it

tools/app/app/cards/[key]/edit/page.tsx
Comment (samutoljamo-pareto) [reply tools/app/app/cards/[key]/edit/page.tsx]: Yes, and I removed it

arime-pareto (approved):

PR #42, opened 2024-06-28
Bugfix: Workflow transition menu sticks around

samutoljamo-pareto:
StateSelector used a combination of mui and joy style.

### DIFF
--- a/tools/app/app/components/StateSelector.tsx
+++ b/tools/app/app/components/StateSelector.tsx
@@ -24,10 +24,6 @@ const StateSelector: React.FC<StateSelectorProps> = ({
   workflow,
   onTransition,
 }) => {
-  const handleMenuItemClick = (transition: WorkflowTransition) => {
-    onTransition(transition)
-  }
-
   const { t } = useTranslation()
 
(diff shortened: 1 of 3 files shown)


tools/app/app/components/StateSelector.tsx
```diff
@@ -16,64 +24,50 @@ const StateSelector: React.FC<StateSelectorProps> = ({
...
+        <Typography fontWeight={600}>
+          {t('stateSelector.status', {
+            state: currentState.name,
+          })}
+        </Typography>
+      </MenuButton>
+      <Menu>
         {availableTransitions.map((transition) => (
           <MenuItem
             key={transition.name}
```
arime-pareto, line 73: Maybe onClick could just call onTransition directly?

arime-pareto (approved):
Just a minor comment, which you can consider fixing or not. Approved anyway. Looks good! :)

tools/app/app/components/StateSelector.tsx
samutoljamo-pareto: Maybe it could even call redux directly, but I updated it to use onTransition for now

PR #45, opened 2024-07-03
Add displayName, description and category to template schema

samutoljamo-pareto:
I used "description" instead of "templateDescription". Since we are already dealing with templates, I do not think the "template" in the variable name adds any clarity.  Description, displayName and category are optional

### DIFF
--- a/tools/data-handler/src/show.ts
+++ b/tools/data-handler/src/show.ts
@@ -4,7 +4,7 @@ import mime from 'mime-types';
 
 // cyberismo
 import { attachmentPayload } from './interfaces/request-status-interfaces.js';
-import { attachmentDetails, card, cardListContainer, cardtype, fetchCardDetails, fieldtype, moduleSettings, project, resource, template, workflowMetadata } from './interfaces/project-interfaces.js';
+import { attachmentDetails, card, cardListContainer, cardtype, fetchCardDetails, fieldtype, moduleSettings, project, template, workflowMetadata } from './interfaces/project-interfaces.js';
 import { Project } from './containers/project.js';
 
 export class Show {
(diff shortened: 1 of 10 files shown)


tools/data-handler/src/show.ts
```diff
@@ -281,9 +281,14 @@ export class Show {
      * @todo: add unit tests
      * @returns all templates in a project.
      */
-    public async showTemplatesWithDetails(projectPath: string): Promise<(resource)[]> {
+    public async showTemplatesWithDetails(projectPath: string): Promise<template[]> {
         Show.project = new Project(projectPath);
-        return await Show.project.templates();
+        const promiseContainer = [];
```
arime-pareto, line 286: Suggestion: could it be cleaner to use map over templates instead a for loop and push?

tools/data-handler/src/show.ts
```diff
@@ -281,9 +281,14 @@ export class Show {
...
+        for (const template of await Show.project.templates()) {
+            promiseContainer.push(Show.project.createTemplateObjectByName(template.name).then(t => t?.show()));
+        }
+        const result = await Promise.all(promiseContainer);
+        return result.filter(item => item) as template[];
```
arime-pareto, line 291: This is fine, but a nice trick to filter undefined values from an array is to use `result.filter(Boolean)`.

arime-pareto (approved):
Looks good to me, left just a couple of improvement suggestions.

arime-pareto (approved):

tools/data-handler/src/show.ts
samutoljamo-pareto: Yes it would, but the same syntax is also used elsewhere. I replace the for loop with map in this one

PR #79, opened 2024-08-14
Macros

samutoljamo-pareto:
This pull requests allows users to define macros:
Here's an example macro: {{{macroName '{"options1": "value1"}'}}}
Macros allow users to use predefined components, which will be rendered using React components.

Macros are defined in DH/src/utils/macros. They define a name(this is the name on the UI), tagName(html-tagname used for the component) and schema(Schema is the schema of its options).

React components for the macros are defined in app/components/macros.
The index file contains the components in key/value pairs, where the key should match a name of a macro defined in DH.

Even though the macros are defined in the data-handler, they imported to the app **directly** without any apis between. This was done because so the frontend would not have to send any requests to be able to show a preview. This also means that schemas are now available on the frontend. I made a script, which generates a typescript schemas file automatically.

If you think importing directly from DH is bad, we could have another module called common, which would contain things shared by both the DH and the app.

The data flows like so:
adoc read by DH from file --> sent to app via api --> app calls `handleMacros` directly -->  now the macros have been replaced with passthrough components --> asciidoctor generates html --> html is replaced with react components --> html is rendered

The same approach unfortunately doesn't work with the current export method, because it refers directly to the source adoc. I think we need to create a new ticket for this and decide whether we:
1. Keep the current method(macros can't be used and they will be ugly)
2. Paste the adoc contents manually and handle macros in between. Macros already have a second way of handling them, which is meant for exports

Note: I won't merge this to temp-main, but if I used main, this PR would've been a mess to review before temp-main has been merged to main

### DIFF
--- a/tools/data-handler/src/utils/macros/index.ts
+++ b/tools/data-handler/src/utils/macros/index.ts
@@ -80,9 +80,18 @@ export function registerMacros(instance: typeof Handlebars, mode: Mode) {
 export function handleMacros(content: string, mode: Mode) {
   const handlebars = Handlebars.create();
   registerMacros(handlebars, mode);
-  return handlebars.compile(content, {
-    strict: true,
-  })({});
+  try {
+    return handlebars.compile(content, {
+      strict: true,
+    })({});
+  } catch (err) {
+    return handleMacroError(err, {
+      name: '',
+      tagName: '',
+      handleStatic: () => '',
+      handleInject: () => '',
+    });
+  }
 }
 
 /**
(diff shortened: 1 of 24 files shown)


samutoljamo-pareto:
I marked this as ready for review, because I think we should make a separate ticket for changes in the way the asciidoc is rendered.

tools/data-handler/src/utils/macros/index.ts
```diff
@@ -0,0 +1,169 @@
...
+    'lineNumber' in error &&
+    typeof error.lineNumber === 'number'
+  ) {
+    message += ` at line ${error.lineNumber}`;
+  }
+  return createAdmonition('WARNING', 'Macro Error', message);
+}
+
+/**
+ *
```
Fuzzbender, line 141: Empty block

Fuzzbender (approved):

tools/data-handler/src/utils/macros/index.ts
samutoljamo-pareto: Fixed

PR #80, opened 2024-08-15
Support for links

samutoljamo-pareto:
Added the following rules:
- Cannot add description if it is not allowed
- Cannot create a link if source is not defined in source card types(empty means any is allowed)
- Cannot creata a link if destination is not defined in destination card types(empty means any is allowed)

On the UI:
- Display only link types, which can be created for the specific card type
- After link type is chosen, only show destinations, which are defined in the configs
- After link type is chose, enable linkDescription box, if link type permits it

Note: Remember to change base branch to main after temp-main has been merged

### DIFF
--- a/tools/app/app/components/ContentArea.tsx
+++ b/tools/app/app/components/ContentArea.tsx
@@ -211,21 +211,21 @@ export function LinkForm({
             )}
           />
         </Stack>
-        <Controller
-          name="linkDescription"
-          control={control}
-          disabled={
-            !selectedLinkType || !selectedLinkType.enableLinkDescription
-          }
-          render={({ field }) => (
-            <Input
-              {...field}
-              color="primary"
-              startDecorator={<Edit />}
-              placeholder={t('linkForm.writeDescription')}
-            />
-          )}
-        />
+
+        {selectedLinkType && selectedLinkType.enableLinkDescription && (
+          <Controller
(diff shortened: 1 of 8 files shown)


Fuzzbender:
Looks good to me.  Can you add the "create link" functionality to CLI as well? 
If you don't want to do it, make a JIRA ticket about it - we should have 'data'-level functionality in the CLI as well.

tools/app/app/components/ContentArea.tsx
```diff
@@ -60,23 +60,96 @@ interface LinkFormSubmitData {
...
-  const { control, handleSubmit, reset } = useForm<LinkFormSubmitData>({
-    defaultValues: { linkType: '', cardKey: '', linkDescription: '' },
+export function LinkForm({
+  cards,
+  linkTypes,
+  onSubmit,
+  cardType,
+}: LinkFormProps) {
+  const { control, handleSubmit, reset, watch } = useForm<LinkFormData>({
+    defaultValues: { linkType: -1, cardKey: '', linkDescription: '' },
```
Fuzzbender, line 86: Magic values could have a local const value for easier code reading.

Fuzzbender (approved):
Add support or create ticket to add create link functionality to CLI as well.

samutoljamo-pareto:
We already have a command for that.

PR #92, opened 2024-08-21
Attachment actions

samutoljamo-pareto:
Adds the following actions:
- Download
- Open in editor

Also, the icons are now hidden by default.

If a command is defined in user preferences, it will try to launch that, otherwise it will use the default app selector of the OS.

It'll wait 1 second for the custom app to start and if it has exited with a non-zero exit code, it will use the default app selector of the OS.

There wasn't a good option for starting an app on windows. Windows seemed to put the app to background. I used a workaround, which starts a cmd prompt and it does work but the user will see a cmd window pop up and destroy itself.

Security:
Custom commands are not escaped in anyway and a shell is used so it is possible to do anything. But it doesn't matter, because these are defined by the user.

Starting the default app is safe on mac/linux, since no shell is used.
Windows version does use a shell, but since "-characters are not valid filenames on windows, it should be safe to use, because the path is read from the filesystem.

Functionality has been tested on windows/linux, but not on MacOS

The app will use a default system OS also, if the user preferences file doesn't contain the given platform. In general, it would probably be smarter to validate the user-preferences file and overwrite it using defaults if it is not valid.

### DIFF
--- a/tools/data-handler/src/show.ts
+++ b/tools/data-handler/src/show.ts
@@ -30,6 +30,10 @@ import {
   workflowMetadata,
 } from './interfaces/project-interfaces.js';
 import { Project } from './containers/project.js';
+import { spawn } from 'node:child_process';
+import { join, resolve } from 'node:path';
+import { UserPreferences } from './utils/user-preferences.js';
+import { homedir } from 'node:os';
 
 export class Show {
   static project: Project;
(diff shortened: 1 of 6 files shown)


tools/data-handler/src/utils/user-preferences.ts
```diff
@@ -57,7 +113,7 @@ export class UserPreferences {
     }
   }
 
-  public getPreferences() {
+  public getPreferences(): UserPreferencesObject {
```
Fuzzbender, line 116: Would it make sense to remove the constructor and have this method have one parameter and make it static?

So instead of

```
const prefs = new UserPreferences(path).getPreferences(); 
```

You'd have
```
const prefs = UserPreferences.getPreferences(path);
```

tools/data-handler/src/show.ts
```diff
@@ -74,6 +78,89 @@ export class Show {
...
+   * @param filename attachment filename
+   * @param waitDelay amount of time to wait for the application to open the attachment
+   */
+  public async openAttachment(
+    projectPath: string,
+    cardKey: string,
+    filename: string,
+    waitDelay: number = 1000,
+  ) {
+    const attachment = await this.getAttachment(projectPath, cardKey, filename);
```
Fuzzbender, line 114: I think we should at some stage move the `getAttachment` function to somewhere else, it is not that much related to `Show`.  And then move the `openAttachment` functionality also somewhere else, but let's keep them here for now.

Fuzzbender (approved):

tools/data-handler/src/utils/user-preferences.ts
samutoljamo-pareto: Is there a reason to prefer static classes over functions?

tools/data-handler/src/show.ts
samutoljamo-pareto: Yes I had the same thoughts but I figured the CLI command would be something like `cyberismo show attachment xxx file.png`

PR #94, opened 2024-08-23
Fix dependabot YAML-file

Fuzzbender:
The YAML has wrong number of whitespaces which makes it incorrect.


samilempinen (approved):
LGTM, provided that the builds pull through.

PR #105, opened 2024-08-28
INTDEV-393 sync scroll position to titles

samutoljamo-pareto:
I decided to try to sync the scroll position to titles, since there is no real good options for syncing between the generated html / adoc. There also still remains a small inconvenience: If you go back and fourth between edit/preview, it'll move downwards. I will create a low prio ticket for it, since I didn't see a quick solution and I'll prioritize the more important tasks.


tools/app/app/components/ContentArea.tsx
```diff
@@ -211,17 +235,27 @@ export const ContentArea: React.FC<ContentAreaProps> = ({
...
+    for (const header of headerArray) {
       const rect = header.getBoundingClientRect();
-      if (
-        rect.top >= 0 &&
-        rect.bottom <= window.innerHeight &&
-        header.id &&
-        header.id !== ''
-      ) {
-        visibleHeaderIds.push(header.id);
+      if (rect.top < (boxRect?.top || -4) + 1) {
```
Fuzzbender, line 246: Some weird magic...

Fuzzbender (approved):

tools/app/app/components/ContentArea.tsx
samutoljamo-pareto: I guess that one needs a comment :smile:

PR #106, opened 2024-08-29
Fix dependencies: Make ESLint work from all tools

Fuzzbender:
Make ESLint work from all tools (cli, d-h, app).

Screenshot 2024-08-29 at 12 44 45
Screenshot 2024-08-29 at 12 44 30


samutoljamo-pareto (approved):
I fixed remaining linter issues

PR #113, opened 2024-08-31
INTDEV-149 Improve static site generator

samilempinen:
Most of the changes are for the export functionality only, so very little chance of breaking other stuff. There is a change to `tools/app/next.config.mjs` to support the .html redirection in the app.


tools/data-handler/src/export-site.ts
```diff
@@ -58,7 +59,7 @@ export class ExportSite extends Export {
     this.pagesDir = join(this.moduleDir, 'pages');
     this.imagesDir = join(this.moduleDir, 'assets', 'images');
     this.navFile = join(this.moduleDir, 'nav.adoc');
-
+
```
Fuzzbender, line 62: Can you remove this?

tools/data-handler/src/export-site.ts
```diff
@@ -93,7 +94,7 @@ export class ExportSite extends Export {
     const playbook = {
       site: {
         title: ExportSite.project.configuration.name,
-        start_page: `cards:ROOT:${startPage}`,
+        start_page: `cards:ROOT:${startPage}`
```
Fuzzbender, line 97: I think it is better to keep the trailing commas, including on last line.

Fuzzbender (approved):
Check CodeQL comment. Otherwise looks good to me. I didn't look at the resources at all.

resources/ui-bundle/helpers/detag.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+const TAG_ALL_RX = /<[^>]+>/g
+
+module.exports = (html) => html && html.replace(TAG_ALL_RX, '')
```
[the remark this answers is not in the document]
samilempinen: This is from the default Antora UI bundle, and I don't think it's valid here.

PR #114, opened 2024-08-31
INTDEV-347 Try multiple ports upon app start

samilempinen:
Cyberismo app will now try multiple ports starting from 3000 if this port is already taken.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -1194,18 +1194,31 @@ export class Commands {
 
     // Start trying from port 3000 and increase the port number if the port is already in use
     let port = 3000;
-    let args = ['-s', `start`, `--project_path="${projectPath}"`, '--', `--port ${port}`];
-    
+    let args = [
+      '-s',
+      `start`,
+      `--project_path="${projectPath}"`,
+      '--',
+      `--port ${port}`,
+    ];
+
     while (port < 3099) {
       try {
-        console.log(`Starting Cyberismo on http://localhost:${port}. Enter CTRL+C to quit.`);
+        console.log(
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/command-handler.ts
```diff
@@ -1167,18 +1167,27 @@ export class Commands {
...
     // since current working directory changes, we need to resolve the project path
     const projectPath = resolve(path);
 
-    const args = [`start`, `--project_path="${projectPath}"`];
-    execFileSync(`npm`, args, { shell: true, cwd: `${appPath}` });
+    // Start trying from port 3000 and increase the port number if the port is already in use
+    let port = 3000;
+    let args = ['-s', `start`, `--project_path="${projectPath}"`, '--', `--port ${port}`];
+    
+    while (port < 3099) {
```
Fuzzbender, line 1181: Rather theoretical issue, but what happens if all 3000 - 3099 are reserved?
To my eyes, it just finishes with "200".

tools/data-handler/src/command-handler.ts
```diff
@@ -1167,18 +1167,27 @@ export class Commands {
...
+      try {
+        console.log(`Starting Cyberismo on http://localhost:${port}. Enter CTRL+C to quit.`);
+        execFileSync(`npm`, args, { shell: true, cwd: `${appPath}` });
+        break;
+      } catch (error) {
+        port++;
+        args = [`start`, `--project_path="${projectPath}"`, '--', `--port ${port}`];
```
Fuzzbender, line 1188: With single row: 
`args = [`start`, `--project_path="${projectPath}"`, '--', `--port ${++port}`];`

Fuzzbender (approved):
Fix the error-issue from linter, then this is all good.

PR #115, opened 2024-09-02
INTDEV-396 Use template categories in the create dialog

samutoljamo-pareto:
This PR adds categories to app's new card modal.
image
image


samutoljamo-pareto:
Should we sort the categories?

PR #117, opened 2024-09-05
INTDEV-381 E2E acceptance tests

korva:
- Added Cypress to the project as dev dependency for implementing e2e UI tests. Also added start-server-and-test for easier launching of the tests.
- To avoid conflicts with data-handler, added separate test project for app, found in __tests__/test-data/decision-records. This is also now used by unit and API tests.
- Implemented e2e tests for viewing a project, adding and removing a card, modifying card metadata and content and adding an attachment. Success is determined by checking UI elements on screen. See /cypress/e2e/app.cy.ts
- Added ids or classes to certain UI components so that Cypress can find and interact with them

Tests can be run in two ways:
`npm run e2e`: Runs tests within a browser (so you can view the results and steps with the Cypress suite)
`npm run e2e:headless`: runs tests headless (only results posted in console), this should be possible in CI too.

Cypress tests are run with a dev build but using test env (so that .env.test is used to set the test project root). For this an additional npm script `test-env-dev` was needed.

Since the tests modify the test project, a backup is created before running said tests. After the tests, the modified test project is deleted and the initial test project is restored from the backup. See cypress.config.ts.


tools/app/app/components/AppToolbar.tsx
```diff
@@ -34,6 +34,7 @@ export default function AppToolbar({ onNewCard }: AppToolbarProps) {
       </Box>
       <Box sx={{ flexGrow: 1 }} />
       <Button
+        id="createNewCardButton"
```
samutoljamo-pareto, line 37: I think it would be beneficial to consider making the references to all components either by aria-labels/text.

tools/app/package.json
```diff
@@ -9,7 +9,10 @@
     "lint": "next lint",
     "clean": "rm -rf node_modules && rm -f package-lock.json && npm cache clean --force",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
```
samutoljamo-pareto, line 12: Jest could have its own command and maybe we could run e2e in headless mode when using the main `npm test` command. Then it would be also ran in the pipeline

samutoljamo-pareto:
Also, the test "selects a card, shows card content and actions" seems to fail. I think there are some issues the tree view itself.

tools/app/app/components/AppToolbar.tsx
korva: Hmm, can you elaborate? I initially referred to the components via text/content but that can change in the future (or if language changed) which would then break the tests. And shouldn't aria-labels be also human-readable for screen readers and such?

tools/app/package.json
korva: Yep, good idea!

korva:
> Also, the test "selects a card, shows card content and actions" seems to fail. I think there are some issues the tree view itself.

The tests pass for me locally, so I'm thinking there could be some timing related issue here (after certain actions, the tests do not wait long enough for UI to update). I'll look into this. @samutoljamo-pareto could you add the failing cypress test raport here?

Also, the CI test run seems to fail due to some data-handler tests.

tools/app/app/components/AppToolbar.tsx
samutoljamo-pareto: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

Actually, cypress seems to encourage to use data attributes

tools/app/__tests__/test-data/decision-records/.cards/local/.schema
Fuzzbender, line 1: Is this the same "decision-records" as in "data-handler/test/test-data"?

Could we have a symbolic link, or use the same files? There's a bunch changes coming up that affect the test data content; thus I find updating the test data rather frustrating.

Fuzzbender (approved):

tools/app/app/components/AppToolbar.tsx
korva: Yes, that makes sense, I'll change to using data-cy for this!

korva: Fixed in 9a9a35b

samutoljamo-pareto:
I reran the tests a few times in and not in headless mode, but they passed

korva:
Phew, I finally got the tests to work on Linux and Windows CI!
I needed to add cross-env dev dependency so that the correct test env can be set also on Windows.
I also needed to remove the "tree menu clicking to open card" part of the test, because I just could not get it to work reliably. It always passed locally but failed in CI, and the methods I found to mitigate it (wait for page load or API requests) did not work. Now the test navigates directly to the URL, as is the case in other tests too.
The test case in question is 'selects a card, shows card content and actions'

The e2e tests now use the shared /data-handler/test/test-data/valid/decision-records data again.

FYI @samutoljamo-pareto

tools/app/__tests__/test-data/decision-records/.cards/local/.schema
korva: Fixed in a6d4ece

tools/app/package.json
korva: Implemented in [70349af](https://github.com/CyberismoCom/cyberismo/pull/117/commits/70349af39ab127c0a4d922f08d67d4e0838d4789)

PR #118, opened 2024-09-05
Improve parsing of clingo results

samutoljamo-pareto:


tools/data-handler/src/utils/ClingoParser.ts
Fuzzbender, line 1: File name should be clingo-parser.ts so that it follows the same naming as other files.

calc/common/base.lp
Fuzzbender, line 1: Please do not use abbreviations "calc" but full name "calculations"

tools/data-handler/src/calculate.ts
```diff
@@ -402,35 +331,55 @@ fieldtype(X, Field, "cardkeys") :- field(X, Field, _), card(Value) : field(X, Fi
   public async run(
     projectPath: string,
     cardKey: string,
-  ): Promise<ParseResult[] | undefined> {
+  ): Promise<ParseResult | undefined> {
     Calculate.project = new Project(projectPath);
 
-    const card = await Calculate.project.findSpecificCard(cardKey);
+    /*     const card = await Calculate.project.findSpecificCard(cardKey);
```
Fuzzbender, line 337: Should the commented block be removed? If not, add reason why the block is commented out.

tools/data-handler/src/utils/ClingoParser.ts
```diff
@@ -0,0 +1,261 @@
...
+  };
+  deniedOperations: {
+    transition: { transitionName: string; errorMessage: string }[];
+    move: { errorMessage: string }[];
+    delete: { errorMessage: string }[];
+    editField: { fieldName: string; errorMessage: string }[];
+    editContent: { errorMessage: string }[];
+  };
+  results: Result[]; // Nested results
+  [key: string]: any;
```
Fuzzbender, line 34: Instead of "any", what would be the potential actual types? 
string|number|boolean|...

tools/data-handler/src/utils/ClingoParser.ts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+export interface ParseResult {
+  results: Result[];
+  error: string | null;
+}
+
+class ClingoParser {
+  private possibleKeywords = [
```
Fuzzbender, line 43: Remove 'possible' from name.

Fuzzbender (approved):

tools/data-handler/src/utils/ClingoParser.ts
samutoljamo-pareto: Done

calc/common/base.lp
samutoljamo-pareto: Done

tools/data-handler/src/calculate.ts
samutoljamo-pareto: This is still a draft PR and the run method was temporarily changed so that you can try running query files

tools/data-handler/src/utils/ClingoParser.ts
samutoljamo-pareto: Done

tools/data-handler/src/utils/ClingoParser.ts
samutoljamo-pareto: I improved the types. For now, all fieldtypes are parsed as strings in the parser

tools/data-handler/src/calculate.ts
```diff
@@ -398,43 +323,41 @@ fieldtype(X, Field, "cardkeys") :- field(X, Field, _), card(Value) : field(X, Fi
...
+      '-V0',
+      main,
+      queryLanguage,
+      filePath,
+    ];
+    const clingo = spawnSync(this.logicBinaryName, args, {
+      encoding: 'utf8',
+    });
+    // print the command
+    console.log(`Ran command: ${this.logicBinaryName} ${args.join(' ')}`);
```
Fuzzbender, line 357: Perhaps slightly better verb would be 'execute'; thus 'Executed command: ...'

tools/data-handler/src/calculate.ts
```diff
@@ -398,43 +323,41 @@ fieldtype(X, Field, "cardkeys") :- field(X, Field, _), card(Value) : field(X, Fi
...
+
     if (clingo.stdout) {
+      console.log(`Clingo output: \n${clingo.stdout}`);
```
Fuzzbender, line 360: Do we need to state that it is "clingy" output? Would "Calculation output" do?

tools/data-handler/src/calculate.ts
Fuzzbender: On the other hand we are using 'run' as a parameter. ---> Ignore the above.

tools/data-handler/src/utils/clingo-parser.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+interface Link {
```
Fuzzbender, line 13: This interface should perhaps be somewhere elsewhere, as it is general link interface (even if there aren't any other users for it yet). Move it somewhere?

Fuzzbender (approved):

tools/data-handler/src/utils/clingo-parser.ts
samutoljamo-pareto: Console logging the outputs should be a temporary solution anyway. Maybe it's a good reminder to fix it sooner or later? And as a debug log, I don't think it's a bad idea to specify it is clingo output, because then the debugger knows it is clingo and not clingraph for example.

tools/data-handler/src/calculate.ts
samutoljamo-pareto: Console logging the outputs should be a temporary solution anyway. Maybe it's a good reminder to fix it sooner or later? And as a debug log, I don't think it's a bad idea to specify it is clingo output, because then the one who debugs knows it is clingo and not clingraph for example.

tools/data-handler/src/utils/clingo-parser.ts
samutoljamo-pareto: It actually differs from the general Link interface. I renamed it to CalculationLink. I think it should be in the same place as other interfaces related to calculations. Though we might want to move them too.

PR #119, opened 2024-09-05
Sort exported cards by rank

samilempinen:


Fuzzbender (approved):

samutoljamo-pareto (approved):
I think we should provide a sort function for cards, because it makes no sense we have recreate the same function in many locations. But maybe not in this ticket

PR #130, opened 2024-09-13
INTDEV-385 log validation failures

Fuzzbender:
Validation now always logs to console error all of the findings. 
Instead of appended string, it is now using string array, which helps showing each failure on its own line.

Improved some of the logging messages to make more sense and avoid having [object] in them.

Fixed all of the invalid projects to fail like they should fail, and not just random failures due to not being updated with latest changes. 

Finally, added a unit test that goes through all of the invalid projects and expects that they all fail validation.


korva (approved):
Looks good!

PR #131, opened 2024-09-13
INTDEV-368 Validate template cards

Fuzzbender:
Add card content validation to template cards. 
They have similar requirements as project cards, but workflowState must be empty.


tools/data-handler/test/test-data/valid/minimal/.cards/local/templates/test-template/template.json
```diff
@@ -1,4 +1,4 @@
 {
-  "buttonLabel": "Create a New Decision",
-  "namePrompt": "The title of the new decision:"
+    "buttonLabel": "",
```
samutoljamo-pareto, line 2: We just removed buttonLabel/namePrompt, so either this must be reverted and rebased or just remove them also from here

tools/data-handler/src/validate.ts
```diff
@@ -423,35 +428,54 @@ export class Validate {
...
         }
       } else {
-        validationErrors += `Workflow of '${cardType.workflow}' cardtype '${cardType?.name}' does not exist in the project`;
+        validationErrors.push(
+          `Workflow of '${cardType.workflow}' cardtype '${cardType?.name}' does not exist in the project`,
+        );
       }
     }
-    return validationErrors;
+    return validationErrors.join('\n');
```
samutoljamo-pareto, line 478: I think it would be better to represent validation errors using a list of either objects or just strings. But that deserves an own PR

samutoljamo-pareto (approved):

tools/data-handler/test/test-data/valid/minimal/.cards/local/templates/test-template/template.json
Fuzzbender: I'll rebase this and fix if there are issues.

tools/data-handler/src/validate.ts
samutoljamo-pareto: Looks like #130 does that :smile:

PR #133, opened 2024-09-16
INTDEV-394 remove buttonLabel and namePrompt

samutoljamo-pareto:
Removed all occurrences of namePrompt/buttonLabel

### DIFF
--- a/tools/data-handler/test/command-handler.test.ts
+++ b/tools/data-handler/test/command-handler.test.ts
@@ -1143,8 +1143,7 @@ describe('create command', () => {
   // template
   it('template (success)', async () => {
     const templateName = 'template-name_first';
-    const templateContent =
-      '{"buttonLabel": "Button1", "namePrompt": "Prompt1"}';
+    const templateContent = '{}';
     const result = await commandHandler.command(
       Cmd.create,
       ['template', templateName, templateContent],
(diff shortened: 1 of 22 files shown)


tools/data-handler/test/command-handler.test.ts
```diff
@@ -1186,8 +1183,7 @@ describe('create command', () => {
   });
   it('template with "123"', async () => {
     const templateName = 'loc/123';
-    const templateContent =
-      '{"buttonLabel": "Button1", "namePrompt": "Prompt1"}';
+    const templateContent = '{}';
```
Fuzzbender, line 1186: As `templateContent` is optional, there could be a test that doesn't pass the `content` parameter at all.
Either a new test, or encapsulate to one of the existing ones.

Fuzzbender: Actually, if the `content` file is always empty, shouldn't we remove the whole parameter from CLI?

Fuzzbender (approved):
Check if the content parameter could be removed as unnecessary from CLI.

samutoljamo-pareto:
Its optional so I think we should just leave it. Maybe you'd want to give a displayName while creating the template

tools/data-handler/test/command-handler.test.ts
samutoljamo-pareto: I'll add a new test

PR #136, opened 2024-09-17
Empty 'person' fieldtype causes perplexing validation error

Fuzzbender:
Running validation in our "isms" content repo causes perplexing `Cannot read properties of undefined (reading 'length')` error. This is caused by empty `person` data-type in one of the field types. 

To avoid data-type validation to fail, check that the object exists.


korva (approved):
OK by me!

PR #140, opened 2024-09-19
Further implementation on calculations

samutoljamo-pareto:
While this PR was originally for INTDEV-277, it also contains INTDEV-439 and INTDEV-437.
Created features
* Generation of workflows and cardtypes to logic programs
* Removing newlines from the values of fields
   * This is a temporary solution so that it generates valid logic programs in the cyberismo-isms project
* Basic way to run these queries from frontend

Some thoughts:
I don't think we should merge this yet and we could talk more about this on monday. After a little of bit thinking, the clingo queries can be thought to be readonly SQL queries. And while using these named queries is easy at this point, it ties the frontend also to clingo. I was thinking that we could rather make normal HTTP routes, which use clingo on the backend. In this way, the implementation of the HTTP calls could be replaced in the future without making changes to the actual app.


samutoljamo-pareto:
Also, here's an example of .calc directory using this branch:
image

Fuzzbender:
Yes, I think these should - at least later - go through data-handler. Otherwise we'll be in the deep end with the app too in the multi-user-version. 

I still need to read through the code with some thought. Initially, looks good.

samutoljamo-pareto:
I decided to add calculate.generate() call to the tree query, which makes it a bit slower. That way, all the previous functionalities work still. I'll add windows support to this and then we could push this version to main

samutoljamo-pareto:
Also, the progress indicator is a placeholder can there's a separate ticket about amount.

samutoljamo-pareto:
It seems to now work on windows and the pipeline is also working

tools/app/app/components/TreeMenu.tsx
```diff
@@ -12,38 +12,28 @@
...
-
-const RenderTree = (onCardSelect?: (node: NodeApi<CardNode>) => void) =>
-  function RenderNode({
-    node,
-    style,
-    dragHandle,
-  }: NodeRendererProps<CardNode>) {
+const RenderTree = (onCardSelect?: (node: NodeApi<Result>) => void) =>
+  function RenderNode({ node, style, dragHandle }: NodeRendererProps<Result>) {
+    const workflowStateCategory = getString(node.data, 'workflowStateCategory');
```
korva, line 36: Would it be cleaner if app backend (/api/tree/route.tsx) would parse the Result(s) into an object defined in definitions.ts? That way we would not need to access these named fields in app code and do null checks.

korva:
During e2e tests lots of clingo output is printed, is this necessary?

```
Ran command: clingo - --outf=0 --out-ifs=\n -V0 ../data-handler/test/test-data/valid/decision-records/.calc/main.lp ../data-handler/test/test-data/valid/decision-records/.calc/queryLanguage.lp /Users/jaakko/dev/cyberismo/cyberismo/calculations/queries/tree.lp
Clingo output: 
order(1,1,"rank","ASC")
order(2,1,"rank","ASC")
field(decision_5,"title","Decision Records")
field(decision_5,"rank","0|a")
field(decision_6,"title","Document Decisions with Decision Records")
field(decision_6,"rank","0|a")
field(decision_5,"workflowStateCategory","initial")
field(decision_6,"workflowStateCategory","closed")
childResult(decision_5,decision_6)
result(decision_5)
SATISFIABLE
```

tools/data-handler/test/test-data/valid/decision-records/.calc/imports.lp
```diff
@@ -0,0 +1,10 @@
+% ../data-handler/test/test-data/valid/decision-records/.calc/resources
+#include "/home/samu/projects/cyberismo/cyberismo/tools/data-handler/test/test-data/valid/decision-records/.calc/resources/decision/cardtypes/decision-cardtype.lp".
```
korva, line 2: These absolute paths are updated when running tests, resulting into this file changing (along with modules.lp). Can we use relative paths here?

Fuzzbender:
https://github.com/CyberismoCom/cyberismo/pull/140#issuecomment-2373335809
I think we should add logging to file feature ASAP. Sometimes these kinds of lines can be really useful, but not everyone needs to see them.

korva:
Screenshot 2024-09-25 at 11 12 18
Small bug (not sure if related to this PR):
After creating a new card under an existing root card, the tree menu is not expanded to show the new card after creation.

tools/app/app/components/TreeMenu.tsx
samutoljamo-pareto: I think that probably makes sense. Unfortunately these cannot use the definitions of data-handler but it also kind of makes sense. Then we'll have separate definitions for REST API / the actual data model. Though, I'd rather do these by just extending the parseResult. We'd also want to have some kind of validation that makes sure clingo actually returns what we expect it to return. Maybe this could be actually done at data-handler level. Do you think we gain anything if we rename the variables? E.g. I don't mind using "results" for the children, as long as we use it everywhere

tools/data-handler/test/test-data/valid/decision-records/.calc/imports.lp
samutoljamo-pareto: Actually, I think we shouldn't add those to source control

samutoljamo-pareto:
I don't think the last one is related to this. But I can try if that works with the current main branch

samutoljamo-pareto:
Hmm I wasn't able to reproduce that in main or this branch. Also tried unified-sdl and cyberismo-isms

tools/app/app/components/TreeMenu.tsx
korva: Some sort of validation would be good. All of this could also be on data-handler side, e.g. in addition to Calculate.runQuery we could have convenience functions like getTree(queryName) which would return a validated "Tree" object based on the given query.

As for naming, IMHO clearer is better; "result" is generic and will not give clues on what we expect to find there.

samutoljamo-pareto: I would be kind of against validating. I think we should define typescript types for them and the "runQuery"-method could the type, by assuming the data is correct. The reason for this is that these queries are like SQL queries, and we wouldn't validate SQL-queries because we assume they give us the right result, if the query is correct. I think same kind of thinking could be applied here. We could make tests for the queries though

korva: Yeah, above way of doing it sounds good!

tools/data-handler/test/test-data/valid/decision-records/.calc/imports.lp
samutoljamo-pareto: I removed them from source control

tools/app/app/components/TreeMenu.tsx
samutoljamo-pareto: Did what I described so I resolved this

henryhaverinen:
Generating the tree view from a query seems to work great, and I even tested the hiddenInTreeView predicate successfully.

I noticed that with this branch, the titles of cards look weird in the tree menu, if there are parentheses in the title:
image

calculations/queries/tree.lp
```diff
@@ -0,0 +1,10 @@
+field(Card, "workflowStateCategory", Category) :-
+    field(Card, "workflowState", State),
+    field(Card, "cardtype", CardType),
+    field(CardType, "workflow", Workflow),
+    workflowState(Workflow, State, Category).
+
+select("title";"workflowStateCategory";"base/fieldtypes/progress";"rank").
```
Fuzzbender, line 7: No need to fix this in this PR, but.
This has now a dependency on module "base" and will not work without it. 
We should somehow skip calculations where the base module is not installed. 
Or we should always include base module in the software.

tools/app/__tests__/components.test.tsx
```diff
@@ -348,3 +349,251 @@ const testProject: Project = {
     },
   ],
 };
+
+const testTree: ParseResult['results'] = [
+  {
+    key: 'USDL-43',
+    path: '/Users/jaakko/dev/cyberismo/unified-sdl/cardroot/USDL-43',
```
Fuzzbender, line 356: We cannot have absolute paths in tests to user folders.

Fuzzbender: If this is for testing purposes only (the path can be whatever), then please change "jaakko" -> "test" or similar.

tools/app/app/api/tree/route.tsx
```diff
@@ -0,0 +1,50 @@
...
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { NextResponse } from 'next/server';
+import { Calculate } from '@cyberismocom/data-handler/calculate';
+
+export const dynamic = 'force-dynamic';
```
Fuzzbender, line 16: This is not used in the new code?

calculations/queries/tree.lp
samutoljamo-pareto: The progress is defined to be optional, so it works even if base module hasn't been added I think. Thought I do think, there could be at least some kind of base module, which is tied to cyberismo app, so we can create predefined calculated fields(like progress) which can be used in the queries

tools/app/__tests__/components.test.tsx
samutoljamo-pareto: Yes, the app doesn't interact with the path in anyway. I don't think the parseResult even has that field. I'll remove it.

Fuzzbender (approved):

tools/app/app/api/tree/route.tsx
samutoljamo-pareto: It is used. I removed /api/query

samutoljamo-pareto:
And to  @henryhaverinen 's comment: I think these should part of INTDEV-418. We might just escape all characters just like in web if there's no reason to not do that

henryhaverinen:
> And to @henryhaverinen 's comment: I think that should be done in INTDEV-418. We might just escape all characters just like in web if there's no reason to not do that

I looked at the generated logic and that seemed fine to me in these cases where you have parentheses in a string, so I suspect that this problem could be in the parsing of the results.

henryhaverinen (approved):

samutoljamo-pareto:
Yes, you're right. My regex catches the first parenthesis it can find even though it shouldn't

PR #141, opened 2024-09-23
INTDEV-445: Fix to 'Metadata and custom fields do not work in the ISMS project'

korva:
Fix app cardtype endpoint to support new resource naming scheme.
Pass the cardtype name as url parameter because it typically contains a path.
So the API call is now like:
GET /api/cardtypes?name=cisms/cardtypes/task

Added API test for this.


samutoljamo-pareto (approved):
Looks good to me too

PR #146, opened 2024-09-23
INTDEV-425: Print API request error responses in content area

korva:
Handles cases such as card not found, project empty, wrong path etc.


Fuzzbender (approved):

samutoljamo-pareto (approved):
Looks good but I think we should return error messages also as JSON. E.g. {"message": "things went wrong", "statusCode": 500} for now. That could be done in a different ticket though

PR #148, opened 2024-09-23
Make Windows CI tests more stable

Fuzzbender:
1. `try-catch` in problematic tests test data cleaning. Ignore errors.
2. increase timeout for certain cypress-tests.
3. Put tests that don't need test-data to their own sets and do not fiddle with test data in those (saves some time running tests).


korva (approved):
OK by me!

PR #151, opened 2024-09-24
INTDEV-438 Validation: Do not accept extra files or folders in project directory

Fuzzbender:
Make validation more strict in regards to extra folders and files. 

Generally, no extra files or folders are allowed in project files. Exceptions are
- on root directory, there can be extra files, but no directories
- under each `a` (attachment) directory there can be any files

### DIFF
--- a/tools/schema/cardtree-directory-schema.json
+++ b/tools/schema/cardtree-directory-schema.json
@@ -11,6 +11,10 @@
           "description": "Configuration for the card tree",
           "type": "object",
           "properties": {
+            "files": {
+              "type": "object",
+              "additionalProperties": false
+            },
             "directories": {
               "type": "object",
               "additionalProperties": false,
@@ -23,8 +27,10 @@
                   "properties": {
                     "files": {
                       "type": "object",
+                      "additionalProperties": false,
                       "properties": {
-                        "cardsconfig.json": {}
+                        "cardsconfig.json": {},
+                        ".schema": {}
                       },
                       "required": [".schema", "cardsconfig.json"]
                     },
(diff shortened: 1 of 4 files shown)


korva:
Running `cyberismo validate` for current cyberismo-isms repo main branch I get:
`At '.calc' non-allowed additional property: resources`

Fuzzbender:
> Running `cyberismo validate` for current cyberismo-isms repo main branch I get: `At '.calc' non-allowed additional property: resources`
> 
> And in unified-sdl: `At '.calc' non-allowed additional property: cardtree.lp`

.calc folder is generated when calculations are done. You get rid of old name of the lp-file nag by removing the folder and re-calculating and then validating. Not sure what the `resources` is, wasn't there previously. Probably something that needs to be just skipped (allowed).

Fuzzbender:
I can't get `resources` to appear under `.calc`. 
Have you executed calculations with a version of binary that is not in main and it has updated the calculations?

Fuzzbender:
I'll change the implementation so that it won't validate `.cards` folder at all. That folder can be seen as temp file for calculations and as such is ignored in validation.

samutoljamo-pareto:
Yes, resources folder is made in #140 . I noticed, it would be easy to generate the other files(workflows, cardTypes) etc if we store them by their name. However, since the name is in the format project/type/name,  I didn't want to store them under .calc directly. I ended up putting all of those to /resources folder.

tools/schema/cardtree-directory-schema.json
```diff
@@ -6,11 +6,18 @@
   "properties": {
     "directories": {
       "type": "object",
+      "additionalProperties": false,
```
samutoljamo-pareto, line 9: Shouldn't we allow any properties at the root level?

Fuzzbender: I think we could. I'll change it.

henryhaverinen (approved):

PR #152, opened 2024-09-26
Improve testing for calculations

samutoljamo-pareto:
Added a few simple tests and also a test for a the full query.

Note: Clingo-parser.test.ts contains more detailed tests for the actual parser

### DIFF
--- a/tools/data-handler/test/calculate.test.ts
+++ b/tools/data-handler/test/calculate.test.ts
@@ -1,0 +1,45 @@
+import { expect, use } from 'chai';
+import { dirname, join } from 'node:path';
+import { mkdirSync, rmSync } from 'node:fs';
+
+import { Calculate } from '../src/calculate.js';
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto:
@Fuzzbender This is ready now. We should definitely add a better test project, where we could test all features

Fuzzbender (approved):

Fuzzbender:
> @Fuzzbender This is ready now. We should definitely add a better test project, where we could test all features

I think the testing data requires a bit of thought. If we have huge test content and we copy-and-delete it numerous times in the tests, it will not fly. It will make Windows tests crawl (even worse than they now do).

samutoljamo-pareto:
We could consider trying to inject the node fs methods that are used. Or make an abstract version of those operations and call memfs methods in tests so that the tests aren't actually using the filesystem. It might reduce the quality of tests a bit though. https://www.npmjs.com/package/memfs

PR #153, opened 2024-09-26
Asynchorous call is not 'await'ed

Fuzzbender:
`save()` function is asynchronous, but it is not `awaited` in `setCardPrefix`. 
Fix the async function calling.


samutoljamo-pareto (approved):
Looks good. Maybe this is the reason for the mysterious windows CI failures

PR #154, opened 2024-09-27
INTDEV-443 Rename does not rename resources

Fuzzbender:
Fix `rename` command after module naming support - INTDEV-420.

Rename now changes all cards and all related resources (card types. workflows, link types, calculations, templates, field types) to new project prefix. Since renaming never affects imported module content, I added to resource fetching from `Project` possibility to fetch just local resources.

Updated cards have their `lastUpdated` timestamps updated.

Additionally, added a check that if project prefix does not change (trying to rename to same prefix), it only complains about that.

Couple changes that are non-essential but I included in this PR:
1) Renamed the `updateMetadata` function from `Project` to `updateMetadataKey` since it only changes only key (and I introduced a new function that updates the whole metadata).
2) Introduce a new helper `writeJsonFile` so that caller doesn't have to remember to `formatJson` when writing JSON file.
3) Functions in `Project` have been reordered alphabetically (scoped by `public`, `protected`, `private`), to make them easier to find.

### DIFF
--- a/tools/data-handler/src/containers/card-container.ts
+++ b/tools/data-handler/src/containers/card-container.ts
@@ -15,7 +15,7 @@ import { basename, join, sep } from 'node:path';
 import { readdir, readFile, writeFile } from 'node:fs/promises';
 
 // ismo
-import { formatJson } from '../utils/json.js';
+import { writeJsonFile } from '../utils/json.js';
 import { getFilesSync, pathExists } from '../utils/file-utils.js';
 
 // interfaces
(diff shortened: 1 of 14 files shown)


korva (approved):
Code looks good. One test kept failing for me, probably because of some of the new folders being created by current main (that has the big calculations PR merged now)? Maybe rebase and check before merging?

1 failing

  1) validate cmd tests
       validate() - decision-records (success):

      AssertionError: expected 54 to equal +0
      + expected - actual

      -54
      +0
      
      at Context.<anonymous> (file:///Users/jaakko/dev/cyberismo/cyberismo/tools/data-handler/test/0_validate.test.ts:20:33)

Fuzzbender:
> Code looks good. One test kept failing for me, probably because of some of the new folders being created by current main (that has the big calculations PR merged now)? Maybe rebase and check before merging?
> 
> 1 failing
> 
> 1. validate cmd tests
>    validate() - decision-records (success):
>    AssertionError: expected 54 to equal +0
>    
>    * expected - actual
>    
>    -54
>    +0
>    at Context. (file:///Users/jaakko/dev/cyberismo/cyberismo/tools/data-handler/test/0_validate.test.ts:20:33)

Not reproducing locally (on top of latest main), nor in CI.

PR #156, opened 2024-09-30
Code changes from hardened ESLint

Fuzzbender:
When investigating a recent CI failures, the ESLint was using a really strict settings. 
These code issues were found with those settings - even if we don't harden the ESLint configuration, it makes sense to fix the problematic findings. 

The changes in this commit are: 
* Silence test result log messages in data-handler/tests; only actual failures are shown
* Use `new Error` when throwing
* Define the JSON type when reading from file
* Remove unnecessary `async` when function does not have any asynchronous activity
* Remove `await` from `return await` as unnecessary 
* Remove `await` when calling sync functions
* Use `bind` to get a valid function reference
* Remove unnecessary empty constructors
* Fix one overly complicated function that fetches supported data types from a schema; and replace it with just a string array.
* Add `await` when calling async function when needed


korva (approved):
Looks good!

PR #158, opened 2024-09-30
INTDEV-420 Change all types to follow naming convention

Fuzzbender:
Remove `examples` folder.
Rename all resource not to have unneeded `-<resourceType>` at the end. For example; `decision-workflow`-> `decision`.

This requires all of the content repos to be updated with the naming changes as well.

**Note** Due to the way how git uses filenames, it is strongly recommended to re-clone this repository before you pull these changes to your local environment. 
After this PR is merged, it is **imperative** that everyone re-clones this and all affected content repositories.


tools/app/app/api/cardTypes/route.tsx
```diff
@@ -17,15 +17,15 @@ export const dynamic = 'force-dynamic';
 
 /**
  * @swagger
- * /api/cardtypes:
+ * /api/cardTypes:
```
samutoljamo-pareto, line 20: Do you think it's beneficial to also use camelCase for routes? I've never seen that. I think most apis use lowercase / underscores.

tools/data-handler/src/interfaces/project-interfaces.ts
```diff
@@ -17,33 +17,33 @@
 export interface dotSchemaContent {
   id: string;
   version: number;
-  cardkeyPrefix?: never;
+  cardKeyPrefix?: never;
   name?: never;
 }
 
-// Project's settings (=cardsconfig.json).
+// Project's settings (=cardsConfig.json).
```
samutoljamo-pareto, line 24: Could we capitalize the interfaces? It's fairly common in typescript. https://ts.dev/style/#identifiers

Fuzzbender: Yes, but I will do these separately:
1) change schemas to follow our naming rules
2) change interfaces to follow to follow TypeScript naming rules

tools/app/app/api/cardTypes/route.tsx
Fuzzbender: If they work, then why not?

Fuzzbender: We are already looking a bit `weird` in the Linux filesystem with `cardRoot` and `cardTypes`. More natural for Linux filesystem folder/file names would be `card-root` and `card-types`.

samutoljamo-pareto (approved):

PR #159, opened 2024-10-01
Fix table of contents highlight

korva:
Update highlight and current active title upon scroll, so position is retained when switching between edit/view modes and table of contents highlight works again.


samutoljamo-pareto (approved):
Looks and tested that it works

korva:
I found some cases where position changes awkwardly when switching to edit mode (e.g. when you are scrolled to the top/title and click edit -> edit mode scrolls to first content header), need to fix that

PR #160, opened 2024-10-01
Encode clingo values

samutoljamo-pareto:
I considered a few options for encoding clingo characters. Since I didn't yet a good idea of characters we could whitelist/blacklist, I ended thinking of encoding/decoding fieldValues for clingo. I considered:
1. Base64
Issue with base64 is that the queries interact with the values and it would be annoying to use base64 for referencing thos e values.
2. Replacing all but certain characters with a formatted string.
I ended up going with this

I put a few things into the whitelist: a-z, A-Z, 0-9 and a few special chars(/, space, |, : and .)

Here's what the generated version looks like:
image

### DIFF
--- a/tools/data-handler/src/utils/clingo-parser.ts
+++ b/tools/data-handler/src/utils/clingo-parser.ts
@@ -13,6 +13,21 @@
 import { Project } from '../containers/project.js';
 import { BaseResult, ParseResult } from '../types/queries.js';
 
+export const CLINGO_VALUE_ENCODING = 'base64url';
+
+export function encodeClingoValue(value: string) {
+  return value.replace(/[^a-zA-Z0-9 \/]/g, (char) => {
+    const code = char.charCodeAt(0);
+    return `'u${('0000' + code.toString(16)).slice(-4)}`;
+  });
+}
+
+export function decodeClingoValue(value: string) {
+  return value.replace(/'u([0-9a-fA-F]{4})/g, (match, code) => {
+    return String.fromCharCode(parseInt(code, 16));
+  });
+}
+
 class ClingoParser {
   private keywords = [
     'queryError',
@@ -83,7 +98,7 @@ class ClingoParser {
     },
     field: (key: string, fieldName: string, fieldValue: string) => {
       const res = this.getOrInitResult(key);
-      res[fieldName] = fieldValue;
+      res[fieldName] = decodeClingoValue(fieldValue);
     },
     label: (key: string, label: string) => {
       const res = this.getOrInitResult(key);
--- a/tools/data-handler/src/utils/clingo-parser.ts
+++ b/tools/data-handler/src/utils/clingo-parser.ts
@@ -16,14 +16,14 @@ import { BaseResult, ParseResult } from '../types/queries.js';
 export const CLINGO_VALUE_ENCODING = 'base64url';
 
 export function encodeClingoValue(value: string) {
-  return value.replace(/[^a-zA-Z0-9 \/]/g, (char) => {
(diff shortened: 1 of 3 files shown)


henryhaverinen:
In general, this approach should be fine, as long as we have enough characters on the allow list so that only problematic special characters end up getting encoded. For example characters such as -, _,  æ, Й, #, (, ), {, }, %, $, @,  €, =, *, ^ etc. should not cause problems and they should be available in the logic program strings as is.
I wonder if there is some easy way to get this allow list done? Or should it be a deny list?

henryhaverinen:
From the Potassco user guide: Furthermore, a string is an arbitrary sequence of characters enclosed in double quotes (" ̈"), where any occurrences of ‘\’, newline, and double quote must be escaped via ‘\\’, ‘\n’, or ‘\"’, respectively.
I wonder how much trouble we would get if there would be a simple deny-list of those 3 characters? And we could even encode them using the Clingo notation?

samutoljamo-pareto:
You're right, I'll make that change

Fuzzbender:
> In general, this approach should be fine, as long as we have enough characters on the allow list so that only problematic special characters end up getting encoded. For example characters such as -, _, æ, Й, #, (, ), {, }, %, $, @, €, =, *, ^ etc. should not cause problems and they should be available in the logic program strings as is. I wonder if there is some easy way to get this allow list done? Or should it be a deny list?

In general, it is safer to use allow-list instead of deny-list.

tools/data-handler/src/utils/clingo-parser.ts
```diff
@@ -13,6 +13,21 @@
 import { Project } from '../containers/project.js';
 import { BaseResult, ParseResult } from '../types/queries.js';
 
+export const CLINGO_VALUE_ENCODING = 'base64url';
+
+export function encodeClingoValue(value: string) {
```
Fuzzbender, line 18: Add comment-blocks for the new exported functions.

tools/data-handler/test/utils/clingo-parser.test.ts
Fuzzbender, line 1: Should there be a unit test for using non-encoded value? Probably leads to an error... but I reckon that could be something that a programmer might attempt at some point.

Fuzzbender (approved):

tools/data-handler/test/utils/clingo-parser.test.ts
samutoljamo-pareto: Yes.

samutoljamo-pareto:
I rewrote the parser.

I went with a blacklist, since  @henryhaverinen provided the values that should be escaped. Now the parser also handled escaped values correctly: If the parser is inside a quoted string, it accepts anything until it finds the next quote. if "\" is seen, it sets a flag "escapeNext" and sets it regardless of the characters. Also seems to fix issues in the UI

tools/data-handler/src/utils/clingo-parser.ts
samutoljamo-pareto: Stopped using base64 encoding

Fuzzbender:
> I rewrote the parser.
> 
> I went with a blacklist, since @henryhaverinen provided the values that should be escaped. Now the parser also handled escaped values correctly: If the parser is inside a quoted string, it accepts anything until it finds the next quote. if "" is seen, it sets a flag "escapeNext" and sets it regardless of the characters. Also seems to fix issues in the UI

I think in this case deny-list is fine, as there are so few values that are denied, and it is likely that the list is not changed in the future.

henryhaverinen (approved):

PR #163, opened 2024-10-02
INTDEV-420 Addendum: Fix all interfaces to follow TS recommendations

Fuzzbender:
Make all D-H interfaces follow TypeScript recommendations. 
Merge overlapping app and data-handler types. Now, `app` re-uses `data-handler` interfaces. In some cases expands and/or renames them. Names are also harmonised between the two. If reviewers have better names, let's have'em!

For some weird reason, `MetadataView` started to complain about `any` types with just naming changes, so some of the types are now defined in `map` and `filter` operations.


tools/data-handler/src/interfaces/project-interfaces.ts
```diff
@@ -215,23 +231,26 @@ export interface fetchCardDetails {
...
 };
 
-export interface importCsvRow {
-  title: string;
-  template: string;
-  description?: string;
-  labels?: string[];
-  [key: string]: string | string[] | undefined; // Custom fields
-}
+// todo: this is apparently not used?
```
Fuzzbender, line 249: If there is some reason why this is apparently unused, I can keep it. Otherwise, before merging this PR, I will delete it.

tools/data-handler/src/interfaces/project-interfaces.ts
```diff
@@ -115,97 +130,98 @@ export interface linkType {
...
   name: string;
   path?: string;
 }
 
 // Template content.
-export interface templateContent {
-  name: string;
-  cards?: card[];
-}
+// todo: not used???
```
Fuzzbender, line 147: If there is some reason why this is apparently unused, I can keep it. Otherwise, before merging this PR, I will delete it.

tools/data-handler/src/interfaces/project-interfaces.ts
samutoljamo-pareto: I think it can be deleted

samutoljamo-pareto (approved):
Looks good to me

korva (approved):
Good PR. Approved!

Fuzzbender:
I'' remove the unused interfaces and reorder alphabetically interfaces in `definitions.ts` and `project-interfaces.ts` files before merging this to `main`.

PR #166, opened 2024-10-03
INTDEV-466: Toolbar ux fixes

korva:
- Harmonize margins and button sizes/types (… and link button, use same borderless variant)
- Add tooltips to icon buttons
- Move cancel button to be next to update button (group similar actions)
- Context menu: move delete to be last, remove metadata action as it is now inline already
- Click on Cyberismo log or project name should redirect to project root (/cards)
- Breadcrumbs: when long paths are shown, toolbar vertical size should change too to accommodate the multi-line path. Also line paddings can be decreased.
- Breadcrumbs: upon hover, should indicate that path sections are links (underline)


tools/app/app/components/CardContextMenu.tsx
```diff
@@ -12,24 +12,18 @@
...
-
-import { CardMetadata } from '../lib/definitions';
 import {
   MenuButton,
   Menu,
   MenuItem,
   Divider,
   Typography,
   Dropdown,
+  Tooltip,
```
Fuzzbender, line 22: In general, it would be helpful to keep the imports in alphabetical order, if possible. 
Makes finding the needed item when reading the code a bit easier.

Fuzzbender:
Looks good to me; caveat: I am not a UI developer.

tools/app/app/components/CardContextMenu.tsx
korva: Wonder if we could get Prettier to do that automatically

korva:
I separated the linking UI improvements into its own task (https://cyberismo.atlassian.net/browse/INTDEV-471) because it likely requires a larger refactoring.

tools/app/app/components/CardContextMenu.tsx
Fuzzbender: Dev depencies are "ok". Doesn't really matter, if it makes our life easier.

samutoljamo-pareto (approved):
Looks good to me. I'd recommend using joy/mui features if possible instead of divs etc directly. And that same applies to spacing

korva:
> Looks good to me. I'd recommend using joy/mui features if possible instead of divs etc directly. And that same applies to spacing

Good point! I admit I got frustrated with trying to get the Status button with the color icon to work nicely with margins, paddings and such, and once I got it to work with <span> and <div> I just continued ahead :)

PR #167, opened 2024-10-03
INTDEV-420 Rename schemas and datatypes according to naming rules

Fuzzbender:
Further changes related to naming changes.

Update schemas and data types to follow the new naming rules.


korva (approved):
Seems good!

PR #171, opened 2024-10-06
INTDEV-462 Validate 'always' and 'optionally' visible fields

Fuzzbender:
Validate that `alwaysVisibleFields` and `optionallyVisibleFields` contain fieldTypes that exist in the project.

Additionally, remove duplicates from returned validation errors (all validation errors, not just these) - as issue with shared resource can be reported multiple times.

### DIFF
--- a/tools/data-handler/src/validate.ts
+++ b/tools/data-handler/src/validate.ts
@@ -98,6 +98,49 @@ export class Validate {
     return join(file.path, file.name);
   }
 
+  private parseValidatorMessage(errorObject: object[]): string {
+    let parsedErrorMessage = '';
+    // todo: get schema name here?
+    for (const error of errorObject) {
+      let instancePath = '';
+      let params = '';
+      let message = '';
+      let fileError = false;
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/validate.ts
```diff
@@ -158,6 +201,35 @@ export class Validate {
...
+    index: number,
+    array: string[],
+  ) {
+    return array.indexOf(value) === index;
+  }
+
+  private async validateArrayOfFields(
+    project: Project,
+    cardType: CardType | undefined,
+    fieldArray: string[] | undefined,
```
samutoljamo-pareto, line 216: I think it would be generally better to handle undefined values at higher levels rather than at lower levels

samutoljamo-pareto (approved):
Looks good to me.

tools/data-handler/src/validate.ts
Fuzzbender: Fixed. Caller now handles `undefined` and uses empty string arrays if values are missing.

PR #180, opened 2024-10-08
INTDEV-487 Collect attachments from one folder only, not from children

Fuzzbender:
Change `doCollectAttachments` to collect only from a given folder (and its non-children-card subfolders).

### DIFF
--- a/tools/data-handler/src/containers/card-container.ts
+++ b/tools/data-handler/src/containers/card-container.ts
@@ -14,7 +14,6 @@
 import { basename, join, sep } from 'node:path';
 import { readdir, readFile, writeFile } from 'node:fs/promises';
 
-// ismo
 import { writeJsonFile } from '../utils/json.js';
 import { getFilesSync, pathExists } from '../utils/file-utils.js';
 
@@ -172,8 +171,7 @@ export class CardContainer {
     return foundCards;
   }
 
-  // Collects all attachments from container.
-  // Function collects attachments from one folder and recurses to valid, potential folders.
+  // Function collects attachments from all cards in one folder.
   private async doCollectAttachments(
     folder: string,
     attachments: CardAttachment[],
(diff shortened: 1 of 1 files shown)


Fuzzbender:
@korva can you check that this works properly in your use-case?

korva:
> @korva can you check that this works properly in your use-case?

Improved but not quite fixed, steps to reproduce:

On feature/samimerila/INTDEV-487 (commit 878c1c28743b9bfee1a3729264b7861ed71d7869)
1. Opening Cyberismo ISMS project in app, go to card /cards/cisms_293 (Security review 13.8.2024)
2. Edit -> add attachment image
3. Navigate to parent card /cards/cisms_128 (2024)
4. Edit -> Attachment image is visible also in parent card attachments list

However, when you go to the next parent card /cards/cisms_71 (Quartal security reviews) the attachment is not visible.
So looks like it still goes one level too deep?

Fuzzbender:
> > @korva can you check that this works properly in your use-case?
> 
> Improved but not quite fixed, steps to reproduce:
> 
> On feature/samimerila/INTDEV-487 (commit [878c1c2](https://github.com/CyberismoCom/cyberismo/commit/878c1c28743b9bfee1a3729264b7861ed71d7869))
> 
> 1. Opening Cyberismo ISMS project in app, go to card /cards/cisms_293 (Security review 13.8.2024)
> 2. Edit -> add attachment image
> 3. Navigate to parent card /cards/cisms_128 (2024)
> 4. Edit -> Attachment image is visible also in parent card attachments list
> 
> However, when you go to the next parent card /cards/cisms_71 (Quartal security reviews) the attachment is not visible. So looks like it still goes one level too deep?

Without looking at the code, I think this is due to that: 
a) app and CLI use the same function
b) CLI needs the child attachments, since it is showing them with `cyberismo show cards --details` 

I'll check the call path; maybe I can filter out the child attachments for app.

korva (approved):
Approved, remaining issue (caused by empty a folder) handled here: https://cyberismo.atlassian.net/browse/INTDEV-491

PR #181, opened 2024-10-08
INTDEV-459 Delete old .calc folder before starting new calculations

Fuzzbender:
Remove old folder as first stage of calculations. The subroutines already automatically create folder(s) if they are missing.


samutoljamo-pareto (approved):
Looks good to me

PR #183, opened 2024-10-09
INTDEV-392: Tree view improvements

korva:
- Enable tree view items to be expanded without navigation
- Implement resizable tree view panel via react-resizable-panels
- Add optional progress chip to tree view
- Improve tree menu elements look&feel

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15043,6 +15043,15 @@
         }
       }
     },
+    "node_modules/react-resizable-panels": {
+      "version": "2.1.4",
+      "resolved": "https://registry.npmjs.org/react-resizable-panels/-/react-resizable-panels-2.1.4.tgz",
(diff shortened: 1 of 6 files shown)


samutoljamo-pareto (approved):
Looks good. Would it actually be better if expanding happened only when clicking the icon? Do you think there should be some kind of hint of the resizing feature?

korva:
> Looks good. Would it actually be better if expanding happened only when clicking the icon? Do you think there should be some kind of hint of the resizing feature?

I tried different variations of expand-on-click, and I think that it is good to let the user see the child cards when you select a card. I did disable card closing when clicked, that seemed off.

And now that you mentioned it, we talked with Vilja that maybe the first level of cards should be expanded on default when opening project. Now a single root card project will look a bit silly at first.

There could be a resize handle to show that resizing is possible. We basically copied the current model from Jira :)

PR #185, opened 2024-10-12
INTDEV-475 Allow creation of resource with long and short names

Fuzzbender:
Allow creation operations to use both long and short names:

`cyberismo create workflow test`
`cyberismo create workflow <prefix>/workflows/test`

Validates that all long name parts match the project prefix and to-be-created resource type.

Drop `local` name part usage from CLI. It is still used internally as a location for local project's resources, but it cannot be referred from CLI commands and shouldn't be visible, other than file paths.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -19,7 +19,7 @@ import { Edit } from './edit.js';
 import { ExportSite } from './export-site.js';
 import { Import } from './import.js';
 import { Move } from './move.js';
-import { pathExists, resolveTilde, sepRegex } from './utils/file-utils.js';
+import { pathExists, resolveTilde } from './utils/file-utils.js';
 import { Project } from './containers/project.js';
 import { Remove } from './remove.js';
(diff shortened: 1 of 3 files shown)


samutoljamo-pareto (approved):
Seems good, but getting the fullname requires a lot of similar code. I think the validateName function could actually just return a valid fullName. Then we could just demand the fullName elsewhere instead of having to handle both cases

Fuzzbender:
> Seems good, but getting the fullname requires a lot of similar code. I think the validateName function could actually just return a valid fullName. Then we could just demand the fullName elsewhere instead of having to handle both cases

The problem is that in INTDEV-468, the resource functions (exists, create, ...) have already been made shared across the resources. I didn't want to bring those changes here, instead I kept the existing "duplicating code" in place. I can ensure in PR for INTDEV-468  that `validateName` returns the long name - it is a good idea.

Fuzzbender:
Actually, I can make the suggested change here. Merging of INTDEV-468 will be painful anyways. :)

PR #190, opened 2024-10-14
INTDEV-489 Ensure that calculation folders are created

Fuzzbender:
Split this away from INTDEV-468 and fix it separately.


samutoljamo-pareto (approved):
Looks good to me

PR #192, opened 2024-10-15
INTDEV-498 Create links with direction (inbound|outbound)

Fuzzbender:
`direction` was not included when `createLink` was called.


samutoljamo-pareto (approved):
Looks good to me.

PR #196, opened 2024-10-16
Show card key / card type in metadata view

samutoljamo-pareto:
One issue with preview persists, I'll wait for #194 to get merged

### DIFF
--- a/tools/app/app/components/MetadataView.tsx
+++ b/tools/app/app/components/MetadataView.tsx
@@ -32,7 +32,7 @@ import EditableField, { EditableFieldProps } from './EditableField';
 
 interface FieldItemProps {
   expanded?: boolean;
-  control: Control;
+  control?: Control;
   defaultValue: MetadataValue | null;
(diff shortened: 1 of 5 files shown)


tools/app/app/components/MetadataView.tsx
```diff
@@ -27,19 +28,71 @@ import {
...
+        )}
+      </AccordionDetails>
+    </Accordion>
+  );
+}
 
 export interface MetadataViewProps {
   initialExpanded?: boolean;
   editMode?: boolean;
   metadata?: CardMetadata;
```
Fuzzbender, line 86: Not directly related to this PR, but I guess `metadata` could be mandatory in app. That would simplify some of the handling. I think app always retrieve metadata with the card information.

Fuzzbender (approved):

tools/app/app/components/MetadataView.tsx
samutoljamo-pareto: Yes, that would be a good idea

PR #205, opened 2024-10-18
INTDEV-470 Add App Basic acceptance test

njuvonen:
add test rest of testcases from https://cyberismo.atlassian.net/browse/INTDEV-470 and creating base-module to run them against

### DIFF
--- a/tools/app/.env.test
+++ b/tools/app/.env.test
@@ -1,1 +1,1 @@
-npm_config_project_path="../data-handler/test/test-data/valid/decision-records"
+npm_config_project_path="../../../cyberismo-bat"
--- a/tools/app/cypress/e2e/app.cy.ts
+++ b/tools/app/cypress/e2e/app.cy.ts
@@ -10,76 +10,137 @@ describe('Navigation', () => {
   beforeEach(() => {
     cy.visit('http://localhost:3000');
   });
-
-  it('opens project and displays project name and content', () => {
-    // User should be redirected to /cards root path
-    cy.url().should('include', '/cards');
-
-    cy.get('h4').contains('decision');
-    cy.get('p').contains('Decision Records');
+  it('Create new base module', () => {
+    // Creates a new base module with name Basic Acceptance Test
+    cy.exec('cd ../../../&&git clone git@github.com:CyberismoCom/module-base.git&&cyberismo create project "Basic Acceptance Test" bat cyberismo-bat');
+    cy.wait(1000); // short wait to avoid random failures
+    cy.exec('cd ../../../cyberismo-bat&&cyberismo import module ../module-base&&cyberismo create card base/templates/page');
+    cy.get('h4').contains('Basic Acceptance Test'); // Verify project name
+    cy.get('p').contains('Please select a card'); // Verify text on cards page
   });
 
-  it('selects a card, shows card content and actions', () => {
-    cy.visit('http://localhost:3000/cards/decision_6');
+  it('delete page and verify empty project', () => {
+    // Creates a new base module with name Basic Acceptance Test
+    cy.get('p').contains('Untitled page').click(); // Navigate to Untitled page in tree menu
+    cy.get('h1').contains('Untitled page'); // Verify Title in content area
 
-    // Check that content elements are visible
-    cy.get('h1').contains('Document Decisions with Decision Records');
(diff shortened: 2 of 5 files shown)


tools/app/cypress/e2e/app.cy.ts
```diff
@@ -89,19 +150,124 @@ describe('Modify project', () => {
...
+    cy.get('.MuiSelect-button').contains('Select Link Type').click(); // Click select link type button
+    cy.get('.Mui-expanded').contains('blocks').click(); // Select blocks
+
+    cy.get('.MuiSelect-button').contains('blocks'); // checks Select Link Type text changed to previously selected blocks option
+    cy.get('.MuiAutocomplete-root').get('[placeholder="Search card"]').click(); // click Search card field
+    cy.get('.MuiAutocomplete-option').contains('Untitled page').click(); // Select Untitled page
+    cy.get('button').contains('Add link').click(); // Click add link button
+
+    cy.get('.css-qmy8nh-JoyStack-root > .MuiStack-root > .MuiTypography-body-sm').contains('blocks');
+    cy.get('.css-qmy8nh-JoyStack-root > .MuiStack-root > .MuiTypography-title-sm').contains('Untitled page');
```
korva, line 191: Referring to UI elements this way breaks tests easily in the future if we do layout small layout changes etc. I would suggest to use the custom data-cy tag (as is done in some cases already) to mark the elements we want to interact with.

tools/app/.env.test
```diff
@@ -1 +1 @@
-npm_config_project_path="../data-handler/test/test-data/valid/decision-records"
\ No newline at end of file
+npm_config_project_path="../../../cyberismo-bat"
```
korva, line 1: This breaks API tests that have relied on the previously used data-handler test project. It's an annoying limitation that we can only have one project_path in a given env...can't think of a good workaround right now!

tools/app/cypress/e2e/app.cy.ts
```diff
@@ -10,76 +10,137 @@ describe('Navigation', () => {
   beforeEach(() => {
     cy.visit('http://localhost:3000');
   });
+  it('Create new base module', () => {
+    // Creates a new base module with name Basic Acceptance Test
+    cy.exec('cd ../../../&&git clone git@github.com:CyberismoCom/module-base.git&&cyberismo create project "Basic Acceptance Test" bat cyberismo-bat');
```
korva, line 15: This creates the base project outside of the cyberismo repo folder. That is a possible cause of conflicts. Maybe we could do this at the cyberismo repo root?

tools/app/cypress/e2e/app.cy.ts
```diff
@@ -10,76 +10,137 @@ describe('Navigation', () => {
...
+    cy.wait(1000); // short wait to avoid random failures
+    cy.exec('cd ../../../cyberismo-bat&&cyberismo import module ../module-base&&cyberismo create card base/templates/page');
+    cy.get('h4').contains('Basic Acceptance Test'); // Verify project name
+    cy.get('p').contains('Please select a card'); // Verify text on cards page
```
korva, line 19: I think the tests should clean up after themselves and delete the cyberismo-bat folder. Currently running tests twice fails because the folder already exists.

korva:
Few general comments:
- There have been many commits to main, would be good to rebase against main and resolve the conflicts
- For some reason (maybe the conflicts?) the tests were not run in CI. Need to test that too.

There are some tricky issues to solve here so we can keep all tests working (and have sensible test data), I'm up for a call to discuss alternatives!

tools/app/.env.test
korva: I tried fixing this by using a separate NODE_ENV for e2e tests (".env.test-e2e") which worked locally BUT caused the following warning:

You are using a non-standard "NODE_ENV" value in your environment. This creates inconsistencies in the project and is strongly advised against. Read more: https://nextjs.org/docs/messages/non-standard-node-env

...so that is not a clean solution. I'll think about this more.

korva:
I ran npm run prettier-fix to this branch so the CI Prettier check will pass. Let's see if the tests work in CI...

korva (approved):
Tests seem to work now reliably locally and in CI 👍

PR #211, opened 2024-10-21
Add separate class to handle paths for Project class

Fuzzbender:
Split away from original INTDEV-468 PR. 

A new class that handles resource paths for Project class. 
Encapsulates paths and simplifies Project. In the future (after few PRs) will be responsible for providing basic needs in resource creation. The new methods are not yet used.

### DIFF
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -16,7 +16,6 @@ import { mkdir, readFile } from 'node:fs/promises';
 import { spawnSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 
-// ismo
 import { Card, Link } from './interfaces/project-interfaces.js';
 import {
   copyDir,
@@ -62,10 +61,6 @@ export class Calculate {
     return pathExists(location) ? location : null;
   }
 
-  private getResourceFolder() {
-    return join(Calculate.project.calculationFolder, 'resources');
-  }
-
   private async generateCardTypes() {
     const cardTypes = await Calculate.project.cardTypes();
     const promises = [];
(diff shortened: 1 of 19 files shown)


korva (approved):
Looks good! Prettier has not been run on some of the changed files

tools/data-handler/src/calculate.ts
```diff
@@ -63,7 +62,7 @@ export class Calculate {
   }
 
   private getResourceFolder() {
```
samutoljamo-pareto, line 64: This could also be in the new class

tools/data-handler/src/calculate.ts
```diff
@@ -415,12 +414,12 @@ export class Calculate {
     const firstCard = cards[0];
     await this.setCalculateProject(firstCard); // can throw
     const cardTreeFile = join(
-      Calculate.project.calculationFolder,
+      Calculate.project.paths.calculationFolder,
```
samutoljamo-pareto, line 417: There's a lot joins in the calculate.ts file. Would it be more clear, if all these paths were provided by the new class?

samutoljamo-pareto (approved):
Very good change

tools/data-handler/src/containers/template.ts
```diff
@@ -90,7 +89,7 @@ export class Template extends CardContainer {
     parentCard?: Card,
   ): Promise<Card[]> {
     const templateIDMap: mappingValue[] = [];
-    const tempDestination = join(this.project.cardRootFolder, 'temp');
+    const tempDestination = join(this.project.paths.cardRootFolder, 'temp');
```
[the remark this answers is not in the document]
Fuzzbender: Yes, it would be good addition.

tools/data-handler/src/calculate.ts
Fuzzbender: I'll move it there.

tools/data-handler/src/calculate.ts
Fuzzbender: I'll check, if we good move it there as well.

PR #216, opened 2024-10-21
Use specified types instead strings when working with resources

Fuzzbender:
Add interfaces for `ResourceTypes` and `RemovableResourceTypes` (not all resources can be removed).
Use these instead of `strings` when handling resources.

This is another one of changes split away from INTDEV-468.

### DIFF
--- a/tools/data-handler/test/command-handler-remove.test.ts
+++ b/tools/data-handler/test/command-handler-remove.test.ts
@@ -274,19 +274,6 @@ describe('remove command', () => {
       );
       expect(result.statusCode).to.equal(400);
     });
-    it('remove() - try to remove unknown type', async () => {
-      const cardId = 'decision_5';
-      const calculateCmd = new Calculate();
-      const removeCmd = new Remove(calculateCmd);
-      await removeCmd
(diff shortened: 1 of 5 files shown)


tools/data-handler/test/command-handler-remove.test.ts
```diff
@@ -279,7 +282,11 @@ describe('remove command', () => {
       const calculateCmd = new Calculate();
       const removeCmd = new Remove(calculateCmd);
       await removeCmd
-        .remove(decisionRecordsPath, 'i-dont-exist', cardId)
+        .remove(
+          decisionRecordsPath,
```
samutoljamo-pareto, line 286: Should we actually remove this test? Now we're kind of testing what happens if a dev does not use the defined types.

samutoljamo-pareto (approved):
Looks good

tools/data-handler/test/command-handler-remove.test.ts
Fuzzbender: Yes. I can take it away. Let's have good tests only. :)

PR #217, opened 2024-10-22
INTDEV-501 Change resource collection functions signature

Fuzzbender:
Get rid of single boolean parameter in the resource collection functions (`Project.calculations()`) and friends).

Single boolean parameters cause confusion when reading the code. 
Additionally, provide means to collect resources from just modules.

Curiously, `calculations()` has different default parameter than the other ones. I didn't change this, as I didn't want to modify the calls that use default values. Maybe something to check a bit later.

BACKGROUND:
A simplified example of why single boolean parameters are bad for reading the code:

```
this.install(false);
```

There is no way of knowing what the parameter does unless you go and check the `install` function itself.
Compare this with: 

```
this.install(Install.skipValidation); 
```

You can immediately see that this call skips validation.

### DIFF
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -243,8 +243,8 @@ export class Calculate {
       Calculate.modulesFileName,
     );
     let modulesContent: string = '';
-    // Collect and return all available calculations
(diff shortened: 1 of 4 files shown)


samutoljamo-pareto (approved):
Looks good to me

PR #223, opened 2024-10-23
INTDEV-486 fix  move card dialog

samutoljamo-pareto:
When you moved the card via move card dialog, user was being redirected to the parent card and treemenu was not updated. Now it stays on the page and updates treemenu.


korva (approved):
Good fix!

PR #229, opened 2024-10-29
INTDEV-458 create stub for calculate

samutoljamo-pareto:
I ended up creating a stub for calculate in the create command. I did also try putting a manual delay to the handleNewCards function without the stub and I didn't have any issues for some reason(which was the initial concern)


Fuzzbender:
Stubbing with sinon is good solution for this. I think there was some ticket created. way back about taking sinon into use. 

Fix looks good to me.

Fuzzbender (approved):

PR #231, opened 2024-10-29
INTDEV-508 return root instead of the template name in case of a root template card

samutoljamo-pareto:
Now parentPath returns "root" also for parents of root template cards. Now it's more easier to know whether a template card is a root card or not

### DIFF
--- a/tools/data-handler/src/containers/card-container.ts
+++ b/tools/data-handler/src/containers/card-container.ts
@@ -51,7 +51,11 @@ export class CardContainer {
   // Finds parent
   private parentCard(cardPath: string) {
     const pathParts = cardPath.split(sep);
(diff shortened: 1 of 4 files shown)


tools/data-handler/src/containers/card-container.ts
```diff
@@ -51,7 +51,11 @@ export class CardContainer {
   // Finds parent
   private parentCard(cardPath: string) {
     const pathParts = cardPath.split(sep);
-    if (pathParts.at(pathParts.length - 2) === 'cardRoot') {
+    if (
```
Fuzzbender, line 54: No need to do in this PR, but.
We should implement proper name parsing for card paths (and resources too).

Fuzzbender (approved):

PR #234, opened 2024-10-30
INTDEV-527 Fix clingo parser

samutoljamo-pareto:
Removed condition of clingo parser, which caused empty strings to become undefined. This also happened only if the empty string was the last argument

### DIFF
--- a/tools/data-handler/test/utils/clingo-parser.test.ts
+++ b/tools/data-handler/test/utils/clingo-parser.test.ts
@@ -76,7 +76,7 @@ describe('ClingoParser', () => {
     expect(result.results[0].fieldName).to.equal(fieldValue);
   });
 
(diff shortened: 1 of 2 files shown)


tools/data-handler/test/utils/clingo-parser.test.ts
```diff
@@ -76,6 +76,13 @@ describe('ClingoParser', () => {
     expect(result.results[0].fieldName).to.equal(fieldValue);
   });
 
+  it('should parser field correctly when last argument is an empty string', async () => {
```
Fuzzbender, line 79: The test description does not make sense.
Maybe,  'should parser **handle** field correctly when last argument is an empty string'?

Fuzzbender (approved):

PR #236, opened 2024-10-31
INTDEV-274 Handle datatypes in calculations

samutoljamo-pareto:
They are almost the same as in the ticket, except:
* Didn't handle dates yet in any other way(They are in ISO Format, but we need to decide should we rather use a number)
* Persons, that was mentioned in the ticket, does not exist in the data model
* Strings, that was mentioned in the ticket, does not exist in the data model
* List: Currently, list is string separated by commas. Before we can interpret it as a multiple option field, we'd have to make changes to the UI


henryhaverinen:
I'm wondering if Boolean values should be constants or strings in the logic program. For some reason, the #true and #false values do not work here, so we have to use constants or strings (or integers).  It doesn't probably matter otherwise than in how natural the code looks. When we're generating terms that have the isEditable field then IIRC we chose to use strings "true" and "false". I wonder if constants true and false would look better in the code after all?

Fuzzbender (approved):
Looks good to me.

henryhaverinen:
Decisions in the sprint meeting: let's use constants `true` and `false` for Boolean values, and let's use integers for dates and datetimes based on the node.js/Python epoch.

tools/data-handler/src/calculate.ts
```diff
@@ -331,21 +231,22 @@ export class Calculate {
       this.project.paths.calculationFolder,
       Calculate.modulesFileName,
     );
-    let modulesContent: string = '';
     // Collect all available calculations
     const calculations = await this.project.calculations(ResourcesFrom.all);
 
+    const builder = new ClingoProgramBuilder();
```
Fuzzbender, line 237: `ClingoProgramBuilder` could be a private member of `calculate.ts`

tools/data-handler/src/utils/clingo-fact-builder.ts
```diff
@@ -0,0 +1,134 @@
+export type AllowedClingoType = string | number | boolean;
```
Fuzzbender, line 1: Missing copyright block.

tools/data-handler/src/utils/clingo-fact-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export type NestedBuilder = (builder: ClingoFactBuilder) => ClingoFactBuilder;
+
+export type ClingoArgument =
```
Fuzzbender, line 5: For maintenance purposes, it might be easier if this is `ClingoArgumentInterna | NestedBuilder` 
That way we need to add new values to `ClingoArgumentInternal` only.

tools/data-handler/src/utils/clingo-fact-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+  private arguments: ClingoArgumentInternal[] = [];
+
+  constructor(predicate: string, end: string = '.') {
+    this.predicate = predicate;
+    this.end = end;
+  }
+
+  /**
+   * Adds an argument, which is any value
+   * @param arg
```
Fuzzbender, line 41: some details missing; description for `arg`, return value.

tools/data-handler/src/utils/clingo-fact-builder.ts
```diff
@@ -0,0 +1,134 @@
...
+   * @returns
+   */
+  addArgument(arg: ClingoArgument | null): ClingoFactBuilder {
+    if (arg === null) {
+      return this;
+    }
+    if (typeof arg === 'function') {
+      const nestedBuilder = new ClingoFactBuilder('', '');
+      this.arguments.push(arg(nestedBuilder));
+    } else if (arg instanceof ClingoFactBuilder) {
```
Fuzzbender, line 51: This `else if` is not needed (does the same as `else`)

tools/data-handler/src/utils/clingo-fact-builder.ts
Fuzzbender: There some other instances of missing param descriptions and return values. I won't repeat the comment, but check the whole file.

tools/data-handler/src/utils/clingo-facts.ts
```diff
@@ -0,0 +1,260 @@
+import { sep } from 'path';
```
Fuzzbender, line 1: Let's always use `node` prefix when importing from `node`.

tools/data-handler/src/utils/clingo-facts.ts
```diff
@@ -0,0 +1,260 @@
...
+  Link,
+  LinkType,
+  Workflow,
+} from '../interfaces/resource-interfaces.js';
+import { ClingoProgramBuilder } from './clingo-program-builder.js';
+import { ClingoFactBuilder } from './clingo-fact-builder.js';
+import { Project } from '../containers/project.js';
+
+export const createWorkflowFacts = (workflow: Workflow) => {
+  const builder = new ClingoProgramBuilder().addFact('workflow', workflow.name);
```
Fuzzbender, line 15: Could the fact names be public string literals/constants?

tools/data-handler/src/utils/clingo-facts.ts
```diff
@@ -0,0 +1,260 @@
...
+        transition.toState,
+      );
+    }
+  }
+  return builder.buildAll();
+};
+
+export const createCardFacts = async (card: Card, project: Project) => {
+  // Small helper to deduce parent path
+  function parentPath(cardPath: string) {
```
Fuzzbender, line 56: No need to do in this PR, but can you make a tech debt ticket that we should make this a public generic function?

tools/data-handler/src/utils/clingo-facts.ts
```diff
@@ -0,0 +1,260 @@
...
+      'field',
+      fieldType.name,
+      'fieldDescription',
+      fieldType.fieldDescription,
+    );
+
+  builder.addFact('field', fieldType.name, 'dataType', fieldType.dataType);
+
+  if (fieldType.enumValues) {
+    let index = 1;
```
Fuzzbender, line 135: Clingo arguments are 1-based array? 
Or is there something "special" in the zero index?

tools/data-handler/src/utils/clingo-program-builder.ts
```diff
@@ -0,0 +1,57 @@
+import { ClingoArgument, ClingoFactBuilder } from './clingo-fact-builder.js';
```
Fuzzbender, line 1: Copyright block

tools/data-handler/test/utils/clingo-fact-builder.test.ts
```diff
@@ -0,0 +1,105 @@
+import { expect } from 'chai';
```
Fuzzbender, line 1: copyright block

tools/data-handler/src/calculate.ts
```diff
@@ -69,31 +76,8 @@ export class Calculate {
...
-      content += `field("${cardType.name}", "workflow", "${cardType.workflow}").\n`;
-
-      let index = 1;
-      for (const customField of cardType.customFields) {
-        content += `customField("${cardType.name}", "${customField.name}").\n`;
-        if (customField.displayName)
-          content += `field(("${cardType.name}", "${customField.name}"), "displayName", "${encodeClingoValue(customField.displayName)}").\n`;
-        if (customField.isEditable)
-          content += `field(("${cardType.name}", "${customField.isEditable}"), "isEditable", "${customField.isEditable}").\n`;
+      const content = createCardTypeFacts(cardType);
```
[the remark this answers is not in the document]
samutoljamo-pareto: Yes, but I'll create a tech debt ticket for restructuring these under project.

tools/data-handler/src/calculate.ts
samutoljamo-pareto: I don't think that makes sense, because then we'd have to create a reset method(so we can reuse it) and then the code won't be able to run asyncronously. We do use mutexes, but I don't think we should depend on them for this case

tools/data-handler/src/utils/clingo-fact-builder.ts
samutoljamo-pareto: Added to all

tools/data-handler/src/utils/clingo-fact-builder.ts
samutoljamo-pareto: Done

tools/data-handler/src/utils/clingo-fact-builder.ts
samutoljamo-pareto: Done

tools/data-handler/src/utils/clingo-fact-builder.ts
samutoljamo-pareto: done

tools/data-handler/src/utils/clingo-facts.ts
samutoljamo-pareto: done

tools/data-handler/src/utils/clingo-facts.ts
samutoljamo-pareto: Done

tools/data-handler/src/utils/clingo-facts.ts
samutoljamo-pareto: Yup

tools/data-handler/src/utils/clingo-facts.ts
samutoljamo-pareto: @henryhaverinen Is there a reason why we use 1-based indeces? I guess they are only used for ordering right now, so it cold technically be any number and still work just fine. But is there some other purpose for this?

tools/data-handler/src/utils/clingo-program-builder.ts
samutoljamo-pareto: done

tools/data-handler/src/utils/clingo-facts.ts
henryhaverinen: There is no reason in Clingo for the choice of 1-based indeces. I probably just first chose some kind of 1-based indexing in the query language for the select term: select(1, "title") for selecting on the top level of the results, and then used 1-based indexes in the next occurrence. We can change to 0-based indexes everywhere if you think that'd be better.

samutoljamo-pareto: This has no impact on functionality as of yet, so we can discuss this later, but I'll merge this ticket

PR #237, opened 2024-10-31
INTDEV-510: Use Clingo query to generate site/html/adoc export

korva:
- Use same query as in app to generate list of cards to show in site/html/adoc export. This guarantees the same content as in app and enables usage of calculated fields in site export
- Tree query results are translated to Card objects since the export functionality uses that internally
- Show progress in site tree menu
- Show additional metadata in site export content area

### DIFF
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -333,13 +333,7 @@ program
   .command('export')
   .description('Export a project or a card')
   .addArgument(
-    new Argument('<format>', 'Export format').choices([
-      'adoc',
-      'csv',
-      'html',
-      'pdf',
-      'site',
-    ]),
+    new Argument('<format>', 'Export format').choices(['adoc', 'html', 'site']),
   )
   .argument('<output>', 'Output path')
   .argument('[cardKey]', 'Path to card')
--- a/tools/data-handler/src/export-site.ts
+++ b/tools/data-handler/src/export-site.ts
@@ -27,6 +27,7 @@ import { errorFunction } from './utils/log-utils.js';
 import { Export } from './export.js';
 import { Project } from './containers/project.js';
 import { sortItems } from './utils/lexorank.js';
+import { Calculate } from './calculate.js';
 
 interface ExportOptions {
   silent: boolean;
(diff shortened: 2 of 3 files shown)


korva:
@henryhaverinen adoc and html exports use the "old" method (reading directories to memory) to compose the export data. Are they relevant, should we update those too to use same methods as site export?

Also, as a side note, CLI help states that pdf and csv exports are possible; both are not implemented.

henryhaverinen (approved):
Works as expected and the results look great on cyberismo-docs and the demo project

henryhaverinen:
> @henryhaverinen adoc and html exports use the "old" method (reading directories to memory) to compose the export data. Are they relevant, should we update those too to use same methods as site export?
> 
> Also, as a side note, CLI help states that pdf and csv exports are possible; both are not implemented.

Please remove the pdf and csv exports from the CLI help. The adoc and html export are still relevant, and I think we should use the same method. Would it make sense to do that as a part of this change, or should we just create another Jira ticket?

korva:
> > @henryhaverinen adoc and html exports use the "old" method (reading directories to memory) to compose the export data. Are they relevant, should we update those too to use same methods as site export?
> > Also, as a side note, CLI help states that pdf and csv exports are possible; both are not implemented.
> 
> Please remove the pdf and csv exports from the CLI help. The adoc and html export are still relevant, and I think we should use the same method. Would it make sense to do that as a part of this change, or should we just create another Jira ticket?

I think we can do them under this same ticket

korva:
OK, now all export features use the Clingo tree query as basis for gathering the data.
@samutoljamo-pareto can you take a look, do you approve the query usage and Card object creation afterwards?
And @Fuzzbender can you also check that I did not massacre any data-handler functionality!

Fuzzbender (approved):

korva:
> I think it would be better if didn't transform queries to cards, since that means we must maintain that function if we make changes. But maybe we could change that later, when we've gotten rid of the card depedency in the app itself

Agreed, once we have a solid model for the "query results with added non-queryable data" data structure, we should use that throughout the service. We can consider this a prototype :)

PR #238, opened 2024-11-04
INTDEV-531 do not mutate template name parameter

Fuzzbender:
Noticed this after some serious debugging when doing the "new command handler" (INTDEV-531).

The bug causes `Project` template names turn into to using short name format. Which leads to all sorts of nasty "not found" cases.

Ie. first templates are `[decision/templates/decision, decision/templates/simplePage, devision/templates/empty]`. 
Once 'decision' is instantiated, the templates then are `[decision, decision/templates/simplePage, devision/templates/empty]` and so on.


samutoljamo-pareto (approved):
Looks good to me

PR #244, opened 2024-11-04
DC-520 Provide a new class that has Commands and Project instance

Fuzzbender:
Add new `CommandManager` class that has one instance of each `Command` and of `Project`.
For now, `CommandManager` does not validate the inputs (will be done in a later PR). 
There is a single instance of `CommandManager`. A new instance can be created with `new` or by providing a new `path` to instance fetching.

Due to now having `Project` that is shared and kept after each command, it requires that the class will now handle adding and removing resources.

Better name for `CommandManager`?


tools/data-handler/package.json
```diff
@@ -16,7 +16,7 @@
     "clean": "rm -rf node_modules && rm -f package-lock.json && npm cache clean --force",
     "lint": "npx eslint .",
     "posttest": "npm run lint",
-    "test": "mocha --require mocha-suppress-logs --disable-warning=ExperimentalWarning --forbid-only \"./test/**/*.test.ts\"",
+    "test": "mocha  --disable-warning=ExperimentalWarning --forbid-only \"./test/**/*.test.ts\"",
```
Fuzzbender, line 19: Note to self: revert package.json change.

tools/data-handler/src/containers/project.ts
```diff
@@ -342,7 +374,9 @@ export class Project extends CardContainer {
     from: ResourcesFrom = ResourcesFrom.localOnly,
   ): Promise<Resource[]> {
     const moduleCalculations =
-      await this.collectResourcesFromModules('calculations');
+      from !== ResourcesFrom.localOnly
```
samutoljamo-pareto, line 377: The local resources won't be updated right? Now that constructor is only ran once, should we move `resourcesSync`-calls from constructor to these methods. Same issue with all other methods

tools/schema/cardTreeDirectorySchema.json
```diff
@@ -390,6 +390,7 @@
                   "description": "A logic program that only includes all the card-specific logic programs",
                   "type": "object"
                 },
+                "imports.lp": {},
```
samutoljamo-pareto, line 393: Why? I thought I removed imports.lp.

tools/data-handler/src/command-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (
+      CommandManager.instance &&
+      CommandManager.instance.project.basePath !== path
+    ) {
+      console.info(`Changing the CommandManager instance path to ${path}`);
+      console.info(CommandManager.instance.project.basePath);
+      CommandManager.instance = new CommandManager(path);
+    }
+    if (!CommandManager.instance) {
+      console.info('Creating a new CommandManager instance');
```
samutoljamo-pareto, line 85: This will be shown on the cli too, is this intentional?

samutoljamo-pareto (approved):
Looks good, but I left a few comments

tools/schema/cardTreeDirectorySchema.json
samutoljamo-pareto: Actually I think what this is about. Imports.lp is not generated anymore, but it is also not removed and that's why validation complains about it

Fuzzbender: Yeah, and it shouldn't be part of this PR, but otherwise tests won't pass.

tools/data-handler/src/command-manager.ts
Fuzzbender: I'll take it away, I was just trying to find if any extra instances are created.

tools/data-handler/src/containers/project.ts
Fuzzbender: True. I'd like to avoid scanning the disc for all of the calls for resources (e.g. call to `templates()`) 
There's the `addResource()` and `removeResource()` methods now. Preferably we should call "sync" only when there is a plethora of new potential resources; or we don't know the state.

Fuzzbender: I moved the resources collection to happen in a method `collectResources()`. It is now called from `Project` constructor and after renaming happens. Are there any other places that would require it?

korva (approved):
Good PR and things seem to work. Love that there is no need to drag the projectRoot everywhere.

CommandManager is an OK name, CommandHandler is another option, but let's go with CommandManager.

tools/data-handler/src/containers/project.ts
samutoljamo-pareto: Maybe not part of this PR, but would it be easier if we did all file system level logic inside Project? Now it's up to the user of project to remember to notify the project if there are changes to the state of the Project

Fuzzbender: Yes, it would be better to have file system notification logic. That way user cannot "sneak in" changes. 
We might have some old ticket about this. NodeJS is probably not the greatest platform for it, but we'll use what we have.

PR #245, opened 2024-11-04
INTDEV-533 Simplify Template constructor

Fuzzbender:
Remove extra `path` parameter as unnecessary. 
Make `Project` the first parameter, so that the parameter order is similar to `Commands`.

### DIFF
--- a/tools/data-handler/src/containers/template.ts
+++ b/tools/data-handler/src/containers/template.ts
@@ -64,10 +64,10 @@ export class Template extends CardContainer {
     },
   ];
 
-  constructor(path: string, template: Resource, project: Project) {
+  constructor(project: Project, template: Resource) {
     // Templates might come from modules. Remove module name from template name.
     const templateName = basename(template.name);
-    super(path, templateName);
+    super(template.path!, templateName);
 
     // prevent constructing a new project object, if one is passed to this class.
     this.project = project ? project : new Project(this.basePath);
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -454,7 +454,6 @@ export class Commands {
     }
     try {
       const addedCards = await this.commands?.createCmd.addCards(
-        path,
         cardTypeName,
         templateName,
         cardKey,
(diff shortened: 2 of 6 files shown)


tools/data-handler/src/containers/template.ts
```diff
@@ -64,10 +64,10 @@ export class Template extends CardContainer {
     },
   ];
 
-  constructor(path: string, template: Resource, project: Project) {
+  constructor(project: Project, template: Resource,) {
     // Templates might come from modules. Remove module name from template name.
     const templateName = basename(template.name);
-    super(path, templateName);
+    super(template.path!, templateName);
```
samutoljamo-pareto, line 70: Why is path optional in the first place?

samutoljamo-pareto (approved):
Looks good to me

tools/data-handler/src/containers/template.ts
Fuzzbender: I think - not sure - the reason is that in some cases when `Resource` is handled, the path is not known and not really necessary. And putting it into the instance of resource would be extra work with no benefits.

Fuzzbender: I'll make a ticket to make it mandatory (don't want to muddle this PR). 
In some cases when templates are created, it is not necessary to have `path` (as `Template` can figure it out itself. But we can always pass empty string and all is well.

Fuzzbender: https://cyberismo.atlassian.net/browse/INTDEV-535

samutoljamo-pareto: Could it be even more clear, if we had separate types for that:
ResourceInfo(what would a better name?) = Omit<Resource, 'path'>

Fuzzbender: `Resource` only has `name` and `path`, so in effect it would be `ResourceName`. 
`ResourceInfo` would be fine name if we expect `Resource` to be expanded in the future.

korva (approved):

PR #246, opened 2024-11-05
INTDEV-468 Make all resource folders optional

Fuzzbender:
Make all resource folders optional in the project.

Additionally, make file-based resources work use same implementation in `create.ts`. Remove resource-specific "exists" functions and provide one generic function. 
Additionally, split the "project interfaces" to two: project related interfaces and resource related interfaces. The latter will be moved to `Resource` class once we get to implement it. 
Finally, renamed the resource interface to be more or less similar; `Template` is an exception for now, as there is `Template` class that is exported.

### DIFF
--- a/tools/data-handler/src/interfaces/resource-interfaces.ts
+++ b/tools/data-handler/src/interfaces/resource-interfaces.ts
@@ -1,0 +1,148 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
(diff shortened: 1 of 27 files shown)


Fuzzbender:
There's at least one issue with the implementation; if you add card to a template, it logs: "Cannot find folder '/Users/samimerila/projects/test/.cards/local/calculations'". Need to investigate where it is coming from.

tools/data-handler/src/interfaces/resource-interfaces.ts
```diff
@@ -0,0 +1,148 @@
...
+ * in a correct place in folder structure.
+ */
+
+// Calculation does not have own metadata.
+export type CalculationMetadata = ResourceBaseMetadata;
+
+// Card type content.
+export interface CardType extends ResourceBaseMetadata {
+  workflow: string;
+  customFields?: CustomField[];
```
samutoljamo-pareto, line 27: Maybe not part of this PR, but we could consider making these required in typescript and add an empty array if the value is missing. It's much easier to work with empty arrays rather than undefined values

samutoljamo-pareto (approved):
Looks good to me

tools/data-handler/src/interfaces/resource-interfaces.ts
samutoljamo-pareto: Yes

samutoljamo-pareto: https://cyberismo.atlassian.net/jira/software/projects/INTDEV/boards/4?selectedIssue=INTDEV-543

PR #247, opened 2024-11-06
INTDEV-529 Prevent card moving to under itself

Fuzzbender:
In data-handler, prevent card being moved to itself. 

Earlier, when this was done, the operation was successful and as last sub-operation 'source' is deleted, so this led to card being deleted. 

Additionally, prevent 'root' from being moved. Earlier it would complain that 'root' is not found from the project.

### DIFF
--- a/tools/data-handler/src/move.ts
+++ b/tools/data-handler/src/move.ts
@@ -27,6 +27,9 @@ import {
 import { ActionGuard } from './permissions/action-guard.js';
 import { Calculate } from './calculate.js';
 
+// @todo - we should have project wide constants, so that if we need them, only the const value needs to be changed.
+const ROOT: string = 'root';
+
 export class Move {
   constructor(
     private project: Project,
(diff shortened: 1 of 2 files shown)


tools/data-handler/src/move.ts
```diff
@@ -36,12 +39,18 @@ export class Move {
...
    * Moves card from 'destination' to 'source'.
    * @param source source card to move
-   * @param destination destination card where source card will be moved to; or 'root'
+   * @param destination destination card where source card will be moved to; or to root
    */
   public async moveCard(source: string, destination: string) {
+    if (source === ROOT) {
+      throw new Error('Cannot move "root"');
+    }
+    if (source === destination) {
```
samutoljamo-pareto, line 48: Shouldn't we also check, that destination is not under the source card?

Fuzzbender: > Shouldn't we also check, that destination is not under the source card?
Yes, I'll add that as a check as well.

Fuzzbender: Fixed

tools/data-handler/src/move.ts
```diff
@@ -61,6 +70,10 @@ export class Move {
       throw new Error(`Card ${destination} not found from project`);
     }
 
+    if (destinationCard.path.includes(source)) {
```
samutoljamo-pareto, line 73: not in the scope of this, but: I think our path includes the full path on the system. Should we actually instead use path relative to project root? Having the whole system path might produce some weird bugs, if user has some folders, which have a meaning to us, higher up in the path

samutoljamo-pareto (approved):
Looks good to me.

tools/data-handler/src/move.ts
Fuzzbender: `Card` and `Resource` paths are indeed full paths in system. 
Both paths could be relative to the projectRoot.

Not going to change it here, but I'll make a JIRA ticket about it.

PR #249, opened 2024-11-06
INTDEV-460 Added improvements for calculations

samutoljamo-pareto:
Added all improvements as in the ticket. I didn't yet use them on the UI.

### DIFF
--- a/calculations/queries/card.lp
+++ b/calculations/queries/card.lp
@@ -1,2 +1,88 @@
-selectAll.
+
 result({{cardKey}}).
+
+% the second level includes metadata for the always visible and optionally visible fields
+
+field(Field, "visibility", "always") :-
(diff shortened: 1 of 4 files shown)


Fuzzbender:
No need to do in this PR. An idea: In the future when we have `Resource` class, maybe a helpful feature would be to have `toCalculation()` method which would return the resource content as calculation string. Then the functions introduced in this PR would be really short. That way, when you implement a new resource type, you'd implement almost automatically how it is used in calculations.

Fuzzbender (approved):

samutoljamo-pareto:
Good idea. Also, a helper class / function would be good for creating those strings. It is very easy to mess up the syntax, which requires quotes, commas and dots in the correct place

PR #251, opened 2024-11-07
INTDEV-465 When template card is added, generate rank for it

Fuzzbender:
When new card is added to a template (using `cyberismo add`) , add "rank" for the card.

**Note** that "rank" operation is not handled in this PR, see [INTDEV-545](https://cyberismo.atlassian.net/browse/INTDEV-545), this only affects "add".

Removed the `save()` calls to project configuration when new template cards were generated.


samutoljamo-pareto (approved):
Looks good but I think the loop is not needed

PR #252, opened 2024-11-07
INTDEV-509: Support macros in exports

korva:
- Run validateMacros for all content to be exported
- Report macros are shown as is, button macros are hidden
- Add valid macros to test data
- Improve html and adoc export layout


tools/data-handler/src/export.ts
```diff
@@ -169,10 +172,25 @@ export class Export {
       card.key,
     );
 
+    let asciidocContent = '';
```
Fuzzbender, line 175: asciidocContent -> asciiDocContent

Fuzzbender (approved):

PR #254, opened 2024-11-08
INTDEV-543 Replace optional arrays in interfaces with mandatory ones

Fuzzbender:
Mandatory arrays can be empty. 
When reading card or metadata from disk, and array values are missing, adds empty arrays in their place. 
This prevents from making changes to schemas, and to the content repositories.

### DIFF
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -191,6 +191,7 @@ export class Project extends CardContainer {
           key: entry.name,
           path: join(entry.parentPath, entry.name),
           children: [],
+          attachments: [],
         });
         await this.readCardTreeToMemory(
           join(entry.parentPath, entry.name),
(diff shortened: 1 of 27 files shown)


korva (approved):
Changes seem fine and not having null fields does make things cleaner in front end.

We do, however, lack this extra info that optional fields provide:
field: null -> feature is not supported or not available
field: [] -> feature is available but has no content

We can work around this in other ways of course when needed, just need to keep it in mind during API design.

tools/data-handler/src/containers/project.ts
```diff
@@ -190,6 +190,7 @@ export class Project extends CardContainer {
           key: entry.name,
```
Fuzzbender, line 190: Just a note to self: It would make a code slightly simpler, if there would be some sort of `static` key creator that wouldn't need to set the empty arrays.

E.g. `const card = Defaults.card(name, path);` would create 
```
{
  key: name,
  path: path, 
  children: [],
  attachments: []
}
```

PR #255, opened 2024-11-08
INTDEV-537: Updated UI for policy checks

korva:
- Two separate collapsible sections in sidebar
- Fix also TOC title to match designs


henryhaverinen (approved):
I tested the branch and the updated UI looks great and works great. Having successful checks collapsed by default is good, and it is also good to have the number of checks visible. I didn't look at the code.

PR #256, opened 2024-11-08
INTDEV-512 Update 'links' in card schema

Fuzzbender:
Define 'direction' for links. 
Make 'linkType' just a string. Validation will validate the content anyways.


tools/schema/cardBaseSchema.json
```diff
@@ -54,18 +54,15 @@
...
           },
+          "direction": {
+            "description": "Link direction; either towards this card, or outwards from this card.",
+            "type": "string",
+            "enum": ["outbound", "inbound"]
+          },
           "linkType": {
-            "description": "The link type is named according to the from-to direction. In user interfaces, the naming according to the link type is shown in the 'from' card.",
+            "description": "The link type is named according to the 'direction'. In user interfaces, the naming according to the link type is shown in the 'from' card.",
             "$comment": "The same link type names are used when indicating links in the index.json of 'to' cards, but the following reverse names shown in user interfaces: is blocked by, is caused by, has been instantiated to, is duplicate of, is mitigated by, is related to, is tested by",
```
henryhaverinen, line 64: This $comment should be removed. Currently, we're recording the links to the "from" card only, and there is no hard-coded list of link types, so the list "is blocked by, is caused by" etc. is not helpful

tools/schema/cardBaseSchema.json
```diff
@@ -54,18 +54,15 @@
...
             "pattern": "^[a-z]+_[0-9a-z]+$"
           },
+          "direction": {
+            "description": "Link direction; either towards this card, or outwards from this card.",
+            "type": "string",
+            "enum": ["outbound", "inbound"]
+          },
           "linkType": {
-            "description": "The link type is named according to the from-to direction. In user interfaces, the naming according to the link type is shown in the 'from' card.",
+            "description": "The link type is named according to the 'direction'. In user interfaces, the naming according to the link type is shown in the 'from' card.",
```
henryhaverinen, line 63: The description should be "The name of the link type. For example \"base/linkTypes/causes\"." because we are now using name link types instead of a hard-code enum.

tools/schema/cardBaseSchema.json
```diff
@@ -54,18 +54,15 @@
             "maxLength": 20,
             "pattern": "^[a-z]+_[0-9a-z]+$"
           },
+          "direction": {
```
henryhaverinen, line 57: To my understanding, we're not using a direction member, because links are recorded in the "from" card only, so the direction member should be removed from here and also from the "required" member.

henryhaverinen (changes requested):

tools/schema/cardBaseSchema.json
Fuzzbender: App is using 'direction' when it is handling link creation in UI.

henryhaverinen (approved):

PR #259, opened 2024-11-11
INTDEV-557 Separate resource handling to its own class

Fuzzbender:
Separate resource handling from `Project` to its own class.  
From now on, `Project` is basically a stack of cards (`CardContainer`) and `ResourceCollector` owner. 
This change is not visible to other parts of the sources; the changes to various commands is due to fixing the `cardType` api (INTDEV-548) in this PR as well.

### DIFF
--- a/tools/data-handler/src/containers/project/resource-collector.ts
+++ b/tools/data-handler/src/containers/project/resource-collector.ts
@@ -1,0 +1,374 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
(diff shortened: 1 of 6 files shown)


tools/data-handler/src/containers/project/resource-collector.ts
```diff
@@ -0,0 +1,359 @@
...
+   * @param from Defines where resources are collected from.
+   * @returns Array of resources.
+   */
+  public async resources(
+    type: string,
+    from: ResourcesFrom = ResourcesFrom.all,
+  ): Promise<Resource[]> {
+    const moduleResources =
+      from !== ResourcesFrom.localOnly
+        ? await this.addResourcesFromModules(type)
```
samutoljamo-pareto, line 353: Could we also store modules in memory? They are also easy to keep in sync, since only the import command updates them

Fuzzbender: Yes, we could. Only when import/remove is used the cached values would change. 
I'll include that change in this PR.

Fuzzbender: **Done** - now importing module calls `Project` to update the imported resources; who in turn calls `ResourceCollector` to make the updates.

PR #261, opened 2024-11-12
Regression fix: Resources cannot be created

Fuzzbender:
Three fixes: 

1) in `create.ts` always fetch the schema content. Otherwise, the creation is just silently skipped.
2) when fetching details of a particular resource, always look for both local and module content.
3) when validating a name in CLI, accept imported values to be used as parameters.

The first two are clear regression from INTDEV-468, but 3) has been there for a while.


Fuzzbender:
I'll do a separate ticket to address the pain points with unit tests. For now, I want to get this in main asap.

PR #263, opened 2024-11-12
INTDEV-539: Notifications

korva:
Added Notifications to app: general info/note shown to user, e.g. next steps to complete card process 
- Get notifications in card query
- Show UI element in sidebar

This PR also changes the TestSuite/TestCase fields used in policy checks to be Category/Title

### DIFF
--- a/calculations/common/queryLanguage.lp
+++ b/calculations/common/queryLanguage.lp
@@ -5,7 +5,7 @@
 % select(field): "field" should be returned on all levels of hierarchy
 % selectAll: all fields should be returned on all levels of hierarchy
 
-% select works similarly for "labels", "links", "policyChecks" and "deniedOperations" as it works for fields
+% select works similarly for "labels", "links", "notifications", "policyChecks" and "deniedOperations" as it works for fields
 
 select(N, Field) :- select(Field), resultLevel(_, N).
 selectAll(N) :- selectAll, resultLevel(_, N).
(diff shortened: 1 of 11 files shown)


calculations/common/queryLanguage.lp
```diff
@@ -93,6 +93,16 @@ showField(Key, Field) :-
     link(Source, Destination, LinkType),
     showAll(Destination).
 
+% notifications
+
+#show notification(Key, TestSuite, TestCase, Message) :
```
Fuzzbender, line 98: Why `TestSuite`, `TestCase`? Terminology doesn't make sense (this was already in earlier changes).

Fuzzbender (approved):

calculations/common/queryLanguage.lp
korva: @henryhaverinen any comments? As these are also connected to policy failures/successes, maybe they could be called "PolicyGroup" and "Policy" or something similar?

korva:
I'm gonna try writing some app tests for this, as policy checks and notifications are totally untested now

calculations/common/queryLanguage.lp
henryhaverinen: We have been reusing terms from JUnit, where test cases are organised in test suites. This probably makes sense for the policyCheckFailures and policyCheckSuccesses, but I agree that it is less natural with the notifications.

Notifications are advice or hints, so they are not necessarily related to policy. Should we keep the test suite and test case terms as they are for policy checks, but try to come up with better terms for notifications?  For example Category and Title? 

On the other hand, Category and Title would probably be more natural also for the policy checks.

henryhaverinen: We don't have a lot of policy check implementations in the modules yet, so I propose that we switch to Category and Title also for policy checks. (Doesn't have to be a part of this PR)

henryhaverinen:
> 1. Our BAT test data (based on moduleBase) does not contain policy checks or notifications. Should it? @henryhaverinen

We should. Should we now fork module-base and start adding usage for all the features we have. We should have all kinds of field types, different hiddenInTreeView cases, denied operations etc.

calculations/common/queryLanguage.lp
korva: It's done now, both Notifications and Policy checks now have category/title in use. I also updated the feature/henryhaverinen/INTDEV-539-notifications branch in unified-sdl repo

PR #264, opened 2024-11-13
Change static macros import to dynamic to avoid warning

korva:
This is to silence the "ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time" warning. It will still appear when the export command is used. This is a intermediate fix that should be removed once we find a better way to deal with this.


Fuzzbender (approved):

Fuzzbender:
The issue is going to go away since the next nodeJS release will contain a change that will remove the experimental warning.

PR #266, opened 2024-11-14
INTDEV-536 Push validation to commands. Simplify CommandHandler

Fuzzbender:
Push as much as possible of input validation to happen in commands themselves. This way both the `CommandHandler` and `CommandManager` both are using the same validation.

Additionally, change all sorts of random things in `CommandHandler`: 
- throw errors from internal methods and only catch them in `commands` method
- remove all the other layers of trap-harnesses
- exit process if no path has been set and cannot be figured out
- fix all command parameter orders to be similar for all commands
- do not pass `this` (or `this.something`) around, but use them directly in the methods
- some logging with stdout has been changed so that the messages are part of `requestStatus`

All references with "long/full name" have been replaced with "(valid) resource name". 
All references with "short name", or "long name"'s name , have been replaced with "identifier".

### DIFF
--- a/tools/data-handler/src/validate.ts
+++ b/tools/data-handler/src/validate.ts
@@ -12,7 +12,7 @@
 
 // node
 import { Dirent, readdirSync } from 'node:fs';
-import { dirname, extname, join, parse, resolve } from 'node:path';
+import { basename, dirname, extname, join, parse, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { readdir } from 'node:fs/promises';
 
@@ -30,6 +30,7 @@ import {
   Card,
   DotSchemaContent,
   ProjectSettings,
+  ResourceTypes,
 } from './interfaces/project-interfaces.js';
 import {
   CardType,
@@ -41,6 +42,10 @@ import {
   Workflow,
 } from './interfaces/resource-interfaces.js';
 
+const invalidNames = new RegExp(
+  '[<>:"/\\|?*\x00-\x1F]|^(?:aux|con|clock$|nul|prn|com[1-9]|lpt[1-9])$', // eslint-disable-line no-control-regex
+);
+
 import * as EmailValidator from 'email-validator';
 
 const baseDir = dirname(fileURLToPath(import.meta.url));
@@ -49,6 +54,9 @@ export interface LengthProvider {
   length: number;
 }
 
+/**
+ * Validates content.
+ */
 export class Validate {
   private static instance: Validate;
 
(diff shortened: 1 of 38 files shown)


korva:
When running tests locally, I get these two fails in export tests:

405 passing (5s)
  2 failing

  1) export command
       missing parent card:

      AssertionError: expected 200 to equal 400
      + expected - actual

      -200
      +400
      
      at Context.<anonymous> (file:///Users/jaakko/dev/cyberismo/cyberismo/tools/data-handler/test/export.test.ts:86:38)

  2) export command
       inaccessible destination:

      AssertionError: expected 200 to equal 400
      + expected - actual

      -200
      +400
      
      at Context.<anonymous> (file:///Users/jaakko/dev/cyberismo/cyberismo/tools/data-handler/test/export.test.ts:93:38)

korva:
These console prints:
console.info(`Changing the CommandManager instance path to ${path}`);
console.info(CommandManager.instance.project.basePath);
...in command-manager.ts mess up the test output a bit, are they needed?

Fuzzbender:
> These console prints: console.info(`Changing the CommandManager instance path to ${path}`); console.info(CommandManager.instance.project.basePath); ...in command-manager.ts mess up the test output a bit, are they needed?

Not really. I can remove them. They were useful when `CommandManager` was created. I'll remove the logging from these.

tools/data-handler/src/validate.ts
```diff
@@ -361,6 +369,19 @@ export class Validate {
     return errors;
   }
 
+  /**
+   * Validates that new name of a resource is according to naming convention.
+   * @param name: resource name
+   * returns true if name is valid, and false otherwise.
+   */
+  public static validateResourceNameIsCorrect(name: string): boolean {
```
samutoljamo-pareto, line 377: Quite a long function name, maybe "isValidResourceName" would suffice?

tools/data-handler/src/validate.ts
```diff
@@ -514,6 +548,44 @@ export class Validate {
...
   }
 
+  /**
+   * Validate that long and short resource names are valid.
+   * @param resourceType Type of resource
+   * @param resourceName Name of resource
+   * @param prefixes currently used project prefixes
+   * @returns resource name as valid resource name (long format); throws in error cases.
+   */
+  public async validResourceName(
```
samutoljamo-pareto, line 558: Should we call the long name "name" and short name "identifier" or something similar? I think it would be more clear and it would be clear what function we wanna use

samutoljamo-pareto (approved):
Good change, left a few comments

tools/data-handler/src/validate.ts
Fuzzbender: Good point. I'll open a discussion on Teams about this. Let's see what the hive mind says.

PR #276, opened 2024-11-15
INTDEV-576 Fix renaming crash

Fuzzbender:
When `Rename` emitted that data had been changed, it still used a project path parameter, which was removed when `CommandManager` was added. `Calculate` interpreted the parameter as a card key and it led to unhandled property reference. 

Fixed by removing the parameter from emit.

Additionally, added additional, related fixes:
- uncomment the code line that renames cardType `name` . I don't know why it was commented. It leaves the `name` with previous name unless it is changed. 
- rename also the `optionallyVisibleFields`

### DIFF
--- a/tools/data-handler/src/rename.ts
+++ b/tools/data-handler/src/rename.ts
@@ -181,13 +181,16 @@ export class Rename extends EventEmitter {
       true,
     );
     if (cardType) {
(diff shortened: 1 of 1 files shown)


korva (approved):
Looks good!

PR #283, opened 2024-11-18
Remove 'direction' from schema

Fuzzbender:
Remove unnecessary 'direction' from schema. 
There is no code changes required, as app deduces the direction from how the links are used.


samutoljamo-pareto (approved):
I think the current implementation decides the direction based on the card it is stored at

PR #285, opened 2024-11-19
INTDEV-569 Validate card changes correctly

Fuzzbender:
Change the order of operations around:
1. validate
2. make card changes (memory)
3. save

To: 
1. make card changes (memory)
2. validate
3. save

Additionally, tweak the `person` validation a bit. Pass a better validation error message forward.


samutoljamo-pareto (approved):
Looks good to me

PR #289, opened 2024-11-21
INTDEV-544 Integrate query with app

samutoljamo-pareto:
Now I've combined the cardQuery and card request into a single endpoint, which is used by the app. At the same, some of the logic has been moved to the queries:
* All fields are given by the query
* Links are given by the query
* Also removed use of fieldTypes/cardTypes/project where possible

I'll add a few more things, but you can already review it:
* rename parsed/content to better names
* rename results to fields/children

Some other notes:
* This is not very optimized so card query is a bit slow: logic programs are still generated anytime a request is made and clingo blocks the event loop
* Tree query and /cards could be also combined


henryhaverinen:
On my computer, the branch is not so slow. Some observations from testing:

- calculated fields now get values, but the values do not update correctly. You can try one of the risks in cisms, and there are two calculated fields (inherent risk and residual risk). If you change the likelihood and impact, the risk value does not update as expected. However, if you query the card using cyberismo calc run, the query results seem to show correct values.
- Even "calculated links" work. Of course data handler does not like it if you try to delete one of them
- Using module-test and the test case "Test all data types of custom fields", I set the Date field as "2024-01-01". After saving, the value was changed to 1969-12-20T00:33:03.488Z. Similar results with the dateTime field

samutoljamo-pareto:
@henryhaverinen Here's a copy of cisms_136, which has the calculated fields in the json as well:
{
    "title": "Supply chain attack",
    "cardType": "cisms/cardTypes/risk",
    "workflowState": "Treatment complete",
    "cisms/fieldTypes/owner": "sami.merila@cyberismo.com",
    "cisms/fieldTypes/inherentLikelihood": "high",
    "cisms/fieldTypes/inherentImpact": "veryHigh",
    "cisms/fieldTypes/inherentRisk": "veryHigh",
    "cisms/fieldTypes/mitigationStrategy": "mitigate",
    "cisms/fieldTypes/residualLikelihood": "veryHigh",
    "cisms/fieldTypes/residualImpact": "veryHigh",
    "cisms/fieldTypes/residualRisk": "moderate",
    "cisms/fieldTypes/lastTransitioned": "2024-07-09T06:42:05.115Z",
    "rank": "0|i",
    "lastUpdated": "2024-11-27T08:36:55.342Z",
    "lastTransitioned": "2024-11-27T07:25:12.222Z",
    "links": []
}

Since the calculated fields are present in the the json file, they are also generated to logic programs. After removing those, the calculated fields are calculated correctly

samutoljamo-pareto:
That test never breaks on windows, but it seems to break only in headless mode on linux/mac. This suggests that it has something to do with timing and it is more likely to happen if the runner is faster. I believe it broke, because I didn't do optimistic updates for the card query

tools/app/app/api/cards/[key]/route.tsx
```diff
@@ -240,11 +240,25 @@ async function getCardDetails(
...
+    if (card.length !== 1) {
+      return new NextResponse(`Query failed. Check card-query syntax`, {
+        status: 500,
+      });
+    }
+
     if (cardDetailsResponse) {
       return NextResponse.json({
-        ...cardDetailsResponse,
+        ...card[0],
```
Comment (korva) [inline tools/app/app/api/cards/[key]/route.tsx:258]: Update also the response changes to the swagger documentation at the start of this function

korva:
Looks good to me at this point, was not too slow either for me (M1 Macbook Pro, though). The mentioned upcoming naming changes are good, and the followup combination of /tree and /cards endpoints too!

tools/data-handler/src/utils/clingo-parser.ts
```diff
@@ -130,7 +130,7 @@ class ClingoParser {
           break;
         case 'date':
         case 'dateTime':
-          res[fieldName] = new Date(parseInt(decoded)).toISOString();
+          res[fieldName] = new Date(parseInt(decoded) * 1000).toISOString();
```
Fuzzbender, line 133: For safety it might be reasonable to provide the radix as a parameter (`parseInt(decoded, 10)`), this might avoid strange conversion issues in the future (somebody will eventually pass hex value here anyways). 
Is there any risk for overflow when multiplying the integer with 1000? What if it is already `maxInt` or close to it?

Fuzzbender (approved):
Looks good to me. I tested the branch; no more cypress errors. 
Well done.

tools/data-handler/src/utils/clingo-parser.ts
samutoljamo-pareto: No risk, because clingo can only handle 32-bit numbers.

PR #299, opened 2024-11-26
INTDEV-574 Rename links

Fuzzbender:
Currently the "new" types of information is not renamed, when project is renamed.
One of the non-supported types is "links". Add functionality to rename all links when project is renamed.

Once macros and reports are changed as well, then "rename" operation works once again.


tools/data-handler/src/rename.ts
```diff
@@ -74,18 +74,38 @@ export class Rename extends EventEmitter {
...
     }
     // Then rename the cards
     for (const card of sortedCards) {
       await this.renameCard(re, card);
     }
   }
 
+  // Rename card links.
+  private async updateCardLinks(re: RegExp, card: Card) {
+    if (!Project.isTemplateCard(card)) {
```
CyberJML, line 90: I would change this if to be opposite. 
if(Project.isTemplateCard(card)) {
  return;
}
rest of the code.

This way we don't have unnecessary nesting in the code.

tools/data-handler/src/rename.ts
```diff
@@ -74,18 +74,38 @@ export class Rename extends EventEmitter {
...
+      const links = card.metadata?.links ? card.metadata?.links : [];
```
korva, line 91: Could be simplified to:
const links = card.metadata?.links ?? []

korva (approved):
Looks good

tools/data-handler/src/rename.ts
Fuzzbender: Sure, why not.

PR #304, opened 2024-11-28
INTDEV-573, INTDEV-575, INTDEV-611 Rename expanded

Fuzzbender:
Add renaming support to card macros, card xrefs and to report's handlebar files.

Macros are in the card content, and we want to avoid making unnecessary changes to the users content. 
Thus, content is first searched with `{{{...}}}` `RegExp` and then inside each of matches, content is replaced with another `RegExp` (using just `<prefix>/`).

xrefs that link to project cards, are also renamed (e.g. `xref:cisms_70[...]` -> `xref:ismo_70[...]`).

Report renaming now also renames references in the handle bar files.

Card content file is only updated, if there are changes to it. And once for all of the various renaming needs (not once per change).

Additionally, noticed that attachment renaming was at some point probably got broken; content should **not** be validated in-between a renaming operation (since some resources have been updated, and others not). Fixed the case so that validation is not performed from rename updates to card content.

Finally, added some logging messages for future, to see what the renaming is doing. These will only be seen in CLI stdout (app shouldn't do renames for now):
Screenshot 2024-12-02 at 7 39 54

### DIFF
--- a/tools/data-handler/src/rename.ts
+++ b/tools/data-handler/src/rename.ts
@@ -23,6 +23,9 @@ import { resourceNameParts } from './utils/resource-utils.js';
 import { Template } from './containers/template.js';
 import { writeJsonFile } from './utils/json.js';
 
+// RE that helps find macros from card content.
+const macrosRe = /^{{{.*?}}}/gms;
+
 /**
  * Class that handles 'rename' command.
  */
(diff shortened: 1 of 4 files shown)


tools/data-handler/src/rename.ts
```diff
@@ -126,13 +151,37 @@ export class Rename extends EventEmitter {
       const copyLinkType = link.linkType.slice(0);
       link.cardKey = link.cardKey.replace(re, this.to);
       link.linkType = link.linkType.replace(re, this.to);
-      changed = copyCardKey !== link.cardKey || copyLinkType !== link.linkType;
+      changed =
+        changed ||
```
samutoljamo-pareto, line 155: ||= ?

tools/data-handler/src/rename.ts
```diff
@@ -126,13 +151,37 @@ export class Rename extends EventEmitter {
...
+        copyCardKey !== link.cardKey ||
+        copyLinkType !== link.linkType;
     });
     if (card.metadata && changed) {
       this.project.updateCardMetadata(card, card.metadata, true);
     }
   }
 
+  // Update card's macros
+  private async updateCardMacros(card: Card) {
```
samutoljamo-pareto, line 165: I think content! is a bit ugly. This function only depends on the content, so maybe just pass it?

samutoljamo-pareto (approved):
Tried and seems to work

tools/data-handler/src/rename.ts
Fuzzbender: Yes, the method could just return if there is no content

PR #310, opened 2024-12-02
INTDEV-546 Support ranking template cards

Fuzzbender:
Add full support to re-rank, re-balance, or setting first rank for template cards.
Module template cards cannot be affected.

### DIFF
--- a/tools/data-handler/src/move.ts
+++ b/tools/data-handler/src/move.ts
@@ -14,7 +14,7 @@
 import { join, sep } from 'node:path';
 
 import { copyDir, deleteDir } from './utils/file-utils.js';
-import { Card } from './interfaces/project-interfaces.js';
+import { Card, FetchCardDetails } from './interfaces/project-interfaces.js';
 import { Project, ResourcesFrom } from './containers/project.js';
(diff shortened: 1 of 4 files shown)


tools/data-handler/src/move.ts
```diff
@@ -244,32 +249,35 @@ export class Move {
...
 
     const children = sortItems(
-      await this.getChildren(card.parent || ROOT),
+      await this.getChildren(card),
       (item) => item.metadata?.rank || EMPTY_RANK,
     );
 
     if (!children || children.length === 0) {
-      throw new Error(`Children not found from card ${card.parent}`);
+      // no children --> go first
```
samutoljamo-pareto, line 268: The reason why it threw an error before was because it is not possible that there would be no children. We are fetching the children of the card's parent ie. it should always have at least the card itself.

samutoljamo-pareto (approved):
Looks good

tools/data-handler/src/move.ts
Fuzzbender: I'll revert this row.

PR #312, opened 2024-12-02
INTDEV-601 Enable tests that check resource removals

Fuzzbender:
BAT tests now successfully check that all resource types are removed.

Screenshot 2024-12-02 at 17 43 48

(I removed the console.log after taking the screenshot)


korva (approved):
Excellent

PR #313, opened 2024-12-03
INTDEV-603 CLI shows wrong name

Fuzzbender:
Fix the old name use.

Additionally fetch CLI project details from `package.json` instead of duplicating them in the `index.ts` file.
This normally shows `Experimental warning`, but I decided to suppress all warnings in CLI (we really don't need to show users deprecation, or experimental warnings when running CLI).
Finally, fixed some help texts in the CLI.

### DIFF
--- a/tools/cli/bin/run
+++ b/tools/cli/bin/run
@@ -2,6 +2,9 @@
 
 import program from '@cyberismocom/cli';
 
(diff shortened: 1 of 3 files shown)


korva (approved):
Good changes

henryhaverinen (approved):

PR #315, opened 2024-12-03
INTDEV-622 Safer way to launch 'cyberismo app'

Fuzzbender:
To remove `shell: true`, the parameters need to be passed as an array without string literals.
Also make `cwd` to use directly the value, instead of wrapping it to a string literal.

See more from: https://github.com/CyberismoCom/cyberismo/security/code-scanning/7
Note that copilot provided fix didn't work for us; the  `cyberismo app` (which leads to `npm start`)  would fail.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -653,11 +653,11 @@ export class Commands {
 
     // since current working directory changes, we need to resolve the project path
     const projectPath = resolve(this.projectPath);
+    const projectPathArg = `--project_path=${projectPath}`;
 
-    const args = [`start`, `--project_path="${projectPath}"`];
+    const args = [`start`, projectPathArg];
     execFileSync(`npm`, args, {
-      shell: true,
-      cwd: `${appPath}`,
+      cwd: appPath,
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto:
On windows: It seems to fail
```
C:\Users\Samu T\projects\cyberismo-isms>cyberismo app
Running Cyberismo app on http://localhost:3000/
Press Control+C to stop.
spawnSync npm ENOENT
```

If I add shell=true
```
C:\Users\Samu T\projects\cyberismo-isms>cyberismo app
Running Cyberismo app on http://localhost:3000/
Press Control+C to stop.
Command failed: npm start --project_path=C:\Users\Samu T\projects\cyberismo-isms
```

Npm is a path variable, but I think that is not enough, if we are starting the app without shell=true.

For the second one: Allthough I started it with shell=true, it fails because my file path has spaces

Fuzzbender:
In that case, there is very little to do. We could use `cwd` without the string literal, but that is not the main thing that the securityBot complains about. 

I guess I just close this PR and we'll live with the security issues.

samutoljamo-pareto (approved):

PR #318, opened 2024-12-03
INTDEV-610 Use 'es2023' as 'lib' value

Fuzzbender:
Since we already:
*  have TS 5.2.+ (we are using 5.6.3 at the moment)
* we have nodeJS 20+
* even ESLint supports `es2023` since 8.44; we are using 9.x+

--> We can move to "eslint2023". 

This gets us some convenience APIs that were not available earlier.

Haven't spotted any regression, as of yet.

### DIFF
--- a/tools/cli/tsconfig.json
+++ b/tools/cli/tsconfig.json
@@ -13,8 +13,8 @@
 
     "outDir": "./dist",
 
-    "target": "es2022",
-    "lib": ["es2020"],
+    "target": "es2023",
+    "lib": ["es2023"],
     "types": ["node", "mocha"]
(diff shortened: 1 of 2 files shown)


samutoljamo-pareto:
I think we could also increase the target. To my understanding, lib-option only tells that es2023 features are available so it will compile the code from ts to js like it would normally, but it leaves es2023 features as is without complaining. This means, that it shouldn't run on older versions of node that do not support es2023 even if target is es2022(unless we provide polyfills for the used functions).

samutoljamo-pareto (approved):

PR #320, opened 2024-12-04
INTDEV-595 do not try to collect fieldType information for predefined

ghost:
I defined predefined fields in a new constants.ts file and used toString method for converting the predefined types. This however means, that even dates are just passed as strings. However, since they are ISO strings anyway and we plan to do that also for custom fields, I dont think it's an issue

### DIFF
--- a/tools/data-handler/src/utils/constants.ts
+++ b/tools/data-handler/src/utils/constants.ts
@@ -1,3 +1,5 @@
+import { PredefinedCardMetadata } from '../interfaces/project-interfaces.js';
+
 export const INT32_MAX = 2147483647; // 2^31-1
 /**
     Cyberismo
@@ -21,4 +23,10 @@ export const PREDEFINED_FIELDS = [
   'workflowState',
   'lastUpdated',
   'lastTransitioned',
-];
+] satisfies (keyof PredefinedCardMetadata)[];
+
+export function isPredefinedField(
+  value: string,
+): value is keyof PredefinedCardMetadata {
+  return (PREDEFINED_FIELDS as string[]).includes(value);
+}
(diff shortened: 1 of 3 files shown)


tools/data-handler/src/utils/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+/**
+ * These are names of the fields that are non-custom fields that present in metadata
+ */
+export const PREDEFINED_FIELDS = [
```
Fuzzbender, line 16: Would it somehow be possible to use interface `CardMetadata` (from project-interfaces.ts) and make its members (not the array) content of this array?
That way when we add new members to the card metadata, they would be automatically added here.

Fuzzbender (approved):
Looks good to me. If we can re-use the existing interface that would be super good.

tools/data-handler/src/utils/constants.ts
samutoljamo-pareto: I'll take a look but I dont think it's possible with interfaces, because they do not exist in the runtime.

Fuzzbender: Ok. Then at least add comment to the interface description that if one expands it, remember to add `PREDEFINED_FIELDS` new value.

PR #321, opened 2024-12-04
INTDEV-593 Hide file extensions in resource handling

Fuzzbender:
Hide file extension when using Resources in CLI / App.


korva (approved):
All good!

PR #323, opened 2024-12-05
INTDEV-559: Scorecard macro

korva:
This is a work-in-progress PR for intermediate review, not ready for merging yet.

Here we introduce a new UI macro: scoreCard, which displays a value with added context.

It can be used in card Asciidoc content as follows:
`{{{ scoreCard '{"title": "SDL control adoption", "value": 30, "unit": "%", "legend": "In progress" }' }}}`

It can also be used inside other macros in project resources, such as the report macro Handlebars template, as follows (.hbs code):
```
{{#each results}}

  {{{scoreCard (scoreCardParams "Internal control adoption" this.[base/fieldTypes/progress] "%" )}}}

  {{{scoreCard (scoreCardParams "Security issues that require attention" this.[usdl/fieldTypes/issueCount] )}}}

{{/each}}
```
scoreCardParams is a Handlebars helper that creates a JSON string out of the given parameters. This was the only way I could come up with supporting macros inside other macros .hbs template 😅

App support for the macro also implemented, app will replace html placeholder with material/joy ui element.

**Open issue: how to support this in exports?**
If we want shared graphical representation as in app, we cannot use material elements at the moment.
Inline html+css would be possible as this is a simple element, but is not very maintainable.
Currently I implemented only a simple Asciidoc representation for export cases.

### DIFF
--- a/tools/schema/scoreCardMacroSchema.json
+++ b/tools/schema/scoreCardMacroSchema.json
@@ -1,0 +1,24 @@
+{
+  "type": "object",
+  "$id": "scoreCardMacroSchema",
+  "properties": {
+    "title": {
+      "type": "string",
+      "description": "Title text on top of the card (optional)"
+    },
+    "value": {
+      "type": "number",
+      "description": "Value shown as content of the card"
+    },
+    "unit": {
+      "type": "string",
+      "description": "Unit shown after the value (optional)"
+    },
+    "legend": {
+      "type": "string",
+      "description": "Description text shown below the value (optional)"
+    }
+  },
+  "additionalProperties": false,
+  "required": ["value"]
+}
--- a/tools/app/app/components/macros/ScoreCard.tsx
+++ b/tools/app/app/components/macros/ScoreCard.tsx
@@ -1,0 +1,78 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Box, Card, Typography } from '@mui/joy';
+import { MacroContext } from '.';
+import React from 'react';
+
+export type CreateCardsProps = {
+  title?: string;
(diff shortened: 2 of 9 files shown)


Fuzzbender:
As a very brief check, looks good. I will check this later again.

Some very general issues:
1. Validation should be taken into account in macros (not just scorecards).
If there is an error in the macro, validation should complain about it. This mainly means that the referred items in the macro are invalid; not so much invalid handlebar / HTML syntax.

2. Check that rename operation works with the changes. It should rename references to resources inside handlebars.

korva:
> 1. Validation should be taken into account in macros (not just scorecards)

@Fuzzbender do you mean that running `cyberismo validate` would also validate the macros in the project content? Currently macros are validated only runtime when they are evaluated.

korva:
We discussed that as the scoreCard macro has limited use cases as end-user macro (we don't have means & documentation on how to use e.g. card fields in the content, only static data can be added), we would treat this as an internal macro without documentation for now.

Its usage for now would be to provide a visually better UI element for report macros (inside report macro handlebars templates).

So, besides this PR, we would only need to update unified-sdl project (and any other content repo) report templates to use scoreCard instead of asciidoc.

Fuzzbender:
> > 1. Validation should be taken into account in macros (not just scorecards)
> 
> @Fuzzbender do you mean that running `cyberismo validate` would also validate the macros in the project content? Currently macros are validated only runtime when they are evaluated.

Yes. Macros cannot contain code that is only validated when it is about to be executed. At least syntax should be checked  that is conforms to whatever limitations we currently impose.

korva:
> > > 1. Validation should be taken into account in macros (not just scorecards)
> > 
> > 
> > @Fuzzbender do you mean that running `cyberismo validate` would also validate the macros in the project content? Currently macros are validated only runtime when they are evaluated.
> 
> Yes. Macros cannot contain code that is only validated when it is about to be executed. At least syntax should be checked that is conforms to whatever limitations we currently impose.

Upon thinking about this with Samu I suggest we make this macro validation support as a separate task. In order to do validation properly (e.g. run json validation and check for runtime errors) and support all cases (macros with different static and inject versions) we might want to introduce a new macro mode 'validate' which just runs the validation and throws errors (instead of producing Asciidoc content like the current two modes). This is not limited to scoreCard macro which is the theme of this PR.

tools/schema/scoreCardMacroSchema.json
```diff
@@ -0,0 +1,24 @@
+{
+  "type": "object",
+  "$id": "scoreCardMacroSchema",
+  "properties": {
+    "title": {
+      "type": "string",
+      "description": "Title text on top of the card (optional)"
```
Fuzzbender, line 7: No need to fix to this PR (unless you really want to), but there is no need to highlight that field is "optional". There is `required`  member that defines that.

Fuzzbender (approved):

PR #343, opened 2024-12-09
INTDEV-566: Update table header row background color in app and site export

korva:
PRs for content repos to be created to remove manual style declarations from Asciidoc content.

### DIFF
--- a/resources/ui-bundle/css/cyberismo.css
+++ b/resources/ui-bundle/css/cyberismo.css
@@ -37,3 +37,8 @@ footer a {
 .nav-panel-explore .context {
     display: none;
 }
+
+/* Table header row */
+.doc table.tableblock th {
+  background: #fbfcfe;
+}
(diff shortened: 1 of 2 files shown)


henryhaverinen (changes requested):
I tried this with the following table:
```
[width="100%",cols="50%,50%",options="header",]
|===
|Role |Person
|Product responsible |
|Architect / Lead developer |
|Developer |
|Test engineer |
|Security specialist |
|User manual technical writer |
|===
```

The result looks like this:

image

For some reason, the gray background is applied not just to the header row but to all rows.

Another example without a header row is completely gray as well:

```
[cols="1,1"]
|===
|Cell in column 1, row 1
|Cell in column 2, row 1

|Cell in column 1, row 2
|Cell in column 2, row 2

|Cell in column 1, row 3
|Cell in column 2, row 3
|===
```
Looks like this:

image

korva:
Ah yes, slight mistake in the CSS selector, fixed now

henryhaverinen (approved):
Now it is looking good!

korva:
I'll merge this and create PRs for content repos next

PR #344, opened 2024-12-10
INTDEV-632 - Add template search for create new card popup

CyberJML:
Small adjustment to current NewCardModal to show only filtered/searched results. 
Default view is currently:
image

The search box is using autofocus to make the usage a bit more easier. The autofocus returning after it has been cleared did not work automatically, not even when I added inputRefs, most likely since the content is generated after the filter has been cleared. 

Search results are always parsed with .toLowerCase(), like:
image

This will parse the template names and categories

### DIFF
--- a/tools/app/app/components/modals/NewCardModal.tsx
+++ b/tools/app/app/components/modals/NewCardModal.tsx
@@ -26,7 +26,10 @@ import {
   ModalClose,
   CardOverflow,
   Grid,
+  Input,
+  IconButton,
 } from '@mui/joy';
+import ClearIcon from '@mui/icons-material/Clear';
 import { useTranslation } from 'react-i18next';
 import { useCard, useTemplates } from '@/app/lib/api';
 import { useAppDispatch } from '@/app/lib/hooks';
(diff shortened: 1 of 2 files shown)


korva:
Nice feature and very useful as is! Few comments:

- The modal dialog size keeps changing based on number of search results, which is kinda annoying. Would be good if we could keep the dialog size unchanged.

- I would also add a "No templates found" empty state

- The dialog also retains state after closing, i.e. the previous search term remains in the text input. Maybe clear that when dialog closes?

CyberJML:
I added some changes to keep the Modal same size, even though the results change. Not sure if % scaling is the correct way to do this. This introduced couple new problems. 
- There's unnecessary X axis scrollbar in the view, so I disabled it. 
- The buttons at the bottom of the Modal moved with the content, so I anchored them to the bottom with auto margin
- onClose was fed from outside of the Modal, so I created new handleClose function to do that + clear the state
- I don't like prettier, hard to get out the actual problems and I should just trust the --write?

korva:
The dialog size fix and clear-upon-close work nicely, great!
Two small nitpicks:

- Now the list shows "No templates available" once per category, which can lead to the search result not fitting on screen (see screenshot). I think it would work better if we show the "No templates available" only when there are no results at all and otherwise just hide the categories without results (if there are some categories with results to show).
Screenshot 2024-12-11 at 11 47 27

- If you select a template and then start searching, the selected template can be hidden but it is still selected (and you can create a card based on it). Maybe clear the selection once search is updated? This is a minor issue.

korva (approved):
All good now! 👍

PR #345, opened 2024-12-10
INTDEV-582, INTDEV-583 Change resources

Fuzzbender:
First version on how to update resources. This can be used to play around with the functionality in CLI.

This now only supports changing the `name` of a file-based resource (INTDEV-582) or changing `workflow` of a card type.
If workflow's name changes, it can automatically update all the referring cardTypes (but not calculations, macros or other referring elements). I am not going to implement the missing renames for now.

This is how it now looks:
Screenshot 2024-12-10 at 14 31 52

In the near future I will replace this implementation with one that uses "resource classes". 
The CLI api and tests probably remain as-is. Why you might ask. 
Reason is that handling all intricacies of a specific resource in one "update command" class becomes nightmarish to update. It is a lot easier if we introduce a class per resource type and that class then handles its own exceptions, if any.

### DIFF
--- a/tools/data-handler/src/update.ts
+++ b/tools/data-handler/src/update.ts
@@ -1,0 +1,117 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Project, ResourcesFrom } from './containers/project.js';
+import { Validate } from './validate.js';
+import {
+  resourceNameParts,
+  resourceNameToString,
+  ResourceName,
+} from './utils/resource-utils.js';
+import { CardType, Workflow } from './interfaces/resource-interfaces.js';
+
+/**
+ * Class that handles 'update' commands.
+ */
+export class Update {
+  constructor(private project: Project) {}
+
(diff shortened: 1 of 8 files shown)


henryhaverinen:
I tried this in cisms with
```
cyberismo update cisms/fieldTypes/mitigationStrategy name cisms/fieldTypes/strategy
```
The tool responded "Done", but git status didn't show any changes

henryhaverinen:
With the new changes, I was able to rename a fieldType. The cardTypes that used the fieldType were not updated, however. They still use the old name. I suppose that is expected?

Fuzzbender:
> With the new changes, I was able to rename a fieldType. The cardTypes that used the fieldType were not updated, however. They still use the old name. I suppose that is expected?

Yes, the only automatic thing it does is when changing the name of a workflow, it also updates all of the cardTypes that use the workflow.

tools/data-handler/src/update.ts
```diff
@@ -0,0 +1,132 @@
...
+    resourceName: ResourceName,
+    key: string,
+    value: unknown,
+  ) {
+    const resource = await this.resource(resourceName);
+    if (resource && resource.name) {
+      if (key === 'name') {
+        resource.name = value as string;
+      } else if (key === 'workflow') {
+        (resource as CardType).workflow = value as string;
```
samutoljamo-pareto, line 44: Could consider making a typescript type, which maps ResourceName to a a value type. Then this function could be generic

Fuzzbender: Can you expand on that? `ResourceName` is just `name` (prefix/type/identifier) and `path`

samutoljamo-pareto: I thought something like this would work, but looks like the compiler did not manage to figure out the type in the if block:
```
type Resource = 'workflow' | 'fieldtype';

export interface ResourceName<T extends Resource> {
  prefix: string;
  type: T;
  identifier: string;
}

export type ValueMap = {
  workflow: string;
  fieldtype: number;
};

function func<T extends Resource>(name: ResourceName<T>, value: ValueMap[T]) {
  if (name.type === 'fieldtype') {
    // value is now number
  } else {
    // value is now string
  }
}
```

PR #346, opened 2024-12-10
INTDEV-607 implement new query language features

samutoljamo-pareto:
Features:

The child results in the query language are generalised: there could be multiple types of child results, with different collection names, yielding different array objects as children of the result.

The data type of fields should be included in the query results. Query language could produce 4-member field terms that include the data type, and a more complex term for enums and lists. QueryParser does not depend on project anymore.

For enum and list data types, the results should be objects that include the value, display value and index of the enum value

When sorting by an enum field, used the index property of the enum value.

When sorting by an list field, used the smallest index property of the list value.

Also changed results->children in tree query and results->fields  and results->enumValues in cardQuery

### DIFF
--- a/calculations/common/base.lp
+++ b/calculations/common/base.lp
@@ -9,21 +9,10 @@ ancestor(A, C) :- parent(A, B), ancestor (B, C), card(A), card(B), card(C).
 % if the card type is given, then it's a card
 card(C) :- field(C, "cardType", _).
 
-% data types for the default card fields
-dataType(Card, "cardType", "shortText") :-
-    field(Card, "cardType", _).
-
-dataType(Card, "title", "shortText") :-
-    field(Card, "title", _).
+% data types for non-shortText default values
 
 dataType(Card, "lastUpdated", "dateTime") :-
     field(Card, "lastUpdated", _).
-
-dataType(Card, "workflowState", "shortText") :-
-    field(Card, "workflowState", _).
-
-dataType(Card, "rank", "shortText") :-
-    field(Card, "rank", _).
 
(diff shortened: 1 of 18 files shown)


samutoljamo-pareto:
@henryhaverinen I made some changes to queryLanguage. Inside the "fields"-collection, "value"-field contains the value. However, if we set the dataType to "enum", it will fail, because #enumValue assumes that the key is the card and field is the field. Here, Key is the field and field is "value". In card query, I overwrote the dataTypes of enum/lists to "shortText".

We didn't talk about how those enums / lists should represented normally. I used {
    value: "enumValue",
    displayName: "enum display name",
    index: 3
}.

I didn't implement the same for lists yet, since list means a string, that separates the items by commas

henryhaverinen:
I tested this and it is working pretty much as expected. For example getting enum values and display values works in reports. As we talked separately:
- sorting by an enum field should be done based on the index
- let's call the enum display value `displayValue` rather than `displayName`, to be consistent with the fieldType schema

samutoljamo-pareto:
All changes now should be made. listField is not finished, since it will implemented in INTDEV-623 differently

henryhaverinen:
This seems to work great otherwise, but ordering by an enum field is not working yet. When ordering by an enum, we should order by the index value. You can find an example in https://github.com/CyberismoCom/unified-sdl/tree/feature/henryhaverinen/INTDEV-607-testing, see the bottom of the card usdl_bljf0j6k, which is based on the threatList report

List data type can be finalised later as part of INTDEV-623.

Fuzzbender (approved):

PR #348, opened 2024-12-12
INTDEV-541 Add support for labels in cli

samutoljamo-pareto:
Added support for the following commands:
cyberismo create label <cardId>
cyberismo remove label <cardId> [label name]
cyberismo show labels

### DIFF
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -222,8 +222,8 @@ create
 create
   .command('label')
   .description('Create a label')
-  .argument('<cardKey>')
-  .argument('<label>')
+  .argument('<cardKey>', 'Key of the card that the label will be added to')
+  .argument('<label>', `Name of the created label. ${nameGuideline}`)
   .option('-p, --project-path [path]', `${pathGuideline}`)
   .action(async (cardKey: string, label: string, options: CardsOptions) => {
     const result = await commandHandler.command(
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -220,6 +220,20 @@ create
   });
 
 create
+  .command('label')
+  .description('Create a label')
+  .argument('<cardKey>')
+  .argument('<label>')
+  .option('-p, --project-path [path]', `${pathGuideline}`)
+  .action(async (cardKey: string, label: string, options: CardsOptions) => {
+    const result = await commandHandler.command(
+      Cmd.create,
+      ['label', cardKey, label],
+      options,
+    );
+    handleResponse(result);
+  });
+create
   .command('link')
   .description('Create a link')
   .argument('<source>', 'Source card key of the link')
@@ -536,6 +550,20 @@ remove
   });
 
 remove
+  .command('label')
+  .argument('<cardKey>', 'Source card key of the link')
+  .argument('[label]', 'Label being removed')
+  .option('-p, --project-path [path]', `${pathGuideline}`)
+  .action(async (cardKey: string, label: string, options: CardsOptions) => {
+    const result = await commandHandler.command(
+      Cmd.remove,
+      ['label', cardKey, label],
+      options,
+    );
+    handleResponse(result);
+  });
+
+remove
   .command('link')
   .argument('<source>', 'Source card key of the link')
   .argument('<destination>', 'Destination card key of the link')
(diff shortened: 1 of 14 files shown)


tools/cli/src/index.ts
```diff
@@ -219,6 +219,20 @@ create
     handleResponse(result);
   });
 
+create
+  .command('label')
+  .description('Create a label')
```
Fuzzbender, line 224: Add descriptions for command parameters (just like you did with 'remove')

tools/cli/src/index.ts
```diff
@@ -535,6 +549,20 @@ remove
...
 
+remove
+  .command('label')
+  .argument('<cardKey>', 'Source card key of the link')
+  .argument('[label]', 'Label being removed')
+  .option('-p, --project-path [path]', `${pathGuideline}`)
+  .action(async (cardKey: string, label: string, options: CardsOptions) => {
+    const result = await commandHandler.command(
+      Cmd.remove,
+      ['label', cardKey, label],
```
Fuzzbender, line 560: Add 'label' to `RemovableResourceTypes`

tools/cli/src/index.ts
```diff
@@ -642,7 +670,7 @@ program
   .description('Shows resource types in a project')
   .argument(
     '<type>',
-    'resource types: attachments, card, cards, cardType, cardTypes, linkType, linkTypes, project, reports, template, templates, workflow, workflows',
+    'resource types: attachments, card, cards, cardType, cardTypes, labels, linkType, linkTypes, project, reports, template, templates, workflow, workflows',
```
Fuzzbender, line 673: Is there any point to show details of one type? Probably not if it is just one string.

tools/data-handler/src/create.ts
```diff
@@ -408,6 +408,23 @@ export class Create extends EventEmitter {
     await this.createResource('fieldType', content);
   }
 
+  public async createLabel(cardKey: string, label: string) {
```
Fuzzbender, line 411: 'label' should follow generic naming rules.
At least not empty string, must be at least N characters long and no longer than M characters. 
Should not contain extra characters.
There's at least `isValidResourceName()` function that could be used. It was intended for resource identifiers, but labels could follow similar rules.

tools/data-handler/src/show.ts
```diff
@@ -269,6 +269,28 @@ export class Show {
...
+      }
+      // Recursively collect labels from subcards, if they exist
+      if (card.children) {
+        labels.push(...this.collectLabels(card.children));
+      }
+      return labels;
+    }, []);
+  };
+
+  public async showLabels(): Promise<string[]> {
```
Fuzzbender, line 286: Add comment block to public API.

tools/data-handler/test/command-handler-create.test.ts
```diff
@@ -392,6 +392,23 @@ describe('create command', () => {
...
+  it('create a label (success)', async () => {
+    const result = await commandHandler.command(
+      Cmd.create,
+      ['label', 'decision_6', 'test'],
+      options,
+    );
+    expect(result.statusCode).to.equal(200);
+  });
+
+  it('try create a label - label exists', async () => {
```
Fuzzbender, line 404: Add tests for invalid names (empty strings + if you add any other limitations; length, invalid characters).

Fuzzbender (approved):
Looks good. Some minor comments.

As labels can be added to template cards; then that template can be a in a module for another project. Thus, another project can receive labels from modules. 
Is this ok, or not? 
If yes - then labels should be collected also from template cards.
If not - then disallow setting a label to a template card.

Fuzzbender:
> 
>If yes - then labels should be collected also from template cards.
Discussed with @henryhaverinen. Labels should be possible in template cards. Thus, ensure that the labels are also collected (in show) for templates. Perhaps add a test where unit test adds a label to a template card as well.

tools/cli/src/index.ts
samutoljamo-pareto: Maybe if you wanted a list of all cards with a specific label.

tools/cli/src/index.ts
samutoljamo-pareto: done

tools/data-handler/test/command-handler-create.test.ts
samutoljamo-pareto: added those tests

samutoljamo-pareto:
@Fuzzbender I noticed that the import from CSV, functionality overwrites the template labels, if they are defined. If they are not in the csv, labels are used from the template. I think I'll change it so that it also simply adds labels so it will have both the labels defined in the templates and in the csv

PR #366, opened 2024-12-17
INTDEV-637: Validate macros

korva:
- Introduce a new mode for macro evaluation: validate, which only runs validation for macro content and produces no output.
- Validation mode checks that necessary parameter are available in correct format and throws errors if not. Other modes (inject & static) produce Asciidoc error output in case of errors.
- Use this mode during `cyberismo validate `command to validate macros inside project card contents
- Update macro tests to use validation mode and remove old validateMacros function, which was only used by tests

There is a remaining issue where the following warning is printed to CLI output:
`ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time`
This is probably because of macro JSON schemas that are being loaded with import in code. I do not have a fix for this issue yet.

### DIFF
--- a/tools/data-handler/src/macros/createCards/index.ts
+++ b/tools/data-handler/src/macros/createCards/index.ts
@@ -27,22 +27,28 @@ class CreateCardsMacro extends BaseMacro {
   constructor() {
     super(macroMetadata);
   }
+
+  handleValidate = (data: string) => {
+    this.validate(data);
+  };
+
   async handleStatic() {
     // Buttons aren't supported in static mode
     return '';
   }
 
   handleInject = async (_: MacroGenerationContext, data: string) => {
+    const options = this.validate(data);
+    return createHtmlPlaceholder(this.metadata, options);
+  };
+
+  private validate(data: string): CreateCardsOptions {
     if (!data || typeof data !== 'string') {
       throw new Error('createCards macro requires a JSON object as data');
     }
-    const options = validateMacroContent<CreateCardsOptions>(
-      this.metadata,
-      data,
-    );
 
-    return createHtmlPlaceholder(this.metadata, options);
-  };
+    return validateMacroContent<CreateCardsOptions>(this.metadata, data);
+  }
 }
 
 export default CreateCardsMacro;
--- a/tools/data-handler/src/macros/index.ts
+++ b/tools/data-handler/src/macros/index.ts
@@ -48,10 +48,19 @@ export function validateMacroContent<T>(
   if (!macro.schema) {
     throw new Error(`Macro ${macro.name} does not have a schema`);
   }
-  return validateJson<T>(JSON.parse(data), {
-    schemaId: macro.schema,
-    validator,
-  });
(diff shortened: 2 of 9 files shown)


tools/data-handler/src/macros/createCards/index.ts
```diff
@@ -27,6 +27,15 @@ class CreateCardsMacro extends BaseMacro {
   constructor() {
     super(macroMetadata);
   }
+
+  handleValidate = (context: MacroGenerationContext, data: string) => {
+    if (!data || typeof data !== 'string') {
+      throw new Error('createCards macro requires a JSON object as data');
+    }
```
samutoljamo-pareto, line 34: handleInject could call this.handleValidate instead of repeating the logic twice

tools/data-handler/src/macros/index.ts
```diff
@@ -167,57 +172,36 @@ export async function evaluateMacros(
...
       for (const macro of macroInstances) {
-        result = await macro.applyMacroResults(result);
+        result = await macro.applyMacroResults(result, context);
       }
       if (macroCount(result) === 0) {
         break;
       }
     } catch (err) {
-      return handleMacroError(err, emptyMacro);
+      return handleMacroError(err, emptyMacro, context);
```
samutoljamo-pareto, line 181: Could probably leave a comment since handleMacroError will throw on validate and otherwise return asciidoc

samutoljamo-pareto (approved):

Fuzzbender:
"There is a remaining issue where the following warning is printed to CLI output:
ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
This is probably because of macro JSON schemas that are being loaded with import in code. I do not have a fix for this issue yet."

If you use dynamic import then at least, then CLI can filter out the warnings.
A sort of kludge...

samutoljamo-pareto:
image

As of v23.1, importing JSON modules is stable. v23 is the "current" release and will never be LTS, but v24 will be released as LTS October 2025. I wouldn't be worried that they'd change during v22(current LTS, what we use), but the warning is of course annoying

Fuzzbender:
> image
> 
> As of v23.1, importing JSON modules is stable. v23 is the "current" release and will never be LTS, but v24 will be released as LTS October 2025. I wouldn't be worried that they'd change the feature during v22(current LTS, what we use), but the warning is of course annoying

It is really annoying, as it is shown with every CLI operation.

samutoljamo-pareto:
--no-warnings=ExperimentalWarning should work on v21 and v22

tools/data-handler/src/macros/createCards/index.ts
korva: I added a private validate() -function that can be reused in different modes. Validation returns a class-specific `options` object so I did not do this at the class interface level.

tools/data-handler/src/macros/index.ts
korva: Added

PR #370, opened 2024-12-19
INTDEV-623 Update the Metadata fields according to UI concept

samutoljamo-pareto:
Things to review:
* I used the default datepicker or do we want to use something else? MUI's datepicker is very close to the design, but it means we must add mui back as dependency for app
* I used a tooltip for the fieldDescription, but design stated "description is shown if the user clicks the icon" --> We can make the tooltip clickable, but do we want to?
* "Change the metadata editing pen icon as the hand icon" --> I used 'pointer' if that was what was meant by "hand icon"


henryhaverinen:
This is looking very good! A couple of comments:

- Yes, I think the "hand icon" refers to pointer
- the solution for fieldDescription is good in my opinion (that the description shows when you hover over the info icon)
- the date and date time pickers are fine
- In edit mode, the current values of date and datetime fields are shown in (Finnish) locale, which is good. But when not in edit mode, we these fields are shown as ISO format strings. I think we should use the localised versions also in the "view mode"
- please add spaces before the opening "(" character in "Enter number (decimal)" and "Enter number (integer)"
- When you add newlines in a long text, it looks fine in edit mode, but in the view mode, the field is shown without newlines. I think we should show long texts with newlines also in view mode
- editingFieldDenied works better now, but the message is not shown. This is probably not a part of this story?

henryhaverinen:
Some more comments:

- If you set fields to null, then the app shows all other fields simply as empty, except for Boolean fields it shows "No", which is incorrect. The exported site shows N/A for all null fields. I think we should simply show null fields as empty (value is missing) in both the app and the exported site.
- The display values of boolean fields need some work. The app uses "Yes" and "No" as display values in the view mode, but "True" and "False" in the edit mode.  Exported site shows boolean values as "true" and "false". I propose we use "Yes" and "No" as the boolean display values everywhere 
- In the exported site, newlines in long texts are ignored similarly as in the view mode of the app. Long text fields should be "long", if there are newlines, in the site, too
- In the exported site, enum and list values are shown as the enumValues. I was expecting them as the enumDisplayValues 
- For the list data type in app view mode and the exported site, the options should be shown in the same order as in the field type definition (ordered by the "index"). In my testing, I got them in the same order in the app, but in some other order in the exported site
- Date and date time fields are shown as raw ISO texts in the exported site. Should be localised dates and times
- On the exported site, there is a bit too much space between the field name and the value. In the app, the gap is smaller and better. See the image below.

image

henryhaverinen:
One more thing: the placeholder text for an email address should be firstname.lastname@example.com, because the example.com domain has been reserved for this kinds of purposes.

Fuzzbender (approved):
Code-wise looks good to me.

samutoljamo-pareto:
> In edit mode, the current values of date and datetime fields are shown in (Finnish) locale, which is good. But when not in edit mode, these fields are shown as ISO format strings. I think we should use the localised versions also in the "view mode"

This is not that easy, because the implementation of the native datepicker(input field with type date or datetime-local) uses system's locale, which is unavailable from JS. JS has access to navigator.language, but that won't be equivalent. I can make a separate ticket on the options so we get this PR done.
E.g.: Below are my locale settings, which are used by the input field while navigator.language gives 'en-US'.
image
> editingFieldDenied works better now, but the message is not shown. This is probably not a part of this story?

yes

PR #381, opened 2024-12-23
INTDEV-542 Add labels support to app

samutoljamo-pareto:
App: can now add/remove/view labels.
Site export: Labels are also visible

Labels are now validated by DH's validate function. Currently they use the same convention as resource names, but that can be changed if needed


Fuzzbender (approved):
Looks good to me

PR #382, opened 2024-12-29
INTDEV-421 Add support for custom themes

samilempinen:
Usage:

    $ cyberismo export site /tmp/site -t <path-to-ui-bundle>

This is equivalent to running the export without the `-t` option:

    $ cyberismo export site /tmp/site -t ../cyberismo/resources/ui-bundle

### DIFF
--- a/resources/ui-bundle/partials/footer-content.hbs
+++ b/resources/ui-bundle/partials/footer-content.hbs
@@ -1,3 +1,10 @@
 <footer class="footer">
-  <p>Built with <a href="https://cyberismo.com">Cyberismo</a>, the open source solution for Security-As-Code.</p>
+  <p>Built with <a href="https://cyberismo.com/">Cyberismo</a>.</p>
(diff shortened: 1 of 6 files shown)


Fuzzbender:
Looks good to me, but I'd like to have some unit tests too. For example, what happens when path is incorrect.

Fuzzbender (approved):

samilempinen:
> Looks good to me, but I'd like to have some unit tests too. For example, what happens when path is incorrect.

Yeah, I was actually adding these, but ran out of skills. I will unstash them and continue.

samilempinen:
@Fuzzbender I have some trouble figuring out the right way to handle errors in the CLI commands. I refactored the export to return proper requestStatus, and added a test for it. However, if I verify the request status in the test, the command line does not really give a proper indication that the command failed. It seems that for the CLI to give a proper error message, I need to pass the exception through to `doHandleCommand()`, but this makes it difficult to use the requestStatus in the tests.

What should be the best practice here?

I left things so that the CLI works but the test fails.

PR #383, opened 2024-12-30
INTDEV-646 Sort customFields in logic programs

samutoljamo-pareto:
Logic programs made from fieldstypes generated the index of customFields based on the customFields array rather than alwaysvisibleFields/optionallyvisible-fields. Fixed it by sorting customFields. Now the order is:
1. Order in alwaysVisibleFields
2. Order in optionallyVisiblefields
3. If not found at all, last


Fuzzbender (approved):
Looks good to me

PR #394, opened 2024-12-31
INTDEV-406: Added graph macro

samutoljamo-pareto:
1. Data handler processes all macros of the index.adoc of a card and encounters a graph macro
3. Validate the mandatory parameters of the graph macro
4. Validate the parameters of the graph macro against the view-specific parameter schema (if given)
5. Generate the view logic file using Handlebars and the macro parameters
6. Run clingo for the main, model, view files  and check that it was successful, to get a clingraph logic program
7. Run clinggraph on the clingraph logic program. Save the diagrams to a temporary location
8. Replace the spinner or the cached old copy of the diagram with the new version

Note: I changed the our temp directory to be at root of the project in the `.temp`-folder. This means it must be also added to .gitignore of each project. Templates also use `.temp/cards`. @Fuzzbender 

> If possible, it would be nice to render the results of the macro in phases so that displaying the rest of the card is not blocked by the graph. So while waiting for the image to be created, either a spinner or a previously cached version of the diagram could be shown.
This can be left to a follow-up story, if needed.

Everything else done except for the spinner, which I think should not be specific to this. At simplest, it could be implemented by introducing another macro mode, which could be called "preInject". Each macro could decide either to display the actual content(buttons) or a spinner if the operation is expected to take a while. Then the app would first get the contents using preInject and then replace the content when the call made using the normal mode is ready.

### DIFF
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -52,6 +52,7 @@ export class Calculate {
 
   private logicBinaryName: string = 'clingo';
   private clingGraphBinary: string = 'clingraph';
+  private pythonBinary: string = 'python';
   private static importResourcesFileName: string = 'resourceImports.lp';
   private static importCardsFileName: string = 'cardTree.lp';
   private static modulesFileName: string = 'modules.lp';
@@ -597,12 +598,23 @@ export class Calculate {
       `--name-format=${randomId}`,
       `--dir=${this.project.paths.tempFolder}`,
     ];
-    spawnSync(this.clingGraphBinary, clingGraphArgs, {
-      encoding: 'utf8',
-      input: firstLine,
-      timeout,
-      maxBuffer: 1024 * 1024 * 100,
-    });
+
+    // python is only used for windows
+    const pythonArgs = [
+      '-c',
+      `from clingraph import main;import sys; sys.argv = [sys.argv[0], ${clingGraphArgs.map((arg) => `"${arg}"`).join(',')}];sys.exit(main())`,
+    ];
+
+    spawnSync(
+      process.platform === 'win32' ? this.pythonBinary : this.clingGraphBinary,
+      process.platform === 'win32' ? pythonArgs : clingGraphArgs,
+      {
+        encoding: 'utf8',
+        input: firstLine,
+        timeout,
+        maxBuffer: 1024 * 1024 * 100,
+      },
+    );
 
     const filePath = join(this.project.paths.tempFolder, randomId);
 
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -12,7 +12,7 @@
 
 // node
 import path, { basename, join, resolve } from 'node:path';
-import { mkdir, readFile } from 'node:fs/promises';
+import { mkdir, readFile, rm } from 'node:fs/promises';
 import { spawnSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 
@@ -44,12 +44,14 @@ import {
   createWorkflowFacts,
 } from './utils/clingo-facts.js';
 import { ClingoProgramBuilder } from './utils/clingo-program-builder.js';
+import { generateRandomString } from './utils/random.js';
 
 // Class that calculates with logic program card / project level calculations.
 export class Calculate {
   private static mutex = new Mutex();
 
   private logicBinaryName: string = 'clingo';
+  private clingGraphBinary: string = 'clingraph';
   private static importResourcesFileName: string = 'resourceImports.lp';
   private static importCardsFileName: string = 'cardTree.lp';
   private static modulesFileName: string = 'modules.lp';
@@ -446,30 +448,38 @@ export class Calculate {
       content = compiled(options);
     }
 
-    const result = await this.run({
+    const clingoOutput = await this.run({
       query: content,
     });
+
+    const result = await this.parseClingoResult(clingoOutput);
 
     if (result.error) {
       throw new Error(result.error);
     }
     return result.results as QueryResult<T>[];
   }
-
   /**
-   * Runs a logic program.
+   * Runs a logic program using clingo.
    *
    * @param filePath Path to a query file to be run in relation to current working directory
    * @param timeout Specifies the time clingo is allowed to run
    * @returns parsed program output
    */
-  public async run(
+  public async runLogicProgram(data: { query?: string; file?: string }) {
+    const clingoOutput = await this.run(data);
+
+    return this.parseClingoResult(clingoOutput);
+  }
+
+  private async run(
     data: {
       query?: string;
       file?: string;
     },
+    argMode: 'graph' | 'query' = 'query',
     timeout: number = 5000,
-  ): Promise<ParseResult<BaseResult>> {
+  ): Promise<string> {
     const main = join(
       this.project.paths.calculationFolder,
       Calculate.mainLogicFileName,
@@ -485,12 +495,21 @@ export class Calculate {
       );
     }
 
-    const args = ['-', '--outf=0', '--out-ifs=\\n', '-V0', main, queryLanguage];
-    if (data.file) {
-      args.push(data.file);
-    }
-
     return Calculate.mutex.runExclusive(async () => {
+      const args = ['-', '--outf=0', '-V0', '--warn=none'];
+
+      if (argMode === 'graph') {
+        args.push('--out-atomf=%s.');
+      } else {
+        args.push('--out-ifs=\\n');
+        args.push(queryLanguage);
+      }
+      args.push(main);
+
+      if (data.file) {
+        args.push(data.file);
+      }
+
       const clingo = spawnSync(this.logicBinaryName, args, {
         encoding: 'utf8',
         input: data.query,
@@ -510,7 +529,7 @@ export class Calculate {
           stdout: clingo.stdout,
           stderr: clingo.stderr,
         });
-        return this.parseClingoResult(clingo.stdout);
+        return clingo.stdout;
       }
 
       if (clingo.stderr && clingo.status) {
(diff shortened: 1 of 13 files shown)


samutoljamo-pareto:
If you want to test this, I think easiest way is to install miniconda and use clingo / clingraph via that. I have not yet updated our pipelines and there is no test for the graphs yet so they pass.

```
conda create -n env-name
conda activate env-name
conda install -c potassco clingraph
cyberismo app # should work now
```

samutoljamo-pareto:
About windows: The current implementation won't work without tricks, since conda installation provides a bash script for windows too. I created a .bat file that runs clingraph and it worked after adding also shell: true to spawnSync-function.

henryhaverinen:
I made some tests with this branch https://github.com/CyberismoCom/module-dataflow/tree/feature/henryhaverinen/INTDEV-406-visualisations and the macro seems to work great!
It is fast enough on my machine without any spinners.

tools/data-handler/src/macros/common.ts
```diff
@@ -12,11 +12,13 @@
 
 // important that this file imports only the metadata
 import createCards from './createCards/metadata.js';
-import scoreCard from './scoreCard/metadata.js';
+import graph from './report/metadata.js';
```
henryhaverinen, line 15: should this be: `import graph from './graph/metadata.js';`

tools/schema/graphMacroBaseSchema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "title": "Graph",
+  "$id": "graphMacroBaseSchema",
+  "description": "A report object provides supplemental information about a report",
```
henryhaverinen, line 4: The description could say "A graph object provides suppplemental information about a graph"

tools/data-handler/src/calculate.ts
```diff
@@ -482,12 +492,21 @@ export class Calculate {
       );
     }
 
-    const args = ['-', '--outf=0', '--out-ifs=\\n', '-V0', main, queryLanguage];
-    if (data.file) {
-      args.push(data.file);
-    }
-
     return Calculate.mutex.runExclusive(async () => {
+      const args = ['-', '--outf=0', '-V0', '--warn=none'];
```
Fuzzbender, line 496: To make the exclusive section of the code shorter, you could form the `args` before and then just use it inside the `runExclusive` section.

tools/data-handler/src/calculate.ts
```diff
@@ -552,4 +571,57 @@ export class Calculate {
       );
     });
   }
+
+  /**
+   * Runs given logic program and creates a graph using clingraph
+   * @param data
+   * @param timeout
+   * @returns
```
Fuzzbender, line 579: Missing the comment block descriptions

tools/data-handler/src/calculate.ts
```diff
@@ -552,4 +571,57 @@ export class Calculate {
...
+
+    const filePath = join(this.project.paths.tempFolder, randomId);
+
+    let fileData;
+    try {
+      fileData = await readFile(filePath + '.png');
+      await rm(filePath + '.png');
+    } catch (e) {
+      throw new Error(
+        `Graph: Failed to read image file after generating grape: ${e}`,
```
Fuzzbender, line 614: throw can come from delete as well (very unlikely; could be some permission thing).

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+import { createImage, validateMacroContent } from '../index.js';
+
+import { MacroGenerationContext } from '../../interfaces/macros.js';
+import macroMetadata from './metadata.js';
+import { Project } from '../../containers/project.js';
+import { Calculate } from '../../calculate.js';
+import Handlebars from 'handlebars';
+import BaseMacro from '../BaseMacro.js';
+import { readFile } from 'fs/promises';
+import { join } from 'path';
```
Fuzzbender, line 22: Should be `node:path`. We always include to node imports the `node:` prefix.

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public
+    License along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { createImage, validateMacroContent } from '../index.js';
+
+import { MacroGenerationContext } from '../../interfaces/macros.js';
```
Fuzzbender, line 15: Imports could be alphabetically sorted.

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  model: string;
+  view: string;
+}
+
+class ReportMacro extends BaseMacro {
+  constructor() {
+    super(macroMetadata);
+  }
+
+  handleValidate = (data: string) => {
```
Fuzzbender, line 39: Different style from other code. 

Should be public APIs, e.g. `public handleValidate(data: string) { ... }`

Fuzzbender: Also, does it need to have `handle` in the beginning? Isn't this just `validate()`?

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.validate(data);
+  };
+
+  handleStatic = async (context: MacroGenerationContext, data: string) => {
```
Fuzzbender, line 43: Is this needed, as it is the same as calling `handleInject` directly?

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    return this.handleInject(context, data);
+  };
+
+  handleInject = async (context: MacroGenerationContext, data: string) => {
+    const project = new Project(context.projectPath);
+    const calculate = new Calculate(project);
+
+    const resourceNameToPath = (name: string, ending: string) => {
```
Fuzzbender, line 51: `ending` --> `extension`

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -0,0 +1,130 @@
...
+      return join(
+        project.paths.modulesFolder,
+        prefix,
+        type,
+        identifier,
+        ending,
+      );
+    };
+
+    const options = this.validate(data);
```
Fuzzbender, line 65: Do this after sorting out the schema.

Fuzzbender: And maybe rename to something else than `options`. Maybe `validationResult`?

Fuzzbender, line 65: A bit strange name for validation result `options`.

tools/data-handler/src/macros/graph/index.ts
Fuzzbender: No, this is not extension, but filename (based on code)

Fuzzbender (approved):
Some minor editorial comments. Looks good to me.

tools/data-handler/src/macros/common.ts
samutoljamo-pareto: done

tools/schema/graphMacroBaseSchema.json
samutoljamo-pareto: done

tools/data-handler/src/calculate.ts
samutoljamo-pareto: done

tools/data-handler/src/calculate.ts
samutoljamo-pareto: done

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: done, also added node: for fs

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: done

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: Yes, since it's an abstract function inherited from base macro

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: Changed name to `fileName`

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: I changed the function name to `parseOptions`

henryhaverinen:
Now that this branch includes the `concatenate` Python function, I updated the test content in module-dataflow and got rid of its Python function `cluster`. Works fine: https://github.com/CyberismoCom/module-dataflow/tree/feature/henryhaverinen/INTDEV-406-visualisations

samutoljamo-pareto:
I'll wait with this until #423 is merged and then I'll rebase to main.

tools/data-handler/src/macros/graph/index.ts
samutoljamo-pareto: Fixed all macros

samutoljamo-pareto: Actually, I reverted this change since it's much easier to use the setter syntax, since we use the functions outside of the class. Of course, we could bind each function in the constructor

henryhaverinen (approved):

tools/data-handler/src/calculate.ts
samutoljamo-pareto: done

PR #423, opened 2025-01-06
Python functions daysSince and concatenate

henryhaverinen:
Add two first Python functions for logic programs.

`@daysSince(Date)` returns the number of days since `Date`. If `Date` is not a valid ISO date, then `@daysSince` returns 0. This function is useful, for example, when checking whether a review has expired.

`@concatenate(*args)` takes a variable number of arguments and returns a string that concatenates all arguments. The arguments can be numbers, strings or constants (which are a special case of Clingo functions). This function can be used, for example, to create dynamic strings for notifications and policy check failure error messages.

### DIFF
--- a/.github/workflows/build-and-test-windows.yml
+++ b/.github/workflows/build-and-test-windows.yml
@@ -30,8 +30,9 @@ jobs:
       with:
         python-version: 3.12
         auto-update-conda: true
-    - name: Install Clingo from Potassco channel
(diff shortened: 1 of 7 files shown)


samutoljamo-pareto (approved):
Looks good to me. I separated the tests for the python functions, created a yml file for conda and updated clingo in Windows/Linux CI to a python supported version

PR #426, opened 2025-01-08
Move shared dependencies to root level

Fuzzbender:
Move all shared dependencies from specific tool's package.json to workspace's package.json. 
This way the same version of dependency is shared by all of the tools, if necessary. And we limit the number of Pull Requests that dependabot automatically creates.

Also, move "sinon" dependency in data-handler's package.json to be devDepedency instead of production dependency.

Tested that majority of "scripts" still work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,21 @@
   },
   "devDependencies": {
     "concurrently": "^9.1.2",
-    "prettier": "^3.3.2"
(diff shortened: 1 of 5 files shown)


samutoljamo-pareto (approved):
Looks good to me and app in dev mode seems to work still in wsl

samilempinen (approved):

PR #435, opened 2025-01-09
Add Dockerfile to enable container image building

CyberJML:
Some notes:
- This Dockerfile is not properly optimized
- Image should be based on some smaller image, like Alpine
- Clingo should be built (with python support) in builder container and installed on the final image
- Current image size is 2.69GB, but could be brought down to 1.5GB

This version is just a enabler for the use of container environments and further optimizations will follow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+FROM docker.io/continuumio/miniconda3:24.11.1-0
+
+# Set environment variables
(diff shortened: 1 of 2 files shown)


CyberJML:
By adding Clingraph, image size went 3.3GB

henryhaverinen (approved):
I tested this (using the instructions in the corresponding cyberismo-docs PR) and it worked great.

PR #437, opened 2025-01-14
INTDEV-635 Change ResourceFolderType to use plural resource names

Fuzzbender:
Optimize & simplify current implementation.

By making the `ResourceFolderType` use plural names of resources, we can skip some mapping and naive pluralisation in the code.

Additionally, remove functions that become unnecessary: `resourceFullName` and `singularType`.

Edit. Note that this is intentionally targeting a feature branch. Once INTDEV-636 has been merged, content repos need to be updated. Then this can be merged.


samutoljamo-pareto (approved):
Looks good to me

PR #438, opened 2025-01-14
INTDEV-594 add all resource types to module schemas

Fuzzbender:
Add `calculations` and `reports` to the module part of the schema.

Make all of the resource types to be references, so that we don't need to duplicate them in local resources and module resources parts.

**Note** that this is intentionally targeting a feature branch. Once PRs for INTDEV-636 and INTDEV-635 have been merged, content repos need to be updated. Then this can be merged.


samutoljamo-pareto (approved):
Good change, looks good to me.

PR #440, opened 2025-01-15
Improve Python function daysSince and implement Python function today

henryhaverinen:
Previously, daysSince did not work for datetimes. We also need a function to get today's date for checking whether a due date has passed.

### DIFF
--- a/calculations/common/pythonFunctions.lp
+++ b/calculations/common/pythonFunctions.lp
@@ -2,7 +2,7 @@
 
 #script (python)
 from clingo import Number, String, SymbolType
-from datetime import date
+from datetime import date, datetime, timezone
 
 class Functions:
     def concatenate(self, *args):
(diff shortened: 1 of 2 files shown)


calculations/common/pythonFunctions.lp
```diff
@@ -23,12 +23,21 @@ class Functions:
...
+            # datetimes are stored in UTC format in Cyberismo, so for them there is no effect
+            userdate = userdate.replace(tzinfo=timezone.utc)
+            today = datetime.now(timezone.utc)
             return Number((today-userdate).days)
         except:
             return Number(0)
 
+    def today(self):
+        try:
+            return String(date.today().isoformat())
```
Fuzzbender, line 37: As this is using system values, I don't think it can really throw?
Maybe if memory is totally consumed and it cannot allocate a `String`. But then the catch will fail as well.

Fuzzbender (approved):

calculations/common/pythonFunctions.lp
henryhaverinen: Do you think we should remove the try/except from this function?

henryhaverinen: I removed the exception handling

PR #445, opened 2025-01-17
INTDEV-671 Disable card specific logic program generation

samutoljamo-pareto:
App depends on generating all logic programs again anyway(since everything is not updated) so removing generation of logic programs after a transition actually also improves performance. However, if you do a transition via CLI and run a logic program using the CLI, you'll need to generate the files before doing so.

The original issue was that the modules.lp was not being generated although we delete the whole folder before doing calculations. Also for some reason, after removing that issue, the emit still caused another issue: Clingraph run successfully, but produced an empty image. For now, it's simpler to simply rely on generating all logic programs before calculations. We should be able to get rid of that once the calculations are moved to the resource classes.

### DIFF
--- a/tools/data-handler/src/calculate.ts
+++ b/tools/data-handler/src/calculate.ts
@@ -228,11 +228,7 @@ export class Calculate {
   }
 
   // Collects all logic calculation files from project (local and imported modules)
-  private async generateModules(parentCard: Card | undefined) {
-    // When generating calculations for a specific module, do not generate common calculations.
-    if (parentCard) {
-      return;
-    }
+  private async generateModules() {
     const destinationFile = join(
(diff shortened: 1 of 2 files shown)


Fuzzbender:
" However, if you do a transition via CLI and run a logic program using the CLI, you'll need to generate the files before doing so."
I think this is fine. If we want to make it better; CLI could automatically generate calculations internally after each "change" command. I think this sounds quite ok and is easy to do.

Fuzzbender:
"We should be able to get rid of that once the calculations are moved to the resource classes."
Which reminds me that we don't have story for this yet; but we have empty implementations in resource classes for the calculations.

Fuzzbender (approved):

PR #447, opened 2025-01-17
INTDEV-620 Rename isEditable to isCalculated

samutoljamo-pareto:
This will rename isEditable to isCalculated and prevent saving those. This will still work with repos that use `isEditable`, but I'll make another PR that will remove the backwards compatibility.

### DIFF
--- a/calculations/queries/card.lp
+++ b/calculations/queries/card.lp
@@ -72,7 +72,24 @@ field(Field, "value", Value) :-
 
 
 % select only non-custom fields
-selectAll.
+select(1, "cardType").
+select(1, "title").
+select(1, "key").
+select(1, "lastUpdated").
+select(1, "workflowState").
+select(1, "notifications").
+select(1, "policyChecks").
+select(1, "links").
+select(1, "deniedOperations").
+select(1, "labels").
+
+select(2, "visibility").
+select(2, "index").
+select(2, "fieldDisplayName").
+select(2, "fieldDescription").
+select(2, "dataType").
+select(2, "isCalculated").
+select(2, "value").
 
 childResult({{cardKey}}, Field, "fields") :-
     alwaysVisibleField(CardType, Field),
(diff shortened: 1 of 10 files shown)


samutoljamo-pareto:
Another PR: https://github.com/CyberismoCom/cyberismo/pull/448

samutoljamo-pareto:
Notes: I made it so that calculated fields are not saved at all. This must be also fixed in other repos. In the future, having a value in a calculated field will raise a validation error

Fuzzbender (approved):
Looks good to me.

henryhaverinen:
> Notes: I made it so that calculated fields are not saved at all. This must be also fixed in other repos. In the future, having a value in a calculated field will raise a validation error.
> 
> Also @henryhaverinen I introduced a new fact, calculatedField

I didn't find the code that would generate these facts. Would it go as part of the facts that are generated from the card type?

I wonder if this kind of syntax would be better aligned with how we generate the other cardtype+fieldtype related facts:

`field((CardType, FieldType), "isCalculated", true).`

In other words, we'd be considering `isCalculated` as a field of the tuple of the card type and the field type, similarly to, for example:
`field(("ismsa/cardTypes/risk", "ismsa/fieldTypes/riskInitialRisk"), "displayName", "Initial risk level").`

henryhaverinen:
I tested this and it works fine. It does seem to save my test calculated field in the index.json file as a null value, though. I suppose that is as it is supposed to be, because the field is required for validation.

henryhaverinen (approved):

samutoljamo-pareto:
> I tested this and it works fine. It does seem to save my test calculated field in the index.json file as a null value, though. I suppose that is as it is supposed to be, because the field is required for validation.

I changed it so that validation will produce an error if it is defined

henryhaverinen:
> > I tested this and it works fine. It does seem to save my test calculated field in the index.json file as a null value, though. I suppose that is as it is supposed to be, because the field is required for validation.
> 
> I changed it so that validation will produce an error if it is defined. It is not supposed to do that, but I assume it was there before?

No, it appears when you instantiate from template. I just updated the main branch of module-test. You can type `cyberismo create card test/templates/allDataTypes`, and then just check out the index.json of the new card, and compare it to the template card.

PR #449, opened 2025-01-20
Add new Dockerfile structure & Cmake based clingo build in image

CyberJML:
Cmake based docker image to reduce the size of docker containers and to make build&deployment easier. The current sizes:
- Cyberismo App image 1.86GB (2.49GB previous)
- cyberismo-clingo-base 346MB (contains self built Clingo, Clingraph, Graphviz, and other essential tools for out images)


Fuzzbender:
@CyberJML was it meant that the 'assignees' would review this?

CyberJML:
Yes :D changing

Fuzzbender (approved):

CyberJML:
For some reason Clingo only works properly with Python 3.11. Other versions are shown as "Python enabled", but hangs while executing Python code...

CyberJML:
> Looks good to me. Python 3.11 is used instead of 3.12 for some reason. Probably a good one. :) Why?

For some reason Clingo only works properly with Python 3.11. Other versions are shown as "Python enabled", but hangs while executing Python code...

PR #450, opened 2025-01-20
INTDEV-673 - Performance: Make card creation faster

Fuzzbender:
Remove 'find' when checking if a created card ID is used. Instead before creating cards, fetch all cards in the project and check if the created cardID is included or not. 

This makes huge difference once there are more than 1000 cards in the project.

Add new API for this that only fetches cardIDs. While at it, rename existing `listAllCards` to `listCards` and replace boolean parameter with an enum. This way we can control where to look for cards.

### DIFF
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -17,6 +17,7 @@ import { readdir, readFile } from 'node:fs/promises';
 import {
   Card,
   CardAttachment,
+  CardLocation,
   CardListContainer,
   CardMetadata,
   CardNameRegEx,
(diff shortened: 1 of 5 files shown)


tools/data-handler/src/containers/project.ts
```diff
@@ -753,15 +822,19 @@ export class Project extends CardContainer {
...
+  public async newCardKey(cardIds: string[] = []): Promise<string> {
     const maxAttempts = 10;
     const base = 36;
     const length = 8;
     for (let i = 0; i < maxAttempts; i++) {
       // Create a key and check that there are no collisions with other keys in project
       const newKey = `${this.settings.cardKeyPrefix}_${generateRandomString(base, length)}`;
-      const exists = await this.findSpecificCard(newKey);
-      if (exists) continue;
+      const found = cardIds.find((item) => item === newKey);
```
samutoljamo-pareto, line 832: could be replaced with `cardIds.includes(newKey)`

tools/data-handler/src/containers/project.ts
```diff
@@ -695,6 +705,65 @@ export class Project extends CardContainer {
...
+    if (
+      cardsFrom === CardLocation.all ||
+      cardsFrom === CardLocation.projectOnly
+    ) {
+      projectCardIDs = (await super.cards(this.paths.cardRootFolder)).map(
+        (item) => item.key,
+      );
+    }
+
+    if (
```
samutoljamo-pareto, line 728: Both ifs are the same

tools/data-handler/src/containers/project.ts
```diff
@@ -695,6 +705,65 @@ export class Project extends CardContainer {
     return cardListContainer;
   }
 
+  /**
+   * Return cardIDs of the cards in the project or from templates, or both.
+   * @param includeCardsFrom Where to return cards from (project, templates, or both)
+   * @returns Array of cardIDs.
+   * @note that cardIDs are not sorted. Project cards are first in the array, followed by template cards.
```
samutoljamo-pareto, line 712: Since this  function is only used if we're checking if a certain id exists, using a set would speed things up too. But at least I think it's more clear to use arrays everywhere for as long as that is not the limiting issue.

samutoljamo-pareto (approved):
Minor comments, looks good to me.

tools/data-handler/src/containers/project.ts
Fuzzbender: I can check if using `Set` here would improve the performance .

Fuzzbender: Based on my local results, it is more or less the same perf when using `Set<string>` vs `string[]`.

PR #458, opened 2025-01-21
INTDEV-674 Read module resources only once 

Fuzzbender:
Previously, module-resources were collected multiple times. 
Also, `CommandManager` is actually initialised twice per CLI command, so also local resources are collected twice. 
Add functionality that resources are collected only once. Note that this is just a collection of what resource names and paths are in the project. It does not fix accessing resource's metadata file multiple times. 

Additionally, provide a better `getFilesSync` function that does not use `lstatSync `.


samutoljamo-pareto (approved):
Looks good to me

PR #459, opened 2025-01-22
INTDEV-681 Fetch workflow data once per instantiated template card

Fuzzbender:
When cards from template are instantiated, workflow's initial state is fetched to card's metadata. 
Unfortunately, the operation fetches the workflow information twice: 
- first, for general access of workflow
- second time, to find initial state of the workflow

Remove the latter, as the code itself can deduce the initial state.


samutoljamo-pareto (approved):
Looks good to me

PR #460, opened 2025-01-22
INTDEV-690  When creating cards, fetch only root level project cards as siblings

Fuzzbender:
Current implementation, fetches ALL project card for each instantiated template card. 
Only the root level cards are needed as potential siblings (needed for ranking). 
This only affects situation where cards are instantiated to root level of project (ie. parent card is not defined)


samutoljamo-pareto (approved):
Looks good to me

PR #461, opened 2025-01-22
INTDEV-679 switch from Math.random to crypto.randomInt

Fuzzbender:
Start using `crypto.randomInt` instead of `Math.random`.

### DIFF
--- a/tools/data-handler/src/utils/random.ts
+++ b/tools/data-handler/src/utils/random.ts
@@ -10,10 +10,15 @@
     License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import { randomInt } from 'node:crypto';
+
+// Maximum base for randomInt.
+const MAX_VALUE = Math.pow(2, 48) - 1;
+
 /**
- * Returns a random string in given base with given length (padded with 0's if necessary)
- * @param {number} base which base to be used, e.g. 36 for base-36, must be an integer
- * @param {number} length length of string, must be an integer
+ * Returns a random string in given base with given length.
+ * @param base which base to be used, e.g. 36 for base-36, must be an integer
+ * @param length length of string, must be an integer
  * @returns a random string
  * @throws if parameters are not integer numbers
  */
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/utils/random.ts
```diff
@@ -23,8 +28,8 @@ export function generateRandomString(base: number, length: number): string {
   }
 
   // Generate a random number between 0 and given base max number - 1
-  const maxBaseNumber = Math.pow(base, length);
-  const randomNum = Math.floor(Math.random() * maxBaseNumber);
+  const maxBaseNumber = Math.min(MAX_VALUE, Math.pow(base, length));
+  const randomNum = randomInt(maxBaseNumber);
```
samutoljamo-pareto, line 32: Actually, this logic still won't work, because we cannot generate longer randomized strings. Ways to fix:
1. Generate a random number for each character(slower)
2. Generate a random number like in the existing code using `Math.min(MAX_VALUE, Math.pow(base, length));`, but instead of continueing right away, it could check if the generated string is long enough. If not, generate another random number. The current code uses padStart to fill the string with 0's if its not long enough

Fuzzbender: If the generated random numbers were larger, they would produce longer random strings eventually?
Thus, we could pass a minimum value to `crypto.randomInt` and there would be no need to check the generated value or ask it to re-generate a new value.

Fuzzbender: Ie. 
``
const maxBaseNumber = Math.min(MAX_VALUE, Math.pow(base, length));
const minValue = Math.pow(2, 10); // just some random largeish value
const randomNum = randomInt(minValue, maxBaseNumber);
...
``

samutoljamo-pareto: I was mainly concerned if base/length are too large, the number will contain a lot of zeros.
e.g.
```
> const a = crypto.randomInt(Math.pow(2, 48)-1)
undefined
> a.toString(36).padStart(20, '0')
'00000000001igbuw99mu'
```

Fuzzbender: Ok. 

How about we generate random value, turn it into string, if it is still too short, repeat the process.
Something like (this is pure JS):
```
function generateRandomString2(base, length) {
  const MAX_VALUE = Math.pow(2, 48) - 1;
  const maxBaseNumber = Math.min(MAX_VALUE, Math.pow(base, length));
  let randomNum = randomInt(maxBaseNumber);
  let generatedStr = randomNum.toString(base);
  while (generatedStr.length < length) {
    randomNum = randomInt(maxBaseNumber);
    generatedStr += randomNum.toString(base);
  }
  return generatedStr.substring(0, length);
}
```

From nodeJS console:
Screenshot 2025-01-24 at 10 57 29

samutoljamo-pareto: Yes, that works. But I think this would be a bit more concise
```
  let generatedStr = ''
  while (generatedStr.length < length) {
    generatedStr += randomInt(maxBaseNumber).toString(base);
  }
```

Fuzzbender: The `crypto` random generator should be faster generally than the `Math` one.

samutoljamo-pareto (approved):

PR #462, opened 2025-01-23
INTDEV-691 Use readdir's recursion when using find or collect from containers

Fuzzbender:
Re-write 'find' and 'collect' implementations of card-container base class to use `readdir` recursion instead of our own.

While the performance is about the same, the implementation is smoother and we don't have to pass a parameter where to collect values, but can use return value only.


samutoljamo-pareto (approved):
Looks good to me

PR #463, opened 2025-01-23
INTDEV-690 Parallel version of listCardIds()

Fuzzbender:
Make `listCardIds()` process container cards (project, all templates) parallel in their own promises. 
Also, add all found cards in one call to `Set` instead of adding one-by-one.

### DIFF
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -713,61 +713,65 @@ export class Project extends CardContainer {
    * Return cardIDs of the cards in the project or from templates, or both.
    * @param includeCardsFrom Where to return cards from (project, templates, or both)
    * @returns Array of cardIDs.
-   * @note that cardIDs are not sorted. Project cards are first in the array, followed by template cards.
+   * @note that cardIDs are not sorted.
    */
   public async listCardIds(
     cardsFrom: CardLocation = CardLocation.all,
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/containers/project.ts
```diff
@@ -713,61 +713,61 @@ export class Project extends CardContainer {
...
-            result.status === 'fulfilled',
-        )
-        .map((result) => result.value)
-        .flat()
-        .forEach((result) => cardIDs.add(result));
     }
-
-    return cardIDs;
+    const allCardIds = await Promise.all(promises);
+    return new Set(allCardIds.flat());
```
samutoljamo-pareto, line 770: Could we just append to set directly?

samutoljamo-pareto (approved):

tools/data-handler/src/containers/project.ts
Fuzzbender: Not sure that I can parallel add to a single instance of `Set`. The whole change is about sequential -> parallel, thus not changing this back to sequential.
I'll change it so that each parallel call adds to its own `Set` and finally the results are collected to a single `Set`.

samutoljamo-pareto: But they are still promises so adding to a single set is fine. Promises do not run in parallel, they are executed on a single-threaded event loop.

PR #464, opened 2025-01-23
INTDEV-692 Optimize template card instantiation

Fuzzbender:
Previously, when template cards were created, they were first created to a `.cards/temp` folder. Then if card creation succeeded, cards were copied to correct place and temp data was deleted. 

To make template card instantiation simpler, create cards directly to correct place. If there is an error instantiating the cards, use the mapping table to remove the partially created template cards.

### DIFF
--- a/tools/data-handler/src/containers/template.ts
+++ b/tools/data-handler/src/containers/template.ts
@@ -12,7 +12,7 @@
 
 // node
 import { basename, join, resolve, sep } from 'node:path';
-import { copyFile, readdir, mkdir, rm, writeFile } from 'node:fs/promises';
+import { copyFile, mkdir, readdir, rmdir, writeFile } from 'node:fs/promises';
 import { readdirSync } from 'node:fs';
 
 // Base class
@@ -27,8 +27,9 @@ import {
   Resource,
 } from '../interfaces/project-interfaces.js';
 import { CardType, Workflow } from '../interfaces/resource-interfaces.js';
-import { copyDir, pathExists, stripExtension } from '../utils/file-utils.js';
+import { pathExists, stripExtension } from '../utils/file-utils.js';
 import { DefaultContent } from '../resources/create-defaults.js';
+
 import {
   EMPTY_RANK,
   FIRST_RANK,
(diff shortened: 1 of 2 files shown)


tools/data-handler/src/containers/template.ts
```diff
@@ -92,184 +87,188 @@ export class Template extends CardContainer {
...
-        from: card.key,
-        to: await this.project.newCardKey(cardIds),
+    const templateIDMap = new Map<string, string>();
+    // Create ID mapping and update ranks
+    const createMappingAndRanks = async () => {
+      const cardIds = await this.project.listCardIds();
+      const newCardIds = this.project.newCardKeys(cards.length, cardIds);
+
+      // Create mapping table
+      cards.map((card, index) => {
```
CyberJML, line 101: Should this be cards.forEach? Since cards.map returns all the items from array that are truthful during the rotation.

Fuzzbender: Sure, I think it could be `forEach`

tools/data-handler/src/containers/template.ts
```diff
@@ -321,6 +320,25 @@ export class Template extends CardContainer {
     return '';
   }
 
+  // Removes cards
+  private async removeCards(cardMap: Map<string, string>) {
+    const tasks: Promise<Card | undefined>[] = [];
+    // Find all cards that need to be removed.
+    cardMap.forEach(async (createdCard) => {
```
CyberJML, line 327: unnecessary async?

tools/data-handler/src/containers/template.ts
```diff
@@ -321,6 +320,25 @@ export class Template extends CardContainer {
...
+      tasks.push(this.project.findSpecificCard(createdCard));
+    });
+    // Remove empty results.
+    const cards = (await Promise.all(tasks)).filter(
+      (item) => item !== undefined,
+    );
+    // Delete card folders.
+    const deleteAll: Promise<void>[] = [];
+    cards.forEach(async (card) => {
```
CyberJML, line 336: unnecessary async?

CyberJML (approved):

PR #466, opened 2025-01-27
INTDEV-706 Add clingo and graphviz binary postinstall & custom path during execution

CyberJML:
Postinstall script that will:
- Download our own built clingo from https://github.com/CyberismoCom/cyberismo-base/tags (not public)
- Extract it under vendor/utils
- Download latest graphviz binary
- Extract it under vendor/utils

Run alteration:
- Added custom path for these two binaries, so that runtime inclusion works properly

### DIFF
--- a/.github/workflows/build-clingo-binaries.yml
+++ b/.github/workflows/build-clingo-binaries.yml
@@ -1,0 +1,377 @@
+name: Build and Package Python + Clingo
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version nubmer'
+        required: true
+        default: 0.0.1
+
+jobs:
+  create-release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Create Release
+        id: create-release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


CyberJML:
Build is failing cause it cannot download that package from our releases...

scripts/install-clingo.js
```diff
@@ -0,0 +1,131 @@
...
+
+    console.log('[clingo-install] Extraction complete.');
+
+    /**
+     * =======================
+     * Graphviz Installation
+     * =======================
+     */
+    const graphvizUrl =
+      'https://gitlab.com/api/v4/projects/4207231/packages/generic/graphviz-releases/12.2.1/windows_10_cmake_Release_Graphviz-12.2.1-win64.zip';
```
samutoljamo-pareto, line 102: win64 on all platforms?

CyberJML: nono, this is for windows only. Hasty test to make it work :D Maybe I should clearn it up a bit, but don't know how to do it for other yet

samilempinen:
I get 

image

Is this expected because the repo is not public?

CyberJML:
Yes, and there's a problem with postinstall. It should delete the previous installation before doing the new one

henryhaverinen (approved):
I propose that we move forward and integrate this. Further improvements are of course always possible separately.

Fuzzbender (approved):

PR #470, opened 2025-01-28
INTDEV-684 Fix preview crash

samutoljamo-pareto:
Preview mode was crashing, since we were trying to render ContentArea in preview mode, before we had the data.


Fuzzbender (approved):
Looks good to me.

PR #471, opened 2025-01-29
INTDEV-629 Separate functions to validate labels and project names

Fuzzbender:
Provide separate functions for validating labels and project names. Previously they were validated as resource names (read: identifiers). 

Labels now accept almost anything. They cannot start with whitespace and can then contain alphanumerical characters plus whitespace and '.', '_', '-'. 

Project names follow similar rules to identifiers, but should be at most 64 characters. This to avoid issues with some Windows configurations when path and filename is longer than 256 characters. The OS can be configured so that path can be 32k+characters, but we can't trust that every user would have it set up. The individual file name (sans path) can be 256 characters in Windows. 

Finally, renamed `isValidResourceName` to `isValidIdentifier` as it was actually just checking the identifier part of a resource name.

This will also fix https://cyberismo.atlassian.net/browse/INTDEV-651.

### DIFF
--- a/tools/data-handler/src/validate.ts
+++ b/tools/data-handler/src/validate.ts
@@ -389,16 +389,45 @@ export class Validate {
   }
 
   /**
-   * Validates that new name of a resource is according to naming convention.
-   * @param name: resource name
-   * returns true if name is valid, and false otherwise.
+   * Validates that new identifier of a resource is according to naming convention.
+   * @param identifier: resource identifier
+   * returns true if identifier is valid, and false otherwise.
    */
-  public static isValidResourceName(name: string): boolean {
+  public static isValidIdentifierName(identifier: string): boolean {
+    const validIdentifier = new RegExp('^[A-Za-z ._-]+$');
+    const contentValidated = validIdentifier.test(identifier);
+    const lengthValidated = identifier.length > 0 && identifier.length < 256;
+    const notInvalidIdentifier = !invalidNames.test(identifier);
+    return contentValidated && lengthValidated && notInvalidIdentifier;
+  }
+
+  /**
+   * Validates that 'name' can be used as a project name.
+   * @param name project name
+   * @returns true if name is valid, and false otherwise.
+   * @note that on Windows, if path + filename is longer than 256 characters, some file operations
+   *       are not possible. Thus, setting the maximum length of project name to 64 characters.
+   *       The 192 characters usually should be enough for the path.
+   */
+  public static isValidProjectName(name: string): boolean {
     const validName = new RegExp('^[A-Za-z ._-]+$');
     const contentValidated = validName.test(name);
-    const lengthValidated = name.length > 0 && name.length < 256;
+    const lengthValidated = name.length > 0 && name.length < 64;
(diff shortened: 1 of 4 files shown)


tools/data-handler/src/validate.ts
```diff
@@ -389,18 +389,47 @@ export class Validate {
...
-   * returns true if name is valid, and false otherwise.
+   * Validates that new identifier of a resource is according to naming convention.
+   * @param identifier: resource identifier
+   * returns true if identifier is valid, and false otherwise.
    */
-  public static isValidResourceName(name: string): boolean {
+  public static isValidIdentifierName(identifier: string): boolean {
+    const validIdentifier = new RegExp('^[A-Za-z ._-]+$');
+    const contentValidated = validIdentifier.test(identifier);
+    const lengthValidated = identifier.length > 0 && identifier.length < 256;
```
samutoljamo-pareto, line 399: Max length / validation details could we stored somewhere else in than in the function directly as we'd want to also add these to the app anyway. Though maybe it would be nice to more reusable format than constants that's also usable on the UI so I don't think it should be done in this PR but let's not forget it

tools/data-handler/src/validate.ts
```diff
@@ -389,18 +389,47 @@ export class Validate {
...
   }
 
+  /**
+   * Validates that 'name' can be used as label name.
+   * Labels are less restricted than other names, as they are never file names.
+   * @param name label name
+   * @returns true if name is valid, and false otherwise.
+   */
+  public static isValidLabelName(name: string): boolean {
+    const validName = new RegExp('^[-a-zA-Z0-9._-]+(?: [a-zA-Z0-9._-]+)*$');
```
samutoljamo-pareto, line 427: Same here

tools/data-handler/test/command-handler-create.test.ts
```diff
@@ -421,7 +421,7 @@ describe('create command', () => {
   it('try create a label - invalid character in name', async () => {
     const result = await commandHandler.command(
       Cmd.create,
-      ['label', 'decision_6', 'testö'],
+      ['label', 'decision_6', ' test'],
```
samutoljamo-pareto, line 424: This test was supposed to test for an invalid character, but for some reason letter 'ö' was removed

samutoljamo-pareto (approved):

tools/data-handler/src/validate.ts
Fuzzbender: Good idea. I'll make a ticket about it. Here: https://cyberismo.atlassian.net/browse/INTDEV-723
Please comment on the ticket if you have any ideas how to provide the details. I was initially thinking on an object with all the values included. That way UI could just instantiate the object (or if it is static, just fetch it). None of the values are dynamic, so we don't need to update the values run-time.

tools/data-handler/test/command-handler-create.test.ts
Fuzzbender: There is a whitespace before the name. That is not allowed. I'll add the 'ö' test back, was removed by accident.

PR #472, opened 2025-01-29
INTDEV-717 Fix site export on windows

CyberJML:
Windows environment does not recognize npx command without parameter shell: true. It will cause an error that is handled like successful run, so it never goes to catch block.

### DIFF
--- a/tools/data-handler/src/export-site.ts
+++ b/tools/data-handler/src/export-site.ts
@@ -12,6 +12,7 @@
 
 // node
 import fs from 'node:fs';
+import os from 'node:os';
 import { appendFile, copyFile, mkdir, writeFile } from 'node:fs/promises';
 import { mkdtempSync, writeFileSync } from 'node:fs';
 import { dirname, join } from 'node:path';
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (approved):
Looks good, but why did the export site test succeed on windows?

CyberJML:
Is the test expecting it to pass? Fail on error? Since this does not generate any errors

samutoljamo-pareto:
Oh yeah you're right. That's probably not what we want. I think it should throw if site export cannot be done for some reason. I think you can detect it based on the status code of the started process

Fuzzbender:
> Is the test expecting it to pass? Fail on error? Since this does not generate any errors

Does `spawnSync` throw? It probably doesn't but returns an error code from the attempted execution. 
Thus, the caller is not aware that there was an error. ---> `CLI` eventually says `Done` like the command would have been a success.

CyberJML:
If you console log the console.log(spawnSync('npx', ['antora', ...ad, you will see the following "error: Error: spawnSync npx ENOENT", but it's still considered to be successful spawnSync. I guess the spawnSync does not throw errors, but return objects about the execution state.

Fuzzbender (approved):

PR #474, opened 2025-01-30
INTDEV-718 Change export site to use local antora installation

CyberJML:


tools/data-handler/src/export-site.ts
```diff
@@ -13,6 +13,7 @@
 // node
 import fs from 'node:fs';
 import os from 'node:os';
+import path from 'node:path';
```
Fuzzbender, line 16: I see that this file has broken our "rule" before, but just import the function(s) you need. 
Some of these node libraries are huuuuuuge. So, just 
`port { resolve } from 'node:path';
and then change the call on row to be without `path.`

Fuzzbender (approved):
Looks good to me.

PR #475, opened 2025-01-30
Add check that template exists when adding cards to it

Fuzzbender:
When adding cards to a template, it doesn't check that the template actually exists, if its identifier matches and other template's identifier that exists in the project. :)

Unfortunately the tests - both BAT and unit tests - related to the feature were both broken in their own ways.

Add the check - use the whole name, not just identifier. 
Fix also the tests.

### DIFF
--- a/tools/data-handler/src/create.ts
+++ b/tools/data-handler/src/create.ts
@@ -31,7 +31,7 @@ import {
 } from './interfaces/resource-interfaces.js';
 import { errorFunction } from './utils/log-utils.js';
 import { readJsonFile, writeJsonFile } from './utils/json.js';
-import { Project } from './containers/project.js';
+import { Project, ResourcesFrom } from './containers/project.js';
 import { Template } from './containers/template.js';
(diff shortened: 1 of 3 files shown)


tools/data-handler/src/create.ts
```diff
@@ -131,6 +131,13 @@ export class Create extends EventEmitter {
     // Use slice to get a copy of a string.
     const origTemplateName = templateName.slice(0);
     templateName = resourceNameToString(resourceName(templateName));
+
+    const templates = await this.project.templates(ResourcesFrom.all);
```
samutoljamo-pareto, line 135: We could add `templateExists`-method to a project, since I think this is not the only function that checks if a template exists

samutoljamo-pareto (approved):

tools/data-handler/src/create.ts
Fuzzbender: We could. I will add a ticket about it.
Eventually, when templates are normal resources, it will not be possible to add cards to a non-existing template.

PR #486, opened 2025-02-03
INTDEV-722 Add Ubuntu binary install & Refactor vendor utils flow

CyberJML:


tools/data-handler/src/calculate.ts
```diff
@@ -606,16 +605,12 @@ export class Calculate {
...
-      process.platform === 'win32' ? this.pythonBinary : this.clingGraphBinary,
-      process.platform === 'win32' ? pythonArgs : clingGraphArgs,
-      {
-        encoding: 'utf8',
-        input: firstLine,
-        timeout,
-        maxBuffer: 1024 * 1024 * 100,
-      },
-    );
+    const clingraph = spawnSync(this.pythonBinary, pythonArgs, {
```
CyberJML, line 608: We talked about this with @samutoljamo-pareto and everything should still work properly. This was mandatory change to make the binary installation work properly

PR #487, opened 2025-02-03
INTDEV-709 use createElement instead of constructor directly

samutoljamo-pareto:
Using constructor in react directly is bad practice so started using createElement itself. Also created a unique key for each macro, which is done on server side, since it's much easier to do so.

Issue: Without the given changes, React crashed because the current component detected that the amount of used hooks changed. This is because the macros were not tracked separately by React. This fix should also fix a lot of other bugs that might have resulted from using constructor directly without a unique key.

### DIFF
--- a/tools/data-handler/src/macros/index.ts
+++ b/tools/data-handler/src/macros/index.ts
@@ -230,6 +230,9 @@ export function handleMacroError(
   }
 }
 
+// This is used to generate unique keys for macros
+// There might be a better way to do this
+let macroCounter = 0;
 /**
  * Creates an injectable placeholder for a macro
  * @param macro - The macro to create the placeholder for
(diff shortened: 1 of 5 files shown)


tools/data-handler/src/macros/index.ts
```diff
@@ -246,7 +249,7 @@ export function createHtmlPlaceholder(
     .join(' ');
 
   // start with a line change to ensure that passthrough ++++ is on its own line
-  return `\n++++\n<${macro.tagName}${optionString ? ` ${optionString}` : ''}></${macro.tagName}>\n++++`;
+  return `\n++++\n<${macro.tagName}${optionString ? ` ${optionString}` : ''} key="macro-${macroCounter++}"></${macro.tagName}>\n++++`;
```
Fuzzbender, line 252: Not really about this PR, but DH should not generate HTML, it should be the next.js server-side functionality.
DH should be about data only, presentation layer functionality should be above it.
No need to fix it now, but eventually this should be re-worked so that this done somewhere in app's server-side code.

Fuzzbender (approved):
Looks good to me.

tools/data-handler/src/macros/index.ts
samutoljamo-pareto: That means all macros must be handled in apps server-side code too

PR #495, opened 2025-02-06
Avoid trying to export tree-branches without cards

Fuzzbender:
Currently, `export-site` tries to handle card-branches where there are no tickets. 
To avoid unnecessary recursion, skip the call If the `children` array exists, but is empty. 

Additionally, clean up the imports a bit (combine imports from same sources and import only what is needed).

### DIFF
--- a/tools/data-handler/src/export-site.ts
+++ b/tools/data-handler/src/export-site.ts
@@ -12,12 +12,10 @@
 
 // node
 import fs from 'node:fs';
-import os from 'node:os';
-import { resolve } from 'node:path';
+import { platform, tmpdir } from 'node:os';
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (approved):
Looks good to me. Unrelated: We could use https://www.npmjs.com/package/@antora/site-generator instead of spawning a new process(the CLI version calls this library)

Fuzzbender:
> Looks good to me. Unrelated: We could use https://www.npmjs.com/package/@antora/site-generator instead of spawning a new process(the CLI version calls this library)

We could. There's also other things to improve on the export operations (they rather inefficiently export one card at a time using own recursion). I think I will make a ticket about this - there's no hurry implementing these, but if someone runs out of things to do in the future. :)

PR #496, opened 2025-02-07
INTDEV-719, INTDEV-720: Fix project renaming issues

Fuzzbender:
Once resources have been renamed, use conversion map and try to find remaining references from certain files (`*.adoc`. `*.json`, `*.hbs` and `*.lp`) and change these references. 

**Note** that this renames also "document-like" names in `*.lp` files (e.g. `cisms_task` --> `new-name_task`).

### DIFF
--- a/tools/data-handler/src/rename.ts
+++ b/tools/data-handler/src/rename.ts
@@ -13,11 +13,11 @@
 // node
 import { assert } from 'node:console';
 import { EventEmitter } from 'node:events';
-import { basename, join, sep } from 'node:path';
-import { rename, readFile, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { rename, readdir, readFile, writeFile } from 'node:fs/promises';
 
 import { Calculate } from './calculate.js';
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/rename.ts
```diff
@@ -79,17 +79,19 @@ export class Rename extends EventEmitter {
...
-    }
-    // Then rename the cards
     for (const card of sortedCards) {
+      // Attachments
+      card.content = await this.updateCardAttachments(re, card);
       await this.renameCard(re, card);
     }
   }
 
+  private scanExtensions(fileName: string): boolean {
```
CyberJML, line 89: This method works for files named "adoc", "hbs", etc. Files with no extension I mean. Not sure does it even matter...

CyberJML (approved):

tools/data-handler/src/rename.ts
Fuzzbender: True. I guess we could do `fileName.includes('.')` first. If it doesn't, just return false;

PR #498, opened 2025-02-10
INTDEV-654 remove max width from content in contentarea

samutoljamo-pareto:
We had unused space on larger screen because of max-width. For now I removed it


Fuzzbender (approved):
Looks good to me.

PR #502, opened 2025-02-11
useResizeObserver: Support all browsers

samutoljamo-pareto:


Fuzzbender (approved):

tools/app/app/lib/hooks/utils.ts
```diff
@@ -61,8 +84,8 @@ export function useResizeObserver() {
     const observer = new ResizeObserver((entries) => {
```
CyberJML, line 84: I needed to add code below to make this work properly on my Firefox

    const observer = new ResizeObserver((entries) => {
      requestAnimationFrame(() => {
        const entry = entries[0];
        setDimensions({
          width: getEntryValue(entry, 'width'),
          height: getEntryValue(entry, 'height'),
        });
      })
    });

samutoljamo-pareto: Also made chrome smoother

samutoljamo-pareto:
@Fuzzbender do you think it'd make CI too slow if we ran e2e tests on multiple browsers?

Fuzzbender:
> @Fuzzbender do you think it'd make CI too slow if we ran e2e tests on multiple browsers?
No, I think it is better than running on separate jobs. The latter eats credits in GitHub.

PR #503, opened 2025-02-13
INTDEV-588 Only read resources once when validating project

Fuzzbender:
Small optimisation to 'validate' command - each validated resource is read only once from file system.

Validation is always started from scratch (all cached `Maps` are cleared), then if a resource is needed and it is not already cached in a `Map` it is read from disk and cached to it. 

This leads to skipping ~3500 unnecessary file reads with current `main` of `module-isms-essentials`. 
With my Mac, this drops the validation command taking ~390-400ms to ~250-260ms. 

It was measured that with Windows - the same repo - the performance increase is from 1.4s to 0.7s. 

This original idea was to explore if we need the resource-cache in the future. In validation it is easy to test, since the operation is contained to one function call.  These improvements quite clearly indicate that building a resource-cache to `project` will increase the performance. Since this brought significant improvements to validation -> created this PR.

Also noticed during the implementation of this PR that changing arrays to `Map`s brought additional gains (already included in the above measurements). Thus, it is quite easy change to make that `ResourceCollector` would use `Maps` instead of arrays in its implementation.  https://cyberismo.atlassian.net/browse/INTDEV-741

### DIFF
--- a/tools/data-handler/src/validate.ts
+++ b/tools/data-handler/src/validate.ts
@@ -66,6 +66,10 @@ export class Validate {
 
   private parentSchema: Schema;
 
+  private validatedCardTypes: Map<string, CardType>;
+  private validatedWorkflows: Map<string, Workflow>;
+  private validatedFieldTypes: Map<string, FieldType>;
+
   static baseFolder: string;
   static jsonFileExtension = '.json';
   static parentSchemaFile: string;
@@ -89,6 +93,9 @@ export class Validate {
     this.directoryValidator = new DirectoryValidator();
     this.parentSchema = readJsonFileSync(Validate.parentSchemaFile) as Schema;
     this.addChildSchemas();
+    this.validatedFieldTypes = new Map();
+    this.validatedWorkflows = new Map();
+    this.validatedCardTypes = new Map();
   }
 
   // Helper to get length from types when needed.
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/validate.ts
```diff
@@ -111,6 +118,26 @@ export class Validate {
     return join(file.parentPath, file.name);
   }
 
+  // Puts resource to a local cache if found and returns the resource.
+  // If value is already cached, returns from cache.
+  private async getAndCacheResource<Type>(
+    project: Project,
+    cachedValues: Map<string, Type>,
+    valueName: string,
+  ): Promise<Type | undefined> {
```
CyberJML, line 127: This function is called within for loop at line 345, but all the resouce calls are awaited in this function and within for loop. Could we improve the performance by switching this to something like this:

    return cachedValues.get(valueName) || 
      project.resource<Type>(valueName).then((resource) => {
        if (!resource) {
          return undefined;
        }
        cachedValues.set(valueName, resource);
        return resource;
      });

Fuzzbender:
Latest measurements (timed from command execution: "cards validate"):

**main**
unified-sdl: 0.27s
cyberismo-docs: 0.33s
cyberismo-isms: 0.62s
modules-isms-essential: 0.72s
 
**feature/samimerila/optimize-validate**
unified-sdl: 0.27s
cyberismo-docs: 0.32s
cyberismo-isms: 0.57s
modules-isms-essential: 0.67s

CyberJML (approved):

tools/data-handler/src/validate.ts
Fuzzbender: Fixed as suggested. Putting all the field type promises to an array and wait them to settle. 

Slightly improves performance with more complex content:

**feature/samimerila/optimize-validate**
unified-sdl: 0.27s
cyberismo-docs: 0.32s
cyberismo-isms: 0.57s
modules-isms-essential: 0.62s

PR #509, opened 2025-02-20
Feature/samutoljamo/link improvements

samutoljamo-pareto:
done:
- INTDEV-638
I had to change the layout an accidentally stumbled upon a layout that I think looks good, but differs a bit from the original design. If you don't agree, we can change it back to the original design:
image
image

- INTDEV-704
For calculated links, an info icon is shown instead of edit/delete options.

- INTDEV-724
Made link descriptions editable. It uses the same link form as adding and the link being edited is highlighted

- INTDEV-738

### DIFF
--- a/tools/app/app/components/ContentArea.tsx
+++ b/tools/app/app/components/ContentArea.tsx
@@ -90,7 +90,10 @@ interface LinkFormSubmitData {
   cardKey: string;
   linkDescription: string;
   direction: LinkDirection;
+  previousLinkType?: string;
+  previousCardKey?: string;
   previousLinkDescription?: string;
+  previousDirection?: LinkDirection;
 }
 
 interface LinkFormData {
@@ -166,6 +169,7 @@ export function LinkForm({
     <form
       ref={formRef}
       onSubmit={handleSubmit(async (formData) => {
+        const oldLinkType = linkTypes.find((t) => t.id === data?.linkType);
         const linkType = linkTypes.find((t) => t.id === formData.linkType);
         if (!linkType) return;
         const success = await onSubmit?.({
@@ -173,8 +177,12 @@ export function LinkForm({
           cardKey: formData.cardKey,
           linkDescription: formData.linkDescription,
           direction: linkType.direction,
+          previousLinkType:
+            state === 'edit' && data ? linkType.name : undefined,
+          previousCardKey: state === 'edit' && data ? data.cardKey : undefined,
           previousLinkDescription:
-            state === 'edit' ? data?.linkDescription : undefined,
+            state === 'edit' && data ? data.linkDescription : undefined,
+          previousDirection: state === 'edit' && oldLinkType ? oldLinkType.direction : undefined,
         });
         if (success) reset();
       })}
@@ -194,7 +202,6 @@ export function LinkForm({
                   width: 180,
                 }}
                 required={true}
-                disabled={state === 'edit'}
               >
                 {linkTypes.map((linkType) => (
                   <Option key={linkType.id} value={linkType.id}>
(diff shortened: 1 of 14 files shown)


tools/app/app/components/ContentArea.tsx
```diff
@@ -70,15 +80,16 @@ type ContentAreaProps = {
...
+  linkFormState: LinkFormState;
+  onLinkFormChange?: (state: LinkFormState) => void;
 };
 
 interface LinkFormSubmitData {
   linkType: string;
   cardKey: string;
   linkDescription: string;
   direction: 'inbound' | 'outbound';
+  oldLinkDescription?: string;
```
Fuzzbender, line 92: Would `previousLinkDescription` be a better name?

tools/app/app/components/ContentArea.tsx
```diff
@@ -101,12 +114,26 @@ export function LinkForm({
...
+      linkType: data?.linkType || NO_LINK_TYPE,
+      cardKey: data?.cardKey || '',
+      linkDescription: data?.linkDescription || '',
+    },
   });
   const { t } = useTranslation();
 
+  useEffect(() => {
+    reset({
+      linkType: data?.linkType || NO_LINK_TYPE,
```
Fuzzbender, line 131: Code lines 122-124 and 131-133 are duplicates. 
Could you make a single `const` object and re-use it in both cases?

tools/app/app/lib/api/card.ts
```diff
@@ -66,18 +66,50 @@ export const useCard = (key: string | null, options?: SWRConfiguration) => {
         ))) ||
       null,
     deleteLink: async (
-      fromCard: string,
-      toCard: string,
+      target: string,
+      direction: 'inbound' | 'outbound',
```
Fuzzbender, line 70: Code `direction: 'inbound' | 'outbound'` is used quite a lot. 
In case of name changes of either direction, we'd need to make quite a few changes here and there. 
Would it be possible to define a Typescript `type` to avoid repeating the values.

henryhaverinen:
Looks great in general! A few comments, partly based on our discussion in the weekly meeting:

* The popup comment about calculated links says "This is an calculated link that cannot be deleted". Let's change this as "This is an automatic link that cannot be deleted"
* Links editing should be a pop-up dialog that allows all parts of the link to be edited
* Links could be collapsible similarly to policy check successes, including the number of links. They could be collapsed by default:

image

* I propose making the font look similar to Policy check successes as well (bold), if we're copying the look from there
* How to position the expand/collapse icon and the + icon for creating a new link? I propose putting the + icon below the expand/collapse icon, because this icon should be hidden when links are collapsed

henryhaverinen (approved):

tools/app/app/components/ContentArea.tsx
samutoljamo-pareto: done

tools/app/app/components/ContentArea.tsx
samutoljamo-pareto: done

tools/app/app/lib/api/card.ts
samutoljamo-pareto: done

samutoljamo-pareto:
@henryhaverinen feel free to try it out, all the mentioned changes are now done, and now it also automatically scrolls

samutoljamo-pareto:
one more change: autoscroll only works if you click add link from content toolbar

henryhaverinen:
> @henryhaverinen feel free to try it out, all the mentioned changes are now done, and now it also automatically scrolls

@samutoljamo-pareto looks awesome, thank you. The only thing I'd still like to change is that the links should be collapsed by default.

PR #510, opened 2025-02-24
INTDEV-578 Do not allow 'remove' when resource is in use

Fuzzbender:
Add new API `usage` to resources. This returns an array of references (card keys, resource names, or calculation file names) where the resource is used. Array is always returned so that card key references are first, then resource names and finally calculation filenames. Cards are sorted first by project prefix, then by card ID. Duplicates are removed by using a `Set`. This is not performance critical for now, thus using `Array` -> `Set` -> `Array` is fine. If it is seen as a problem, `Set` can be replaced with a `filter -> includes-indexOf` call, though it can be slower.

For future consideration. The references could be returned in subarrays (`card-references: [...], resource-references: [...], calculation-references: [...]`), if we need to use the references for anything. For now it is enough to know that the resource is still used. 

Additionally, cleaned up resource `imports` so that base classes re-`export` items that are used by inherited classes. This tidies up the imports in inherited classes, as imports mostly come from base classes.

### DIFF
--- a/tools/data-handler/src/resources/card-type-resource.ts
+++ b/tools/data-handler/src/resources/card-type-resource.ts
@@ -10,26 +10,27 @@
     License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 import {
-  AddOperation,
-  ChangeOperation,
-  RemoveOperation,
-} from './resource-object.js';
-import { Card } from '../interfaces/project-interfaces.js';
-import {
   CardType,
   CustomField,
   LinkType,
 } from '../interfaces/resource-interfaces.js';
-import { DefaultContent } from './create-defaults.js';
 import { FieldTypeResource } from './field-type-resource.js';
-import { FileResource, Operation } from './file-resource.js';
-import { LinkTypeResource } from './link-type-resource.js';
-import { Project, ResourcesFrom } from '../containers/project.js';
 import {
+  AddOperation,
+  Card,
+  ChangeOperation,
+  DefaultContent,
+  FileResource,
+  Operation,
+  Project,
+  RemoveOperation,
+  ResourcesFrom,
   ResourceName,
   resourceName,
   resourceNameToString,
-} from '../utils/resource-utils.js';
+  sortCards,
+} from './file-resource.js';
+import { LinkTypeResource } from './link-type-resource.js';
 import { Template } from '../containers/template.js';
 import { Validate } from '../validate.js';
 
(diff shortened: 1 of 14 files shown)


tools/data-handler/src/resources/card-type-resource.ts
```diff
@@ -352,6 +385,28 @@ export class CardTypeResource extends FileResource {
...
+  /**
+   * List where card type is used. This includes card metadata, card content, calculations and link type resources.
+   * Always returns card key references first, then any resource references and finally calculation references.
+   *
+   * @param cards Optional. Check these cards for usage of this resource. If undefined, will check all cards.
+   * @returns array of card keys, resource names and calculation filenames that refer this resource.
+   */
+  public async usage(cards?: Card[]): Promise<string[]> {
+    const allCards = cards ?? (await super.cards());
+    const cardReferences = [
```
CyberJML, line 397: There's quite many awaits, could we combine them like this 

    const [
      cardsWithType,
      usages,
      relevantLinkTypes,
      calculations
    ] = await Promise.all([
      this.cardsWithCardType(allCards),
      super.usage(allCards),
      this.relevantLinkTypes(),
      super.calculations()
    ])

    const cardReferences = [
      ...cardsWithType,
      ...usages,
    ].sort(sortCards);
    return [
      ...new Set([
        ...cardReferences,
        ...relevantLinkTypes,
        ...calculations,
      ]),
    ];

tools/data-handler/src/resources/field-type-resource.ts
```diff
@@ -208,6 +240,29 @@ export class FieldTypeResource extends FileResource {
...
   }
 
+  /**
+   * List where link type is used.
+   * Always returns card key references first, then any resource references and finally calculation references.
+   *
+   * @param cards Optional. Check these cards for usage of this resource. If undefined, will check all cards.
+   * @returns array of card keys, resource names and calculation filenames that refer this resource.
+   */
+  public async usage(cards?: Card[]): Promise<string[]> {
```
CyberJML, line 250: Could same "optimization" done for this too?

tools/data-handler/src/resources/card-type-resource.ts
Fuzzbender: I can implement all the `usage` calls like this. I had it at one time, but I removed to get rid of array objects. 
But then of course there are multiple `await`s .

PR #522, opened 2025-03-02
INTDEV-769 Add a new python function wrap

henryhaverinen:
Graphviz or clingraph do not wrap texts automatically, so we could do this with a python function. I tested this successfully also with some graphs.

### DIFF
--- a/calculations/common/pythonFunctions.lp
+++ b/calculations/common/pythonFunctions.lp
@@ -7,6 +7,13 @@ import textwrap
 
 class Functions:
     def concatenate(self, *args):
+        """Concatenate strings, constants or numbers as a single string
+
+        Args:
+            A variable number of Clingo Strings, Numbers, or Functions (constants)
+        Returns
+            All the arguments concatenated together as a Clingo string
+        """
         try:
             result = ""
             for arg in args:
@@ -23,6 +30,14 @@ class Functions:
             return String("")
 
     def daysSince(self, isodate):
+        """Calculates the number of dates since a date
+
+        Args:
+            isodate: a Clingo string that is a date in the ISO format
+
+        Returns:
+            A Clingo number that is the number of days since isodate
+        """
         try:
             userdate = datetime.fromisoformat(isodate.string)
             # add UTC timezone in case the isodate was just a date without time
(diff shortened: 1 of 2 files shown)


tools/data-handler/test/calculate.test.ts
```diff
@@ -148,5 +148,26 @@ describe('calculate', () => {
       });
       expect(res.results[0].key.length).to.equal(10);
     });
+    it('wraping a short string should yield the string itself', async () => {
```
Fuzzbender, line 151: A bit off comment, but isn't "wrapping" spelled with double 'p's? 
https://www.gymglish.com/en/conjugation/english/verb/to_wrap

tools/data-handler/test/calculate.test.ts
```diff
@@ -148,5 +148,26 @@ describe('calculate', () => {
...
+    it('wraping a long string', async () => {
+      const res = await calculate.runLogicProgram({
+        query:
+          'result(@wrap("This is a long string that would be too long as a title of a node in a graph")).',
+      });
+      expect(res.results[0].key).to.equal(
+        'This is a long string that<br/>would be too long as a<br/>title of a node in a graph',
+      );
+    });
+    it('wraping an empty', async () => {
```
Fuzzbender, line 166: One could have a test with wrong data type as well; e.g. `query: 'result(@wrap(5)).'

tools/data-handler/test/calculate.test.ts
```diff
@@ -148,5 +148,26 @@ describe('calculate', () => {
...
+        query: 'result(@wrap("A short string")).',
+      });
+      expect(res.results[0].key).to.equal('A short string');
+    });
+    it('wraping a long string', async () => {
+      const res = await calculate.runLogicProgram({
+        query:
+          'result(@wrap("This is a long string that would be too long as a title of a node in a graph")).',
+      });
+      expect(res.results[0].key).to.equal(
```
Fuzzbender, line 162: Generally, it is better to use the `at(0)` in the tests especially, since if there are no results, you get nicer nag than with direct index.

calculations/common/pythonFunctions.lp
```diff
@@ -35,6 +36,23 @@ class Functions:
     def today(self):
         return String(date.today().isoformat())
 
+    def wrap(self, original_text):
+        try:
+            result = ""
+            line_width = 27
```
Fuzzbender, line 42: Why 27?

Fuzzbender (approved):

Fuzzbender:
In general, new functionality and changes should have JIRA ticket ID.

henryhaverinen:
> In general, new functionality and changes should have JIRA ticket ID.

I created INTDEV-769

tools/data-handler/test/calculate.test.ts
henryhaverinen: I was not able to make this to work, because it needs to refer to the `key` property, so Typescript is not happy about the at(0) potentially being undefined

Fuzzbender: Maybe `expect(res.results.at(0)?.key)` would work?
But it is minor; there are other places in the tests that use direct index. We could fix them all with one change.

calculations/common/pythonFunctions.lp
henryhaverinen: I added a comment about this. Just based on experimentation in the graphs.

PR #524, opened 2025-03-03
INTDEV-757 Improve create new card button UI

CyberJML:
Current button looks like this

image


henryhaverinen (approved):
Looks and works great to me

Fuzzbender (approved):

PR #525, opened 2025-03-03
Rename base/fieldTypes/progress as progress

henryhaverinen:
For now, there is a compatibility rule in place as well.

I tested that both the old and the new styles of enabling progress metrics work.


samutoljamo-pareto (approved):
Looks good to me

Fuzzbender (approved):

Fuzzbender:
You could have the JIRA ticket ID in the commit message and PR title that way it is easier to connect dots if/when needed.

PR #527, opened 2025-03-04
INTDEV-598 Show usage of resource in 'show' command

Fuzzbender:
Add option `--show-use` to show command. It will only be handled with resources; otherwise ignored.

When included as an option for `show`, the returned payload JSON will include additional property `usedIn` array that lists all card-keys, resource names and calculation files that use this resource.


tools/data-handler/test/command-handler-show.test.ts
```diff
@@ -156,6 +156,19 @@ describe('shows command', () => {
...
     });
+    it('show particular card type with usage', async () => {
+      optionsDecision.showUse = true;
+      const result = await commandHandler.command(
+        Cmd.show,
+        ['decision/cardTypes/decision'],
+        optionsDecision,
+      );
+      expect(result.statusCode).to.equal(200);
+      if (result.payload) {
```
samutoljamo-pareto, line 167: test will succeed if result.payload is undefined or null

samutoljamo-pareto (approved):

PR #531, opened 2025-03-04
INTDEV-767 Remove project title from move card tree menu

samutoljamo-pareto:
I removed the title from the modal, I think it should not be there


henryhaverinen (approved):
Looks good to me

Fuzzbender (approved):

PR #532, opened 2025-03-05
INTDEV-777 Filter out calculated fields when instantiating cards

Fuzzbender:
When instantiating cards from template to JSON files, the calculated fields can just be skipped. 

Additionally, add to tests a calculated field ("obsoletedBy"). Change test assets accordingly and add a check to one of the tests that ensures that calculated field value is not copied to JSON file.


henryhaverinen (approved):
Looks good to me and works fine

PR #536, opened 2025-03-06
INTDEV-778 Transition: Write all changes to metadata in one operation

Fuzzbender:
There was a recent fix to transitions (INTDEV-774). 
The fix to that exposed a problem in current transition operation that surfaces sporadically:
- write metadata for state change
- this will cause calculations to be updated
- meanwhile, another write to same metadata is issued to update the timestamps

Sometimes either the READ op (for calculations) or WRITE op (for timestamps) will fail. 
This is probably due to racing condition trying to access the same file.

As a fix, write all changes to the metadata (state & timestamps) in one operation.

### DIFF
--- a/tools/data-handler/src/transition.ts
+++ b/tools/data-handler/src/transition.ts
@@ -15,7 +15,6 @@ import { EventEmitter } from 'node:events';
 
 import { ActionGuard } from './permissions/action-guard.js';
 import { Calculate } from './calculate.js';
-import { Card } from './interfaces/project-interfaces.js';
 import {
   CardType,
   Workflow,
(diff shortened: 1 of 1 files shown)


Fuzzbender:
Thus, with `ActionGuard` the same metadata JSON file was before fix for INTDEV-774:
- read once for Action Guard of 'transition'
- written for transition update
- read afterwards for updated calculations
- read once for for Action Guard of 'edit'
- written for timestamps update
- read afterwards for updated calculations
4 READs and 2 WRITEs for the same file for a single operation. :)

Then it was fixed so that edit command was replaced with direct write:
- read once for Action Guard of 'transition'
- written for transition update
- read afterwards for updated calculations
- written for timestamps update
- read afterwards for updated calculations
3 READs and 2 WRITEs. 

And now it is:
- read once for Action Guard of 'transition'
- written for transition update and timestamps
- read afterwards for updated calculations
2 READS and 1 WRITE

henryhaverinen (approved):
I tested in my environment, and this seems to fix INTDEV-779 as well

PR #537, opened 2025-03-07
INTDEV-454 Add statusIndicator

samutoljamo-pareto:
Added the statusIndicator and removed workflowStateCategory. It still exists for each card still, but it is currently not used anywhere. Confirmed the given logic works with cyberismo-isms(swap cardtype with cardType)

### DIFF
--- a/calculations/queries/tree.lp
+++ b/calculations/queries/tree.lp
@@ -1,4 +1,4 @@
-select("title";"workflowStateCategory";"progress";"rank";"cardType").
+select("title";"statusIndicator";"progress";"rank";"cardType").
 result(Card) :- card(Card), not parent(Card, _), not hiddenInTreeView(Card).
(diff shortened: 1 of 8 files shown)


henryhaverinen:
This looks good and works great. The only thing I'd still like to change is the spacing, if the statusIndicator is none: in this case, the layout should still reserve an empty space for the status indicator, so that all the titles will line up correctly in the three view regardless of whether there is an indicator or not. See this example, where I arranged Static code analysis to have the statusIndicator none:

image

henryhaverinen:
Even if the app doesn't need workflowStateCategory anymore, it is correct not to remove it from base.lp because it is used and useful in various reports.

henryhaverinen (approved):

henryhaverinen:
FYI: I added a default status indicator in the base module, with a possibility to disable it so other modules can define a different method if needed: https://github.com/CyberismoCom/module-base/commit/2bbd73b6f8126f78732f7c612255fdaa35f060b9

PR #546, opened 2025-03-11
INTDEV-754 Add keyboard shortcuts

samutoljamo-pareto:
Added a way to easily add keyboard shortcuts. Also added a few simple shortcuts:
Home -> First top-most card
c -> create new card modal
esc in edit --> cancel
e -> edit card

### DIFF
--- a/tools/app/app/lib/hooks/utils.ts
+++ b/tools/app/app/lib/hooks/utils.ts
@@ -10,7 +10,7 @@
     License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { useMemo, useEffect, useState, useCallback } from 'react';
+import { useMemo, useEffect, useState, useCallback, useRef } from 'react';
 import { findParentCard } from '../utils';
 import { useTree } from '../api';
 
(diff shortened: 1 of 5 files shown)


samutoljamo-pareto:
Also I noticed that our there are some issues with esm / commonjs when using jest

tools/app/app/lib/hooks/utils.ts
```diff
@@ -44,6 +44,160 @@ export function createFunctionGuard<T extends unknown[], U>(
...
+const shortcutRegistry: ShortcutRegistry = new Map();
+
+/**
+ * Generates a unique string key for a keyboard combination
+ * Used internally to store and match shortcuts so that only one check is needed per event
+ * @param combo - The keyboard combination
+ * @returns A string representation of the combination
+ */
+export function generateComboKey(combo: KeyCombo): string {
+  const modifiers = [
```
Fuzzbender, line 78: Do these map somehow to Mac keys (option, Command)?

Fuzzbender (approved):
Other than comment about Mac-specific keys, looks good to me.

tools/app/app/lib/hooks/utils.ts
samutoljamo-pareto: Meta key should be windows key / option key

samutoljamo-pareto: Then control should be the command key

PR #547, opened 2025-03-12
INTDEV-712 Support for applying changes when card state changes

Fuzzbender:
Two new internal queries added: 
1) `onTransition` which is executed when state change for a card occurs. 
2) `onCreation` which is executed when cards are created. 

These queries return both "updateFields" array that define fields that need to be changed.

**LIMITATIONS**
Currently, `cardType` and `rank` cannot be changed through queries.
The first since it would basically require the card to be re-written and all new values provided; easier to just delete and then create a new card with a new type. 
The second since I doubt that users can provide valid rank as lexorank value. 

Noticed during implementation of this feature that "static" card metadata could also be fieldtypes which would in some cases make the implementation here and there a bit simpler. Ticket: https://cyberismo.atlassian.net/browse/INTDEV-782

We don't support chaining of transition changes yet. This means that a query would change state of a card to A which would again cause another card to change state and so on. This is due to limitations of running the query in `transition` command. It would require recursion to another transition command, which I find rather distasteful. Instead, when we implement `Card` class, it can easily execute the `onTransition` query whenever its workflow state changes.

### DIFF
--- a/tools/data-handler/src/resources/field-type-resource.ts
+++ b/tools/data-handler/src/resources/field-type-resource.ts
@@ -30,6 +30,7 @@ import {
   FieldType,
 } from '../interfaces/resource-interfaces.js';
 import { CardTypeResource } from './card-type-resource.js';
+import * as EmailValidator from 'email-validator';
 
 /**
  * Field type resource class.
@@ -61,6 +62,14 @@ export class FieldTypeResource extends FileResource {
     ]);
     // Finally, write updated content.
     await this.write();
+  }
+
+  // Converts clingo array: "(option1, option2)" => ['option1', 'option2']
+  private static parseClingoArray(value: string) {
+    const itemsFromParenthesesList = /([^,()]+)/g;
+    const results = value.match(itemsFromParenthesesList);
+    if (!results) return [];
+    return results.map((item) => item.trim());
   }
 
   // Card types that use field type.
@@ -143,6 +152,7 @@ export class FieldTypeResource extends FileResource {
   /**
    * Returns all possible field types.
    * @returns all possible field types.
+   * todo: should return 'DataType' array instead of string array
    */
   public static fieldDataTypes(): string[] {
     return [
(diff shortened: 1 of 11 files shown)


tools/data-handler/src/resources/field-type-resource.ts
```diff
@@ -159,6 +167,47 @@ export class FieldTypeResource extends FileResource {
     ];
   }
 
+  /**
+   * Converts a given 'value' from Clingo result set to a type defined in 'typeName'.
+   * @param value Clingo result value (as string) to be converted.
+   * @param typeName To which type the 'value' needs to be converted to.
+   * @returns converted value.
+   */
+  public static fromClingoResult(value: string, typeName: DataType) {
```
Fuzzbender, line 176: Note to self: Add unit test to new public API

henryhaverinen:
I noticed that the App does not update the metadata fields after workflow state transitions, so even though onTransitionSetField works when the affected card is the same as the transitioned card, the results will show only after reloading the page.

Fuzzbender:
> I noticed that the App does not update the metadata fields after workflow state transitions, so even though onTransitionSetField works when the affected card is the same as the transitioned card, the results will show only after reloading the page.

This is probably due to that D-H cannot "push" updates to App. Thus, App does not know that the content (calculations mostly) have changed.

Fuzzbender:
> I tested this branch with module-base: [CyberismoCom/module-base#16](https://github.com/CyberismoCom/module-base/pull/16)
> 
> The automatic lastReviewed date worked fine, but I could not get the Create transition to work with meeting memos.

Can you share your calculations? I tested this with rather simple "update all certain type of cards" after creating a card. 
At least that worked.

henryhaverinen:
> > I tested this branch with module-base: [CyberismoCom/module-base#16](https://github.com/CyberismoCom/module-base/pull/16)
> > The automatic lastReviewed date worked fine, but I could not get the Create transition to work with meeting memos.
> 
> Can you share your calculations? I tested this with rather simple "update all certain type of cards" after creating a card. At least that worked.

Yes, they are in the pull request I linked above. Try to create a new meeting memo, and the title should include today’s date

Fuzzbender:
> > > I tested this branch with module-base: [CyberismoCom/module-base#16](https://github.com/CyberismoCom/module-base/pull/16)
> > > The automatic lastReviewed date worked fine, but I could not get the Create transition to work with meeting memos.
> > 
> > 
> > Can you share your calculations? I tested this with rather simple "update all certain type of cards" after creating a card. At least that worked.
> 
> Yes, they are in the pull request I linked above. Try to create a new meeting memo, and the title should include today’s date

This is now fixed; there was a wrong handlebar context name in the code. Also the module-base query was incorrect.

Fuzzbender:
> Works in my testing as expected. The problem that app is not showing the changed metadata values without reloading could be done in another story

Yes. It is somewhat common problem. How should we push changes from DH.

CyberJML (approved):

PR #550, opened 2025-03-14
INTDEV-772 Add resource types 'graphModels' and 'graphViews'

Fuzzbender:
Add two new resource types: graph models and graph views.
New resources support all available commands and functionalities (no calculations for them are being generate at the moment though).

Removed resource-specific `show` functions from `show` command; and instead use a generic `showResource()` instead.

**Migrator**
Unzip the package.
In the unzipped directory, execute `node migrator.js <path to the root of a project>`
The tool automatically converts existing graph models and graph views to have proper configuration files and the resource folder to have correct `.schema` file in place. After migration, please check that the migration worked correctly by running validation.  Do not migrate content repos until this PR has been merged.
[migrate_graph_resources.zip](https://github.com/user-attachments/files/19383571/migrate_graph_resources.zip)

### DIFF
--- a/tools/data-handler/src/interfaces/project-interfaces.ts
+++ b/tools/data-handler/src/interfaces/project-interfaces.ts
@@ -109,6 +109,8 @@ export interface ModuleSettings extends ProjectSettings {
   calculations: string[];
   cardTypes: string[];
   fieldTypes: string[];
+  graphModels: string[];
+  graphViews: string[];
   linkTypes: string[];
   reports: string[];
   templates: string[];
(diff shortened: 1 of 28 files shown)


tools/data-handler/src/interfaces/project-interfaces.ts
```diff
@@ -175,6 +181,8 @@ export type ResourceTypes =
   | 'cards'
   | 'cardTypes'
   | 'fieldTypes'
+  | 'graphModels'
```
Fuzzbender, line 184: Note to self: You could just add `ResourceFolderType` here (and remove unnecessary duplicates).

CyberJML (approved):

PR #552, opened 2025-03-19
INTDEV-797 Include HTML escaping in the wrap Python function

henryhaverinen:


henryhaverinen:
We can consider renaming wrap, as it would now do more than just wrap, but 1) renaming would introduce a compatibility break in content modules and 2) what would be a better name? graphTitle?

samilempinen (approved):

Fuzzbender (approved):

PR #561, opened 2025-03-25
INTDEV-805 Retrieve affected cards correctly in Calculate

Fuzzbender:
Fetch the needed details already in `Remove` (it is already fetching card details, so it could fetch ones that `Calculate` can use). Then use the card with received details in `Calculate`. Remove the extra fetch. It was only place where `Card` was used as a parameter for `findSpecificCard` and that API has broken at some point. Remove the `Card` parameter type support from `findSpecificCard` - use only card key string from now on.

### DIFF
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -442,13 +442,12 @@ export class Project extends CardContainer {
 
   /**
    * Returns specific card.
-   * It is also possible to change current card details if you provide card and different details.
-   * @param cardToFind Card key to find, or Card object
+   * @param cardToFind Card key to find
    * @param details Defines which card details are included in the return values.
    * @returns specific card details, or undefined if card is not part of the project.
    */
   public async findSpecificCard(
-    cardToFind: string | Card,
+    cardToFind: string,
     details: ProjectFetchCardDetails = {},
   ): Promise<Card | undefined> {
     let card;
@@ -460,7 +459,7 @@ export class Project extends CardContainer {
     ) {
       card = await super.findCard(
         this.paths.cardRootFolder,
-        cardToFind as string,
+        cardToFind,
         details,
       );
     }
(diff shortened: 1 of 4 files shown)


tools/data-handler/src/calculate.ts
```diff
@@ -271,15 +271,9 @@ export class Calculate {
   }
 
   // Gets either all the cards (no parent), or a subtree.
-  private async getCards(parentCard: Card | undefined): Promise<Card[]> {
+  private async getCards(card: Card | undefined): Promise<Card[]> {
```
CyberJML, line 274: Would this work? 
  private async getCards(card: Card | undefined): Promise<Card[]> {
    let cards: Card[] = [];
    if(!card) {
      return this.project.cards();
    }
    if (card && card.children) {
      cards = Project.flattenCardArray(card.children);
    }
    if (card) {
      card.children = [];
      cards.unshift(card);
    }
    return cards;
  }

Just to simplify the structure a bit

Fuzzbender: I think it could; I didn't change the function; other than removed the duplicate fetch (the data is already correct).

I guess, it could be simplified even more to 
```
private async getCards(card: Card | undefined): Promise<Card[]> {
    let cards: Card[] = [];
    if(!card) {
      return this.project.cards();
    }
    if (card.children) {
      cards = Project.flattenCardArray(card.children);
    }
    card.children = [];
    cards.unshift(card);
    return cards;
  }
```

tools/data-handler/src/containers/project.ts
```diff
@@ -442,13 +442,12 @@ export class Project extends CardContainer {
 
   /**
    * Returns specific card.
-   * It is also possible to change current card details if you provide card and different details.
-   * @param cardToFind Card key to find, or Card object
+   * @param cardToFind Card key to find
```
Fuzzbender, line 445: Note to self: Remove all `cardToFind as string` code segments from the function.

PR #562, opened 2025-03-25
INTDEV-806 Update 'cardtree.lp' correctly when card is deleted

Fuzzbender:
When a card is deleted, the card and its children should be removed from `cardTree.lp`. 
That way there is no need to update the calculations when deletion occurs. 

The existing implementation assumed that cards have relative path, when they nowadays have absolute path. Therefore, the file remained unmodified. Fixed, by using only a partial name when updating the file.

### DIFF
--- a/tools/data-handler/src/commands/calculate.ts
+++ b/tools/data-handler/src/commands/calculate.ts
@@ -391,6 +391,7 @@ export class Calculate {
       let cardTreeContent = calculationsForTreeExist
         ? await readFile(cardTreeFile, 'utf-8')
         : '';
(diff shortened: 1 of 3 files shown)


henryhaverinen (approved):
Looks clear to me. Should there be a test for this?

Fuzzbender:
> Looks clear to me. Should there be a test for this?

Yes, I can add to some remove test that it checks that the card's calculation file is deleted and the row in cardTree.lp is removed.

Fuzzbender:
There was also a bug in the original implementation if deleted card had children (those are removed as well).

PR #565, opened 2025-03-31
Add action specific spinners

samutoljamo-pareto:
Spinners are specific to actions, while updates disable buttons of similar kind globally

### DIFF
--- a/tools/app/src/components/ContentArea.tsx
+++ b/tools/app/src/components/ContentArea.tsx
@@ -113,6 +113,7 @@ interface LinkFormProps {
   inModal?: boolean;
   formRef?: React.RefObject<HTMLFormElement | null>;
   isLoading?: boolean;
+  isUpdating?: boolean;
 }
 
 const NO_LINK_TYPE = -1;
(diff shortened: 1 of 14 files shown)


samutoljamo-pareto:
Note about UX: Currently loading spinners are action specific while disabling is resource specific. For example, a card or a workflow is a resource. However, I'm not sure how much benefit there is to this kind of system. One alternative would be: a global loading state, which does not cause the button to spin, but rather indicate it globally to the user. Also, instead of a disabling a button, the actions could produce a toast if user tries something it should not do. However, let's not go back-and-fourth so this was rather just a note.

tools/app/app/lib/api/action-types.ts
```diff
@@ -0,0 +1,25 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
```
Fuzzbender, line 3: 2024 -> 2025

samutoljamo-pareto:
Waiting for #576 . This one is easier to rebase

tools/app/app/lib/api/action-types.ts
samutoljamo-pareto: done

PR #566, opened 2025-03-31
INTDEV-814 Use inline passthrough instead of block to avoid extra '+'s

Fuzzbender:
Seems that using block passthrough (q.v. https://docs.asciidoctor.org/asciidoc/latest/pass/pass-block/) inside a table leaves extra plusses on either side of macro. 

Switch to using inline passthrough (q.v. https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/); they seem to work both outside and inside of a table.

### DIFF
--- a/tools/data-handler/src/macros/index.ts
+++ b/tools/data-handler/src/macros/index.ts
@@ -248,8 +248,8 @@ export function createHtmlPlaceholder(
     .map((key) => `${key}="${options[key]}"`)
     .join(' ');
 
-  // start with a line change to ensure that passthrough ++++ is on its own line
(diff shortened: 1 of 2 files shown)


henryhaverinen:
The extra + characters are gone and all macros seem to work. However, I noticed that some macros seem to require extra newlines to make the content next to the macro work. For example, in secure development essentials, the "Manage threat models" task looks like this:

image

The asciidoc content is as follows:

`=== 2. Create a threat model document

{{#report}}
  "name": "secdeva/reports/createChildDocument",
  "template": "secdeva/templates/threatModel",
  "buttonLabel": "Create a threat model"
{{/report}}

=== 3. Document the target of the threat model

Follow the instructions in the template.

The  diagram is the most important part of this documentation.
`

PR #574, opened 2025-04-01
INTDEV-811 Optimize showProjectCards function

Fuzzbender:
Instead of function level recursion, use `readdir` with `recursive: true`.

First, builds up a card map for the cards, then fetches metadata for each card and finally builds the hierarchy. 

This takes roughly 70% of the time that the previous implementation took on my Mac. App calls this once per change (status change, or edit metadata), so the gains are cumulative.  

Noted while checking the performance that nodeJS (or Mac file system?) caches `readdir` results. If you need to make a performance improvement, call the `readdir` function once before measurements with the path you are using. That way you can compare different implementations side-by-side.

### DIFF
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -11,7 +11,8 @@
 */
 
 // node
-import { basename, dirname, join, resolve, sep } from 'node:path';
+import { Dirent } from 'node:fs';
+import { dirname, join, resolve, sep } from 'node:path';
 import { readdir } from 'node:fs/promises';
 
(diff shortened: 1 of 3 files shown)


samutoljamo-pareto (approved):
Not sure if spawning that many promises is required for the performance gains, but I don't think it will cause problems so I don't think we need to worry about that today.

Fuzzbender:
> Not sure if spawning that many promises is required for the performance gains, but I don't think it will cause problems so I don't think we need to worry about that today.

Noticed that the last `Promise.all()` is not needed, as there aren't any `async` ops done there. 
Therefore, this one has fewer `Promises` than the earlier function had.

PR #575, opened 2025-04-01
INTDEV-755 INTDEV-591 adjust page leave dialog

samutoljamo-pareto:
Logic is now simpler:
- if any field is touched, the dialog will be shown.
- If you start typing a label, but don't press the "+"-button, it will not trigger the dialog.


tools/app/app/lib/slices/card.ts
```diff
@@ -0,0 +1,35 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2024
```
Fuzzbender, line 3: We should use 2025 in copyright for new files (we cannot claim that the copyright for new implementation was started already last year).

Fuzzbender (approved):

samutoljamo-pareto:
Waiting for #576 . This one is easier to rebase

PR #576, opened 2025-04-03
Feature/samutoljamo/remove next

samutoljamo-pareto:
- Removes next
- Removes jest from app(replaced with vitest)
- Adds a separate backend(which will serve the frontend)

Note: Don't worry about the security issues at this point. The routes were setup that way for dev purposes


samutoljamo-pareto:
Looks like all security issues will resolve after this.

Fuzzbender (approved):

PR #580, opened 2025-04-07
INTDEV-689 Resource cache in Project

Fuzzbender:
Implement resource cache to `Project`. 
Integrate with `file-resources` (thus, also automatically with `folder-resources`). 

Cache is as simple as they go: key is `resourceName` which is unique for each resource and data is their JSON file content. 
As resources are not particularly large (most of them are few hundreds of bytes) this doesn't even consume that much RAM. 

The cache is updated when resource is created, updated, renamed, or deleted.


tools/data-handler/src/containers/project.ts
```diff
@@ -70,6 +70,10 @@ export class Project extends CardContainer {
   private settings: ProjectConfiguration;
   private validator: Validate;
 
+  // Created resources are held in a cache.
+  // In the cache, key is resource name, and data is resource metadata (as JSON).
+  private createdResources = new Map<string, JSON>();
```
samutoljamo-pareto, line 75: Could we simply use Record<string, string> instead of JSON?

samutoljamo-pareto (approved):
Looks good to me

PR #587, opened 2025-04-08
SECURITY FIX: Use non-greedy regex

Fuzzbender:
Code scanning tools complain about this regex that it might cause performance issues on some content. 
Replace the regex with non-greedy one.

See details from: https://github.com/CyberismoCom/cyberismo/security/code-scanning/37

### DIFF
--- a/tools/data-handler/src/macros/index.ts
+++ b/tools/data-handler/src/macros/index.ts
@@ -107,7 +107,7 @@ export function registerMacros(
  * @param input
  */
 export function macroCount(input: string): number {
-  const regex = /{{#.+}}/g;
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (commented):
Handlebars seems to allow spaces and newlines before or after and this would not catch it. The regex is still greedy, but you can make it lazy by adding `?` after `+`. `{{#[\s\S]+?}}` would for all cases without being greedy. Of course, it would be possible to craft it based on the available macros

Fuzzbender:
> `{{#[\s\S]+?}}`

Code scan is not happy with `{{#[\s\S]+?}}`

samutoljamo-pareto (approved):

PR #590, opened 2025-04-14
C api integration

samutoljamo-pareto:
Long story short: I tried WASM, but it didn't seem to work(problems with dynamic memory usage, see  https://github.com/domoritz/clingo-wasm/issues/420). Introduced another package(node-clingo), which contains a c-api based version.

### Install flow
1. See if there's a correct version of the native module in prebuilds/{os}-{arch}
2. If no, proceed to download a prebuild from github
3. If fail, try to build from source

### Building from source
When you are developing node-clingo and want to recompile it manually you can use `pnpm build-prebuildify` to replace the prebuilt version with a local build.

Dev Dependencies:
- MacOS: `brew install clingo` and simply having node seemed to work in CI
- Windows: make a conda env and activate it(environment.yml in the package root). Also requires some tools for building the c++, but they can installed when installing node(or manually, see https://www.npmjs.com/package/node-gyp#on-windows)
- Linux: `sudo apt install gringo build-essential` for clingo headers + build deps

### Runtime dependencies
- MacOS: clingo dylib is bundled and only references dylib's, which should be present in all MacOS's(libc++ and libSystem)
- Windows: clingo dll is bundled and only contains references to windows api dll's, kernel and c++ redistributable dll. C++ redistributable might not exist in windows, but this uses the latest(2015) so it usually is present in windows. This could be mentioned in cyberismo-docs, allthough just having office suite installed already installs it. To my knowledge, this should be handled by an installer
- Linux: I did not bundle .so shared library on linux so installing clingo with `sudo apt install gringo` is required. We could've also included libclingo.so, but I think it's easier for linux environment to simply depend on a system installed clingo.

### Deployment
A new CI was made this purpose. It will be only ran on the main branch and if the version in node-clingo's package json has changed compared to the latest released version. Currently, it builds the following things successfully:
- Intel Mac
- Silicon Mac
- x64 windows
- x64 linux

These are then packaged to their own tar.gz files and a new release is made.
### Testing
Currently, there's no addiotional testing made(unit testing for the node-clingo was done of course). I think it would too much to run all tests in all platforms in every single PR(native version rarely changes), but on the other hand the CI's should become much faster(we should eventually get rid of miniconda too). Maybe we could run a subset of tests in all platforms?

### CI
The CI will use prebuild if available so if you makes changes to the c++ code, but do not upgrade the version, CI won't test your new code(This is also the same behaviour in the main branch, unless you happen to build from source).

### License
should we include clingo's LICENSE with the dll and dylib?


samutoljamo-pareto:
How logic programs are sent to c++:
The node-clingo package has a function `setBaseProgram`, which takes in `program` and a key(if undefined, "default" is used). When you run the logic program, you can give a `program` and the keys of `basePrograms`, which you want to include.

Data-handler uses currently two:
- "base": contains everything including resources, except the queryLanguage.lp
- "queryLanguage": contains queryLanguage.lp

This split was made because queryLanguage is not used when running graphs. Resources could've been separated, but its such a small optimization(to not resend base.lp etc) that I did do it yet

tools/node-clingo/src/helpers.h
```diff
@@ -0,0 +1,26 @@
+#ifndef NODE_CLINGO_HELPERS_H
```
Fuzzbender, line 1: Add copyright blocks to C/C++ code as well; both to headers and source-files.

tools/node-clingo/src/helpers.h
```diff
@@ -0,0 +1,26 @@
...
+#define NODE_CLINGO_HELPERS_H
+
+#include <string>
```
Fuzzbender, line 4: Put the includes in alphabetical order

tools/node-clingo/src/helpers.h
```diff
@@ -0,0 +1,26 @@
...
+#include <vector>
+#include <sstream>
+#include <ctime>
+#include <iomanip>
+#include <clingo.h>
+namespace node_clingo {
+
+// Get the string representation of a clingo symbol
```
Fuzzbender, line 12: Function signatures could have the same comment block structure as we have on TS/JS side.

tools/node-clingo/src/helpers.cc
```diff
@@ -0,0 +1,131 @@
...
+  return result;
+}
+
+std::string html_escape(const std::string& input) {
+    std::string result;
+    result.reserve(input.size());
+    
+    for (char c : input) {
+        switch (c) {
+            case '&': result += "&amp;"; break;
```
Fuzzbender, line 39: Maybe use static string literals for the "..." values. AFAIK, now it makes temp strings for each call of this function.

tools/node-clingo/src/helpers.cc
```diff
@@ -0,0 +1,131 @@
...
+    return result;
+}
+
+std::vector<std::string> text_wrap(const std::string& text, size_t line_width) {
+    std::vector<std::string> result;
+    std::string line;
+    
+    // Split input text into words
+    std::istringstream iss(text);
+    std::vector<std::string> words;
```
Fuzzbender, line 57: Maybe call `reserve` for the vector? 
It is a lot faster to allocate before hand a lot of memory, than for each item separately. It can even "over-allocate" if you don't know the actual needed memory.

samutoljamo-pareto: We could guess the length based on the string size. Maybe length / 3 is fine

.github/workflows/prebuild.yml
```diff
@@ -0,0 +1,222 @@
+name: Build Native Prebuilds
+
+on:
+  push:
+    branches:
+      - main
+      - feature/samutoljamo/wasm-test
```
Fuzzbender, line 7: Remember to remove this one

.github/workflows/prebuild.yml
```diff
@@ -0,0 +1,222 @@
...
+          node tools/node-clingo/scripts/collect-prebuilds.js
+
+      - name: Upload combined prebuilds
+        uses: actions/upload-artifact@v4
+        with:
+          name: node-clingo-prebuilds-all
+          path: tools/node-clingo/prebuilds/
+
+      - name: Package prebuilds for each platform
+        if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/feature/samutoljamo/wasm-test')
```
Fuzzbender, line 185: Remember to remove feature branch

tools/data-handler/src/commands/calculate.ts
```diff
@@ -44,34 +38,64 @@ import {
...
     '../../../../../resources/calculations/common',
   );
-
   constructor(private project: Project) {}
 
+  // Storage for in-memory program content
+  private logicProgram: string = '';
+  private modules: string = '';
+
+  private modulesInited = false;
```
Fuzzbender, line 86: `Inited` is not proper English. `Initialized`?

tools/data-handler/src/commands/calculate.ts
```diff
@@ -86,205 +110,123 @@ export class Calculate {
...
 
-      // write card-specific logic program file
-      const filename = join(destinationFileBase, card.key);
-      const cardLogicFile = `${filename}.lp`;
-      promiseContainer.push(writeFileSafe(cardLogicFile, content));
+    let content = '% SECTION: CARDS_START\n';
+
+    for (const card of cards) {
+      const cardContent = await createCardFacts(card, this.project);
+      content += `% SECTION: CARD_${card.key}_START\n% Card ${card.key}\n${cardContent}\n% SECTION: CARD_${card.key}_END\n\n`;
```
Fuzzbender, line 121: Would it be easier to read the code, if each line was separately added? 
At least I had a hard time spotting the actual logic program line in-between the comments
```
content += `% SECTION: CARD_${card.key}_START\n`; 
content += `% Card ${card.key}\n`;
content += `${cardContent}\n`; 
content += `% SECTION: CARD_${card.key}_END\n\n`;
```

henryhaverinen:
I tried to build this with just pnpm install and pnpm build, which seemed to succeed. But then trying to launch the app resulted this:
`% cyberismo app
Error building clingo: Error: dlopen(/Users/henryhaverinen/cyberismo/cyberismo/tools/node-clingo/prebuilds/darwin-arm64/@cyberismocom+node-clingo.node, 0x0001): Library not loaded: /opt/homebrew/opt/clingo/lib/libclingo.4.dylib
  Referenced from: <D6574BFE-C65F-3C72-8EEB-42A6822A0240> /Users/henryhaverinen/cyberismo/cyberismo/tools/node-clingo/prebuilds/darwin-arm64/@cyberismocom+node-clingo.node
  Reason: tried: '/opt/homebrew/opt/clingo/lib/libclingo.4.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/clingo/lib/libclingo.4.dylib' (no such file), '/opt/homebrew/opt/clingo/lib/libclingo.4.dylib' (no such file)
    at Object..node (node:internal/modules/cjs/loader:1732:18)
    at Module.load (node:internal/modules/cjs/loader:1289:32)
    at Function._load (node:internal/modules/cjs/loader:1108:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:220:24)
    at Module.require (node:internal/modules/cjs/loader:1311:12)
    at require (node:internal/modules/helpers:136:16)
    at load (/Users/henryhaverinen/cyberismo/cyberismo/node_modules/.pnpm/node-gyp-build@4.8.4/node_modules/node-gyp-build/node-gyp-build.js:22:10)
    at file:///Users/henryhaverinen/cyberismo/cyberismo/tools/node-clingo/dist/index.js:5:15
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:578:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5) {
  code: 'ERR_DLOPEN_FAILED'
}
/Users/henryhaverinen/cyberismo/cyberismo/node_modules/.pnpm/node-gyp-build@4.8.4/node_modules/node-gyp-build/node-gyp-build.js:60
  throw new Error('No native build was found for ' + target + '\n    loaded from: ' + dir + '\n')
        ^

Error: No native build was found for platform=darwin arch=arm64 runtime=node abi=127 uv=1 armv=8 libc=glibc node=22.14.0
    loaded from: /Users/henryhaverinen/cyberismo/cyberismo/tools/node-clingo/dist

    at load.resolve.load.path (/Users/henryhaverinen/cyberismo/cyberismo/node_modules/.pnpm/node-gyp-build@4.8.4/node_modules/node-gyp-build/node-gyp-build.js:60:9)
    at load (/Users/henryhaverinen/cyberismo/cyberismo/node_modules/.pnpm/node-gyp-build@4.8.4/node_modules/node-gyp-build/node-gyp-build.js:22:30)
    at file:///Users/henryhaverinen/cyberismo/cyberismo/tools/node-clingo/dist/index.js:9:15
    at ModuleJob.run (node:internal/modules/esm/module_job:271:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:578:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)

Node.js v22.14.0
`

tools/node-clingo/lib/index.ts
```diff
@@ -0,0 +1,73 @@
+import path from 'path';
```
Fuzzbender, line 1: `... from 'node:path';`

Fuzzbender: And you only use `resolve`, so you could just import that one instead of whole `path`.

Fuzzbender, line 1: Copyright block is missing from this file and few others. Check them all.

tools/node-clingo/scripts/collect-prebuilds.js
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
+
+import fs from 'fs';
+import path from 'path';
```
Fuzzbender, line 4: import only the functions you need

tools/node-clingo/scripts/download-prebuild.js
```diff
@@ -0,0 +1,148 @@
...
+const platform = process.platform;
+const arch = process.arch;
+
+const prebuildsDir = path.join(import.meta.dirname, '..', 'prebuilds');
+const targetDir = path.join(prebuildsDir, `${platform}-${arch}`);
+const versionFilePath = path.join(targetDir, '.version');
+
+// --- Check if correct prebuilt binary version already exists ---
+let needsDownload = true;
+if (fs.existsSync(targetDir) && fs.existsSync(versionFilePath)) {
```
Fuzzbender, line 20: uses `existsSync` and then `readFileSync` for `versionFilePath` which is unnecessary. 
Just try to read the file, if it fails, it will be caught in the `catch` block.

tools/node-clingo/scripts/download-prebuild.js
```diff
@@ -0,0 +1,148 @@
...
+  if (!response.ok) {
+    throw new Error(`Failed to download: Status Code ${response.status}`);
+  }
+
+  console.log('Download successful. Extracting with tar-stream...');
+
+  const extract = tarStream.extract();
+
+  extract.on('entry', (header, stream, next) => {
+    const nameParts = header.name.split(/[\/]/).filter((p) => p);
```
Fuzzbender, line 90: preferably do not use single letter variables (`p`), but descriptive names (`part`)

tools/node-clingo/src/binding.cc
```diff
@@ -0,0 +1,338 @@
...
+// Helper function to handle Clingo errors
+void HandleClingoError(const Napi::Env& env) {
+    // If clingo returns an error, we throw an error to the javascript side
+    if (clingo_error_code() != 0) {
+        throw Napi::Error::New(env, clingo_error_message());
+    }
+}
+
+// Callback for Clingo's ground function
+bool ground_callback(clingo_location_t const *location, char const *name,
```
Fuzzbender, line 33: put each parameter to own row

tools/node-clingo/src/binding.cc
```diff
@@ -0,0 +1,338 @@
...
+    if (it != handlers.end()) {
+        return it->second(arguments, arguments_size, symbol_callback, symbol_callback_data);
+    }
+    
+    // If function name not matched, we simply do not handle it
+    return true;
+}
+
+// Callback for Clingo's solve function
+bool on_model(clingo_model_t const *model, void *data, bool *goon) {
```
Fuzzbender, line 51: if `goon` represents "go on" (i.e. to continue), then maybe better to rename it to `continue` or `go_on`

tools/node-clingo/src/binding.cc
```diff
@@ -0,0 +1,338 @@
...
+// Store base programs in a map, keyed by name
+std::unordered_map<std::string, std::string> g_basePrograms;
+
+// Silent logger function that discards all messages
+void silent_logger(clingo_warning_t code, char const *message, void *data) {
+    // Do nothing - this silences all Clingo output
+}
+
+// Helper function to handle Clingo errors
+void HandleClingoError(const Napi::Env& env) {
```
Fuzzbender, line 25: Function name style clashes with other function names (Pascal case vs snake case).

tools/node-clingo/src/binding.cc
```diff
@@ -0,0 +1,338 @@
...
+        success = false;
+    }
+    
+    delete[] atoms;
+    *goon = success;
+    return success;
+}
+
+// Wrapper for the solve event callback
+bool solve_event_callback(uint32_t type, void* event, void* data, bool* goon) {
```
Fuzzbender, line 115: If using `uint32_t` include `stdint`

tools/node-clingo/src/binding.cc
```diff
@@ -0,0 +1,338 @@
...
+#include <sstream>
+#include <chrono>
+#include <iostream>
+#include <cstring>
+#include <ctime>
+#include <iomanip>
+#include <fstream>
+#include <unordered_map>
+#include "helpers.h"
+#include "function_handlers.h"
```
Fuzzbender, line 14: Put includes in alphabetical order; system includes first, then followed by own.

tools/node-clingo/src/function_handlers.cc
```diff
@@ -0,0 +1,176 @@
...
+    void* symbol_callback_data) 
+{
+    if (arguments_size != 1) {
+        return false;
+    }
+
+    // get type
+    clingo_symbol_type_t type = clingo_symbol_type(arguments[0]);
+    if (type != clingo_symbol_type_string) {
+        // return 0
```
Fuzzbender, line 59: remove

tools/node-clingo/src/function_handlers.cc
```diff
@@ -0,0 +1,176 @@
...
+        return symbol_callback(&sym, 1, symbol_callback_data);
+    }
+    
+    const char* date_str;
+    if (!clingo_symbol_string(arguments[0], &date_str)) {
+        return false;
+    }
+    
+    // Parse ISO date string with improved function
+    std::time_t date_time = parse_iso_date(date_str);
```
Fuzzbender, line 71: Instead of using `time_t` can you not use `std::chrono` here. For example; https://en.cppreference.com/w/cpp/chrono/duration/from_stream

tools/node-clingo/src/function_handlers.cc
```diff
@@ -0,0 +1,176 @@
...
+    else if (arg_type == clingo_symbol_type_number) {
+        // skip
+    }
+    else {
+        // Unsupported type
+        return false;
+    }
+    
+    // Wrap text with line width of 27 (as specified in the original implementation)
+    std::vector<std::string> lines = text_wrap(text_to_wrap, 27);
```
Fuzzbender, line 146: For long text it would be faster if you `reserve` vector. It does not need match exactly, the allocator expands the memory as needed. For example the estimated amount of rows is `const rows = (text_to_wrap.length() / 27) + 1`.
Thus, we could `lines.reserve(rows);`

tools/node-clingo/src/function_handlers.cc
```diff
@@ -0,0 +1,176 @@
...
+    
+    // Create result with HTML escaping and <br/> tags
+    std::string result;
+    for (size_t i = 0; i < lines.size(); ++i) { 
+        result += html_escape(lines[i]);
+        if (i < lines.size() - 1) {
+            result += "<br/>";
+        }
+    }
```
Fuzzbender, line 155: To make that faster (increasing string size multiple times, can lead to allocation woes) 
```
std::ostringstream result;
for (auto it = lines.begin(); it != lines.end(); ++it) {
    result << html_escape(*it);
    if (std::next(it) != lines.end()) {
        result << "<br/>";
    }
}
return result.str();
```

tools/node-clingo/src/helpers.cc
```diff
@@ -0,0 +1,131 @@
...
+    
+    if (words.empty()) {
+        return result;
+    }
+    
+    line = words[0];
+    for (size_t i = 1; i < words.size(); ++i) {
+        if (line.length() + 1 + words[i].length() <= line_width) {
+            // Word fits on current line with a space
+            line += " " + words[i];
```
Fuzzbender, line 71: Using `+` in string operations makes temp strings. 
Also `+=` makes temp strings. `append()` does not.

Would  `line.append(" "); line.append(words[I]);` work?

Fuzzbender (approved):
I added gazillion minor comments; most of them related to editorial issues, or small performance improvements in the C++ side.

Looks exceptionally good. Well done.

henryhaverinen:
I did a small test: I run `brew install clingo`, and after that this branch worked great. It is speedy! Reports and graphs that use the external functions (@today, @concatenate, @wrap, @daysSince) also seem to work fine. Then I run `brew uninstall clingo`, after which the problem that I wrote about above returned.

samutoljamo-pareto:
Looks like rpath requires that I reference to the dynamic libraries also via rpath. I thought it's simply a lookup directory for dynamic libraries

tools/node-clingo/src/helpers.h
samutoljamo-pareto: Done

tools/node-clingo/src/helpers.h
samutoljamo-pareto: Done

tools/node-clingo/src/helpers.h
samutoljamo-pareto: Done

tools/node-clingo/src/helpers.cc
samutoljamo-pareto: done

tools/data-handler/src/commands/calculate.ts
samutoljamo-pareto: done

tools/data-handler/src/commands/calculate.ts
samutoljamo-pareto: done

tools/data-handler/test/remove.test.ts
```diff
@@ -27,29 +26,13 @@ describe('remove card', () => {
...
-      })
-      .catch(() => {
-        expect(false);
-      });
+    await removeCmd.remove('card', cardId);
 
-    // After deleting the card, check that calculations files are correctly updated.
-    cardTreeFileContent = (await readFile(cardTreeCalcFile)).toString();
-    expect(cardTreeFileContent.includes('cards/decision_5.lp')).to.equal(false);
-    // Since decision_6 is decision_5's child, it should have been removed as well.
```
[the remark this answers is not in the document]
samutoljamo-pareto: done

tools/node-clingo/lib/index.ts
samutoljamo-pareto: good observation

tools/node-clingo/scripts/collect-prebuilds.js
samutoljamo-pareto: done

tools/node-clingo/scripts/download-prebuild.js
samutoljamo-pareto: done

tools/node-clingo/src/binding.cc
samutoljamo-pareto: done, should we add a c++ linter

tools/node-clingo/src/binding.cc
samutoljamo-pareto: done

tools/node-clingo/src/binding.cc
samutoljamo-pareto: done

tools/node-clingo/src/binding.cc
Fuzzbender: Clang formatter in VS Code is pretty good.
https://code.visualstudio.com/docs/cpp/cpp-ide

tools/node-clingo/src/helpers.cc
samutoljamo-pareto: Done

tools/node-clingo/src/function_handlers.cc
```diff
@@ -0,0 +1,176 @@
...
+    
+    if (date_time == 0) {
+        // Return 0 on failure
+        clingo_symbol_t sym;
+        clingo_symbol_create_number(0, &sym);
+        return symbol_callback(&sym, 1, symbol_callback_data);
+    }
+    
+    // Calculate days difference
+    std::time_t now = std::time(nullptr);
```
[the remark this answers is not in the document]
samutoljamo-pareto: Fixed, but provided as fallback

tools/node-clingo/src/function_handlers.cc
samutoljamo-pareto: AFAIK, there's no performance benefit if you return a vector from a function anyway

samutoljamo-pareto: But I added a reserve below it for result

tools/node-clingo/src/helpers.cc
samutoljamo-pareto: Done

PR #593, opened 2025-04-16
Feature/samimerila/optimization random stuff

Fuzzbender:
Optimising the amount of calculations needed; and limiting the changes so that only card specific files are updated if only card changes.

Additionally; before app launches --> automatically generate calculations.

Removed one `generate()` call from App, and changed the other one to be specific to the card that it is changing.
Also removed `generate()` from `ActionGuard` since that should be a simple check can/cannot do an operation; DH has failed if the calculations are not up-to-date when `ActionGuard` is invoked.

Had to add one `generate()` call to `transition` but I'd like to get rid of it in the near future. 
Also, the one in App needs to go; `generate()` should only be called based on data changes in DH (plus when app starts).


tools/data-handler/src/commands/calculate.ts
```diff
@@ -484,9 +500,18 @@ export class Calculate {
    * When card changes, update the card specific calculations.
    * @param changedCard Card that was changed.
    */
-  public async handleCardChanged(changedCard: Card) {
-    await this.generate(changedCard.key);
-    return { statusCode: 200 };
+  public async handleCardChanged(changedCard?: Card | string) {
```
CyberJML, line 503: Personally I like the idea of if(!changedCard) return; to make the code simpler, but same functionality

CyberJML (approved):

Fuzzbender:
This PR has been postponed to happen after https://github.com/CyberismoCom/cyberismo/pull/590 has been merged.

PR #594, opened 2025-04-18
INTDEV-773 Create calculation files for modules, reports and templates

Fuzzbender:
Add facts from modules, reports and templates.

I decided to rename the current logo program file `modules.lp` to `calculations.lp` since the file actually collect the separate calculation files from modules. The new `modules.lp` contain the facts from the modules (basically names and prefixes).

### DIFF
--- a/tools/data-handler/src/commands/calculate.ts
+++ b/tools/data-handler/src/commands/calculate.ts
@@ -43,6 +43,10 @@ import {
   createCardTypeFacts,
   createFieldTypeFacts,
   createLinkTypeFacts,
+  createModuleFacts,
+  createProjectFacts,
+  createReportFacts,
+  createTemplateFacts,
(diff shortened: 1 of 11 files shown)


henryhaverinen (changes requested):
I reviewed the resulting logic programs with the demo project. Everything looks great, as specified in INTDEV-773, except the parent fact for the top level template cards.
For example, for the template `secdeva/templates/project`, the parent fact for the top level card is generated as
`parent(secdeva_1viupwcw, project).`
But this should be
`parent(secdeva_1viupwcw, "secdeva/templates/project").`

henryhaverinen (approved):
Now the parent thing is looking great

Fuzzbender:
> Now the parent thing is looking great

Third time's the charm. :D

PR #595, opened 2025-04-22
INTDEV-817 Remove unnecessary 'pathExists' calls

Fuzzbender:
Remove as many of the `pathExists` calls as possible. 

Generally, it is a bad code to first check if something exists and then try to read it. Instead just try to read the file / dir and handle the error / exception if any. This way we can avoid problematic timing issues when:
- check if file exists before an asynchronous read call... --> true, file exists
- before executing the read file, event loop runs some other asynchronous code that removes/renames the file
- ...original code resumes, and now fails, since file is no longer available

These sporadically cause some problems. Therefore, it is best to get rid of most of the "unnecessary" `pathExists` calls.

And it is also somewhat slower to first check the file existence and then read it than directly read it. However, we are talking about microsecond improvements, thus no one will notice.

### DIFF
--- a/tools/data-handler/src/commands/calculate.ts
+++ b/tools/data-handler/src/commands/calculate.ts
@@ -100,9 +100,7 @@ export class Calculate {
   private async generateCardTreeContent(parentCard: Card | undefined) {
     const destinationFileBase = this.project.paths.calculationCardsFolder;
     const promiseContainer = [];
-    if (!pathExists(destinationFileBase)) {
-      await mkdir(destinationFileBase, { recursive: true });
-    }
+    await mkdir(destinationFileBase, { recursive: true });
 
     const cards = await this.getCards(parentCard);
     for (const card of cards) {
@@ -179,9 +177,6 @@ export class Calculate {
 
   // Once card specific files have been done, write the the imports
   private async generateImports(folder: string, destinationFile: string) {
-    if (!pathExists(folder)) {
-      return;
-    }
     const files: string[] = getFilesSync(folder);
 
     const builder = new ClingoProgramBuilder().addComment(folder);
@@ -520,9 +515,7 @@ export class Calculate {
           this.project.paths.calculationCardsFolder,
           `${card.key}.lp`,
         );
-        if (pathExists(cardCalculationsFile)) {
-          await deleteFile(cardCalculationsFile);
-        }
+        await deleteFile(cardCalculationsFile);
         // Then, delete rows from cardTree.lp.
         filteredLines = filteredLines.filter(
           (line) => !line.includes(`cards/${card.key}.lp`),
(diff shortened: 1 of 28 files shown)


tools/data-handler/src/commands/calculate.ts
```diff
@@ -631,8 +624,8 @@ export class Calculate {
         `Graph: Failed to read image file after generating graph: ${e}`,
       );
     } finally {
-      if (pathExists(filePath)) await rm(filePath);
-      if (pathExists(filePath + '.png')) await rm(filePath + '.png');
+      await rm(filePath, { force: true });
+      await rm(filePath + '.png', { force: true });
```
samutoljamo-pareto, line 628: This is a bit risky, since clingraph sometimes produces the dot file(no extension) and it should always generate the png file. Double try catch is kind of ugly too. I think it will work more likely if the png is deleted first. This isn't the best fix, but clingraph will be replaced soon I think it is fine

tools/data-handler/src/containers/card-container.ts
```diff
@@ -88,6 +91,9 @@ export class CardContainer {
           );
         }
       }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (e) {
+      // ignore throws, if currentPaths does not have more values, recursion will stop
```
samutoljamo-pareto, line 96: Could this catch only cover readdir? If there's an error inside the loop, it will just get ignored

tools/data-handler/src/resources/report-resource.ts
```diff
@@ -73,6 +73,17 @@ export class ReportResource extends FolderResource {
     await this.write();
   }
 
+  // Try to read schema file content
+  private readSchemaFile(path: string) {
+    try {
+      const schema = readFileSync(path);
+      return JSON.parse(schema.toString());
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (e) {
```
samutoljamo-pareto, line 82: You could also remove the e variable instead using the linter disable comment

samutoljamo-pareto (approved):
Minor comments

tools/data-handler/src/commands/calculate.ts
Fuzzbender: I don't see the risk; unless the risk was already in the original code.
See: https://nodejs.org/api/fs.html#fspromisesrmpath-options
"force [<boolean>]When true, exceptions will be ignored if path does not exist."

tools/data-handler/src/containers/card-container.ts
Fuzzbender: Added two separate `try-catches` - one for each `readdir`. Looks a but ugly, but it is okay.

tools/data-handler/src/resources/report-resource.ts
Fuzzbender: Removed ALL the unnecessary `error` variables from the code. There were quite a few.

tools/data-handler/src/commands/calculate.ts
samutoljamo-pareto: My bad, did not know force skips exceptions. And yes the risk was in the code already, but it would happen if someone managed to delete the file between `pathExists` and `rm`, which would happen pretty much never. But now that rm does not throw, that issue cannot occur

PR #610, opened 2025-04-28
Add missing projectPath to env variable

CyberJML:
Environment variable needed for report macros. Application worked fine when started without cyberismo app -p <project> flag, but did not resolve path properly when -p flag was used. This environment variable was left empty when Next.js was taken out, since it was set during the initialization of next.js server


Fuzzbender (approved):
I wonder should we log an error in the `evaluateMacros()` if `context.projectPath` has not been set? 

Other than that - looks good to me.

PR #612, opened 2025-04-29
INTDEV-844 Allow public repos to be updated without username and token

Fuzzbender:
Adding even empty string for `username` and `password` (read: token) makes `git` to treat them as defined. This leads to a situation where when user did not had the environment variables (`CYBERISMO_GIT_USER` and `CYBERISMO_GIT_TOKEN`) defined, they are passed on as empty strings and then they are assumed to be the username and the token for this query. Which then leads to 401. 

Changing the implementation so that only if both environmental variables are defined, then they are added to the git url. 

Now for example adding a private repo module without credentials works correctly (i.e. you cannot do it):
Screenshot 2025-04-29 at 6 46 21

### DIFF
--- a/tools/data-handler/src/module-manager.ts
+++ b/tools/data-handler/src/module-manager.ts
@@ -75,17 +75,19 @@ export class ModuleManager {
       module.name = this.repositoryName(module.location);
     }
     const repoUrl = new URL(module.location);
-    repoUrl.username = process.env.CYBERISMO_GIT_USER ?? '';
-    repoUrl.password = process.env.CYBERISMO_GIT_TOKEN ?? '';
-    if (!repoUrl.username) {
-      throw new Error(
-        `No user defined. Cannot clone git repo. Set CYBERISMO_GIT_USER environment variable`,
-      );
-    }
-    if (!repoUrl.password) {
-      throw new Error(
-        `No git token defined. Cannot clone git repo. Set CYBERISMO_GIT_TOKEN environment variable`,
-      );
+    if (process.env.CYBERISMO_GIT_USER && process.env.CYBERISMO_GIT_TOKEN) {
+      repoUrl.username = process.env.CYBERISMO_GIT_USER ?? '';
+      repoUrl.password = process.env.CYBERISMO_GIT_TOKEN;
+      if (!repoUrl.username) {
+        throw new Error(
+          `No user defined. Cannot clone git repo. Set CYBERISMO_GIT_USER environment variable`,
+        );
+      }
+      if (!repoUrl.password) {
+        throw new Error(
+          `No git token defined. Cannot clone git repo. Set CYBERISMO_GIT_TOKEN environment variable`,
+        );
+      }
     }
     await git.clone({
       fs,
--- a/tools/data-handler/src/module-manager.ts
+++ b/tools/data-handler/src/module-manager.ts
@@ -76,18 +76,8 @@ export class ModuleManager {
     }
     const repoUrl = new URL(module.location);
     if (process.env.CYBERISMO_GIT_USER && process.env.CYBERISMO_GIT_TOKEN) {
-      repoUrl.username = process.env.CYBERISMO_GIT_USER ?? '';
+      repoUrl.username = process.env.CYBERISMO_GIT_USER;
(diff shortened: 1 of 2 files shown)


tools/data-handler/test/module-manager.test.ts
```diff
@@ -86,9 +86,7 @@ describe('module-manager', () => {
     await commands.importCmd
       .importModule(gitModule, commands.project.basePath)
       .then(() => expect(false).to.equal(true))
-      .catch((error) =>
-        expect(error.message).to.equal('HTTP Error: 404 Not Found'),
-      );
+      .catch((error) => expect(error.message).to.includes('HTTP Error:'));
```
Fuzzbender, line 89: Had to change this since if you don't provide credentials, git will give 401 for not found repos; and 404 if you have credentials. This is probably to prevent bad persons from querying what repos there are.

tools/data-handler/src/module-manager.ts
```diff
@@ -75,17 +75,19 @@ export class ModuleManager {
...
-    if (!repoUrl.username) {
-      throw new Error(
-        `No user defined. Cannot clone git repo. Set CYBERISMO_GIT_USER environment variable`,
-      );
-    }
-    if (!repoUrl.password) {
-      throw new Error(
-        `No git token defined. Cannot clone git repo. Set CYBERISMO_GIT_TOKEN environment variable`,
-      );
+    if (process.env.CYBERISMO_GIT_USER && process.env.CYBERISMO_GIT_TOKEN) {
```
CyberJML, line 78: Here we are checking if the process.env.CYBERISMO_GIT_USER && process.env.CYBERISMO_GIT_TOKEN exists, and on the next lines they are assigned to variables then checked again. Isn't this a bit unnecessary?

Could we do something like this:
if (!process.env.CYBERISMO_GIT_USER || !process.env.CYBERISMO_GIT_TOKEN) {
  throw new Error(
    `Missing environment variables. Please set both CYBERISMO_GIT_USER and CYBERISMO_GIT_TOKEN.`,
  );
}

repoUrl.username = CYBERISMO_GIT_USER;
repoUrl.password = CYBERISMO_GIT_TOKEN;

tools/data-handler/test/module-manager.test.ts
CyberJML: Can't you use the status directly?


    expect(error).to.have.property('status');
    expect(error.status).to.equal(404);

? Texts can change, status values not

tools/data-handler/src/module-manager.ts
Fuzzbender: With your code, you couldn't clone without the credentials; public repos would also need them. 
This was the point of this PR.

tools/data-handler/src/module-manager.ts
```diff
@@ -75,17 +75,19 @@ export class ModuleManager {
...
+      repoUrl.username = process.env.CYBERISMO_GIT_USER ?? '';
```
Fuzzbender, line 79: The end `... ?? ''` should be removed

tools/data-handler/src/module-manager.ts
Fuzzbender: The throws themselves (in the PR's code) are quite unnecessary, though.

tools/data-handler/test/module-manager.test.ts
Fuzzbender: Welcome to the wonderful world of TypeScript where Error's won't have `status` or `code`, or anything useful beside the message. The more specific errors do, but they depend on what is called. 

This can be solved with: 
`import { Errors } from 'isomorphic-git';`
...
```
        if (error instanceof Errors.HttpError) {
          expect(error).to.have.property('data');
          expect(error.data).to.have.property('statusCode');
          expect(error.data.statusCode).to.equal(404);
        } else {
          expect(false).to.equal(true);
        }
```

Fuzzbender: Of course now we depend on the structure that the module returns us in the call. Maybe they won't change it.

CyberJML (approved):

PR #623, opened 2025-05-02
INTDEV-549 Show report results from CLI

Fuzzbender:
Add new command `report` that enables running reports from CLI. 

Command requires that a parameters file is provided as an input data and there is optional `outputFile` is results are wanted to had into a file instead of stdout.


Fuzzbender:
Made a new PR about this, as this was using wrong JIRA ticket ID.

PR #625, opened 2025-05-06
Add ESLint rule that enforces if 'import type' could be used

Fuzzbender:
Add ESLint rule that requires that additional `type` is needed when importing in TypeScript code. This enforces that only type is imported in the type system, not in runtime. See more from: https://typescript-eslint.io/rules/consistent-type-imports/


samutoljamo-pareto (approved):
Looks good, but could you wait until C-api is merged? This is easy to rebase, but might create another extra conflict

Fuzzbender:
> Looks good, but could you wait until C-api is merged? This is easy to rebase, but might create another extra conflict

Of course. I can wait with all of my PRs until you are ready.

Fuzzbender:
I'll close this and create a new PR - I will not re-order the imports, so it is a smaller change.

PR #626, opened 2025-05-06
INTDEV-821 Provide context when there is an error from a macro

Fuzzbender:
Add context to error validation from macros.

Before 
```
> cyberismo validate -p tools/data-handler/test/test-data/valid/decision-records/
Error: Expected property name or '}' in JSON at position 2 (line 1 column 3)
```

After
```
> cards validate -p tools/data-handler/test/test-data/valid/decision-records/
Error: From card 'decision_5' a macro validation error:

Expected property name or '}' in JSON at position 2 (line 1 column 3).
Card content:
 { aaa
"name": "decision/reports/testReport",
"key3": "key"
}
```

I left out the path to the card for now, but we could add the path, by finding the card from the project and then provide the path in the error message. If reviewers think this is useful, then l can add it in this PR.


tools/data-handler/src/macros/base-macro.ts
```diff
@@ -95,7 +95,14 @@ abstract class BaseMacro {
...
 
     if (context.mode === 'validate') {
-      this.handleValidate(JSON.parse(input));
+      try {
+        this.handleValidate(JSON.parse(input));
+      } catch (error) {
+        if (error instanceof Error) {
+          const errorMessage = `From card '${context.cardKey}' a macro validation error:\n\n${error.message}.\n\nCard content:\n ${input}`;
+          throw new Error(errorMessage);
+        }
```
samutoljamo-pareto, line 104: I think we could add `throw error` here, in case there's some random library that throws something that's not an error

samutoljamo-pareto (approved):

PR #627, opened 2025-05-06
Use translated loading text instead of hard coded

Fuzzbender:
As we already have `Loading...` text, use it from translation file, instead of hard coded string.

### DIFF
--- a/tools/app/src/components/LoadingGate.tsx
+++ b/tools/app/src/components/LoadingGate.tsx
@@ -26,8 +26,9 @@ const LoadingGate: React.FC<LoadingGateProps> = ({
   loadingIndicator,
   children,
 }) => {
+  const { t } = useTranslation();
+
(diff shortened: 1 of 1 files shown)


tools/app/src/components/LoadingGate.tsx
```diff
@@ -26,7 +27,8 @@ const LoadingGate: React.FC<LoadingGateProps> = ({
   children,
 }) => {
   if (isLoading || values?.some((value) => value === null)) {
-    return loadingIndicator ? loadingIndicator : <div>Loading...</div>;
+    const { t } = useTranslation();
```
samutoljamo-pareto, line 30: Do not use the hook inside the `if`. Hooks should always be outside of conditionals

https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level

samutoljamo-pareto (approved):
Dont merge before the hook is on the top level

PR #642, opened 2025-05-08
Add custom CodeQL analysis file

Fuzzbender:
Use custom CodeQL analyser, as the default tries to scan for C++ and Python.

Default scanning has been tuned off.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,26 @@
+name: "CodeQL Analysis"
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 0 * * 0'
+
+permissions:
```
samutoljamo-pareto, line 11: We might need:
```
    permissions:
      contents: read
      security-events: write
      pull-requests: read
```

Source: https://github.com/github/codeql-action/issues/464#issuecomment-828680607

samutoljamo-pareto (approved):
Looks good if permissions are correct

.github/workflows/codeql-analysis.yml
Fuzzbender: Github itse sanoi, että pitäisi riittää nuo. "Since the workflow involves checking out the repository and performing a CodeQL analysis, the contents: read permission is sufficient.

PR #657, opened 2025-05-12
INTDEV-803 Replace clingraph

samutoljamo-pareto:
This will get rid of clingraph by replacing it with a cyberismo report. Report is built in to the source. Also includes some fixes to the error handling of c-api, gcc < 13 fall and a basic readme for the native module


samutoljamo-pareto:
Feel free to review code, all functionality that remains to be done should happen in the logic programs

tools/node-clingo/src/binding.cc
```diff
@@ -229,16 +229,20 @@ Napi::Value SetBaseProgram(const Napi::CallbackInfo &info)
...
     {
-        // Let Napi errors propagate as they are
-        throw e;
+        e.ThrowAsJavaScriptException();
+        return env.Undefined();
     }
     catch (const std::exception &e)
     {
-        throw Napi::Error::New(env, e.what());
+        auto error = Napi::Error::New(env, e.what());
```
Fuzzbender, line 237: Add `const` --> `const auto error = ...`
(and similarly for other places.

tools/node-clingo/src/function_handlers.cc
```diff
@@ -133,17 +139,17 @@ namespace node_clingo
...
         time_t now = time(nullptr);
         std::string today_str;
         std::stringstream ss;
         ss << std::put_time(std::localtime(&now), "%Y-%m-%d");
         today_str = ss.str();
+#else
+        const auto now_point = std::chrono::system_clock::now();
+        const auto current_zone = std::chrono::current_zone();
+        const std::chrono::zoned_time zt{current_zone, now_point};
+        std::string today_str = std::format("{:%Y-%m-%d}", zt);
```
Fuzzbender, line 152: I think this could also be `const auto`

tools/data-handler/package.json
```diff
@@ -53,6 +53,7 @@
     "json-schema": "^0.4.0",
     "jsonschema": "^1.5.0",
     "mime-types": "^3.0.1",
+    "node-graphviz": "^0.1.1",
```
CyberJML, line 56: is there any alternatives? This is 3 years old publish and possibly not going to update anymore?

samutoljamo-pareto: I did not find any wasm or native alternatives. The package we use is based on https://hpcc-systems.github.io/hpcc-js-wasm/getting-started.html . However, that package includes a ton of other deps. Graphviz itself does not many updates: https://github.com/graphp/graphviz . But you're right, that package likely won't be updated. So we'd probably need to create our own graphviz webassembly build, if that happens. However, I feel like there's no reason to do that unless we run into issues

tools/node-clingo/src/function_handlers.cc
samutoljamo-pareto: Done

Fuzzbender: Why?

tools/node-clingo/src/function_handlers.cc
```diff
@@ -133,17 +139,17 @@ namespace node_clingo
...
-        const auto current_zone = std::chrono::current_zone();
-        const std::chrono::zoned_time zt{current_zone, now_point};
-        std::string today_str = std::format("{:%Y-%m-%d}", zt);
-#else
+#if USE_FORMAT_FALLBACK
         time_t now = time(nullptr);
         std::string today_str;
         std::stringstream ss;
         ss << std::put_time(std::localtime(&now), "%Y-%m-%d");
         today_str = ss.str();
```
Fuzzbender, line 147: Even though this hasn't changed; could this be: 
```
std::string today_str = ss.str();
```
And potentially `const` and potentially `auto` instead of provided type.

resources/graphvizReport/index.adoc.hbs
```diff
@@ -0,0 +1,34 @@
...
+{{#if this.isNode}}
+  "{{this.cardKey}}" [{{#each attrs}}{{this.attr}}="{{{this.value}}}" {{/each}}]
+{{/if}}
+{{#if this.isGraph}}
+  {{> subgraph}}
+{{/if}}
+{{/each}}
+}
+{{/inline}}
+{{#each results}}
```
Fuzzbender, line 17: General comment on handlebars code. 

Shouldn't it be still indented if there are nested statements?
Ie.  each-within-each should be:
```
{{#each results}}
  {{#each digraphs}}
  ...
  {{/each}}
  ...
{{/each}}
```

It makes reading the HB-code a bit easier.

resources/graphvizReport/model.lp
```diff
@@ -0,0 +1,141 @@
+% model
+% note: not escapted
```
Fuzzbender, line 2: typo --> escaped

tools/node-clingo/src/function_handlers.cc
samutoljamo-pareto: I thought you meant, zoned_time, but yeah the string could be const auto

tools/node-clingo/src/function_handlers.cc
samutoljamo-pareto: Done

tools/node-clingo/src/function_handlers.cc
samutoljamo-pareto: Changed it

resources/graphvizReport/model.lp
samutoljamo-pareto: Done

resources/graphvizReport/index.adoc.hbs
samutoljamo-pareto: They probably were done like that, because then there would be extra white. But yes it would be more readable, I'll add indentation

tools/data-handler/src/utils/sanitize-svg.ts
```diff
@@ -25,19 +25,17 @@ const removeSvgWidthAndHeight = (node: Element) => {
...
-  const dirty = buffer.toString('utf-8');
-
   DOMPurify.setConfig({ USE_PROFILES: { svg: true } });
   DOMPurify.addHook('afterSanitizeAttributes', removeSvgWidthAndHeight);
 
-  let cleaned = DOMPurify.sanitize(dirty);
+  let cleaned = DOMPurify.sanitize(svg);
 
   // Remove link titles, quick fix for Clingraph/Graphviz generated titles for links that are quite strange
   cleaned = cleaned.replace(/\s*xlink:title=(["']).*?\1/g, '');
```
henryhaverinen, line 41: I wonder if this workaround is still needed when we're not using Clingraph?

resources/graphvizReport/model.lp
henryhaverinen, line 1: Since this is really a report, the name of this file could be query.lp or query.lp.hbs, similarly to regular reports.

tools/data-handler/src/command-manager.ts
```diff
@@ -50,8 +50,8 @@ export class CommandManager {
     this.project = new Project(path);
     this.validateCmd = Validate.getInstance();
 
-    this.showCmd = new Show(this.project);
     this.calculateCmd = new Calculate(this.project);
+    this.showCmd = new Show(this.project, this.calculateCmd);
```
Fuzzbender, line 54: Just an idea for the future, to avoid problems passing commands to each others. No need to make any changes in this PR; just something to think about...

It might make sense to separate "calculation engine" and "calculation command". And the `project` would own "calculation engine". This would avoid passing command instances to other commands and only pass `project` to them [Yes, there's also `showCmd` and `createCmd` being passed in few cases, but `calculateCmd` is basically passed to all commands now]

tools/data-handler/src/commands/calculate.ts
```diff
@@ -342,30 +357,18 @@ export class Calculate {
...
       // For queries, use both base and queryLanguage
-      const basePrograms =
-        argMode === 'query'
-          ? [BASE_PROGRAM_KEY, QUERY_LANGUAGE_KEY]
-          : BASE_PROGRAM_KEY;
+      const basePrograms = [BASE_PROGRAM_KEY, QUERY_LANGUAGE_KEY];
 
       // Then solve with the program - need to pass the program as parameter
-      return solve(data.query as string, basePrograms);
+      return solve(query as string, basePrograms);
```
Fuzzbender, line 366: `query` is already a `string`. Is the `as string` needed here?

tools/data-handler/src/commands/export.ts
```diff
@@ -206,6 +206,7 @@ export class Export {
           projectPath,
           cardKey: card.key,
         },
+        this.calculateCmd,
```
Fuzzbender, line 209: This is not used anywhere, is it?

tools/data-handler/src/macros/index.ts
```diff
@@ -28,15 +28,23 @@ import type {
 } from '../interfaces/macros.js';
 import type BaseMacro from './base-macro.js';
 import TaskQueue from './task-queue.js';
-
+import type { Calculate } from '../commands/index.js';
 const CURLY_LEFT = '&#123;';
 const CURLY_RIGHT = '&#125;';
 
-export interface MacroConstructor {
+export interface SimpleMacroConstructor {
```
Fuzzbender, line 35: Add comments to the exported interfaces what they are and what they are for

tools/data-handler/src/utils/report.ts
```diff
@@ -0,0 +1,79 @@
...
+    handlebars.registerHelper('formatAttributeValue', formatAttributeValue);
+  }
+
+  return handlebars.compile(contentTemplate)({
+    ...options,
+    ...result,
+  });
+}
+
+export function formatAttributeValue(value?: string) {
```
Fuzzbender, line 69: Add comment block to this exported function. 
Functions should be in alphabetical order in the file.

Fuzzbender (approved):
Some minor editorial comments. Looks good to me.

henryhaverinen:
The branch seems to work great and I think the performane is a bit better as well.

The only observation I have is that the titles in some places are aligned too much to the left, like this in the card demo_7s6siekw of cyberismo-demo:

image

henryhaverinen (approved):
Let's not integrate yet, but only once we have made sure that the related change in secure development essentials has been adopted

tools/data-handler/src/utils/sanitize-svg.ts
samutoljamo-pareto: the width and height is still there, but I did not find any xlink:title 's from the generated svg. @CyberJML did you take a look at https://www.npmjs.com/package/sanitize-html ? Seems like very similar to dompurify, but with it we could get rid of jsdom, which is a large portion of our node_modules.

resources/graphvizReport/model.lp
samutoljamo-pareto: Yes

tools/data-handler/src/command-manager.ts
samutoljamo-pareto: Yes the current calculateCmd is kind of a calculate engine that is also used as a command. Since we have this class based approach, a proper IoC container would be nice, because now I feel like we have to avoid dependencies as much as possible, because we have to pass them manually. For example, since the report macro now requires CalculateCmd, it is passed to all macros. If some other macro needs validate, then we need to again change the constructor for all macros or make a separate function / class just for creating a macro class

Fuzzbender: I'll create a ticket about it. While at it; validation command should be split to "command" and "validation engine" parts (similar reasons as Calculation).

PR #658, opened 2025-05-12
INTDEV-833 Add labels to card view

samutoljamo-pareto:
Labels was missing the "forceValue" field


Fuzzbender:
Do we have any easy ways to check the labels in UI tests so that they are there and visible?

Fuzzbender (approved):

samutoljamo-pareto:
Yes, added a test

PR #660, opened 2025-05-12
INTDEV-848 - Use 'import type' when type import is enough

Fuzzbender:
I decided to re-create the PR and not reorder the imports to make the change a smaller one.

Additionally, removed one circular dependency between resource-utils and file-resource files.
There were two methods that used `FileResource` and thus it was imported to `resource-utils` and `file-resource` imported these methods. Fixed it so that other method was removed (not used anywhere) and the other was moved inside the `file-resource` itself.


samutoljamo-pareto (approved):
Looks good to me

PR #662, opened 2025-05-12
Fix: Calc run does not work after C-api

samutoljamo-pareto:
Calc run did not work, because we the file as a filepath instead of reading it.


henryhaverinen (approved):
Works great for me

PR #663, opened 2025-05-12
INTDEV-852 Cyberismo calc testing

samutoljamo-pareto:
Added a test for `cyberismo calc run`. Tested that it broke without the fix, which is already in main. Also added a deprecated warning to `cyberismo calc generate`, because it should be removed. But it will still provide a 0 exit code so that it does not break existing CIs

### DIFF
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -211,16 +211,13 @@ const calculate = program
 
 calculate
   .command('generate')
-  .description('Generate a logic program')
+  .description('DEPRECATED: Generate a logic program')
   .argument('[cardKey]', 'If given, calculates on the subtree of the card')
   .option('-p, --project-path [path]', `${pathGuideline}`)
-  .action(async (filePath: string, options: CardsOptions) => {
-    const result = await commandHandler.command(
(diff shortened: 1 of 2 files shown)


tools/cli/src/index.ts
```diff
@@ -211,16 +211,13 @@ const calculate = program
 
 calculate
   .command('generate')
-  .description('Generate a logic program')
+  .description('DEPRECATED: Generate a logic program')
```
Fuzzbender, line 214: There's NodeJS's own deprecation utility. Maybe it could be used here?
https://nodejs.org/api/util.html#utildeprecatefn-msg-code

Fuzzbender (approved):

CyberJML (approved):

samutoljamo-pareto:
@henryhaverinen DH does not have that functionality right now, but of course we could implement it easily.

tools/cli/src/index.ts
samutoljamo-pareto: I think not in this case. That would be good if we deprecated a function, but now we're deprecating a CLI command.

PR #669, opened 2025-05-15
INTDEV-752 Hide cards in linkform if link already exists

samutoljamo-pareto:
Added a few tests also

### DIFF
--- a/tools/app/src/lib/utils.ts
+++ b/tools/app/src/lib/utils.ts
@@ -34,42 +34,151 @@ import { CardResponse } from './api/types';
 type ItemType<T> = T extends (infer U)[] ? U : never;
 
 /**
- * Flattens the Card tree into a single array of Cards
- * @param tree: array of Cards with possible children Card arrays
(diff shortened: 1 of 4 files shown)


tools/app/src/lib/utils.ts
```diff
@@ -421,3 +424,88 @@ export function expandLinkTypes(
   }
   return result;
 }
+/**
+ * Creates a predicate function from a base function and a list of initial arguments
+ * @param baseFunction: base function to create the predicate from
+ * @param args: initial arguments for the base function
+ * @returns predicate function
+ */
+export function createPredicate<
```
Fuzzbender, line 433: exported functions should be in alphabetical order

Fuzzbender (approved):

tools/app/src/lib/utils.ts
Fuzzbender: Related; we could enforce this as well: https://eslint.org/docs/latest/rules/sort-imports

PR #670, opened 2025-05-15
INTDEV-845 Update cache values when data changes

Fuzzbender:
Added a file watcher for running concurrent instances of a project on same machine. 
This will not work in Docker and is not intended to.

To avoid watching the filesystem unnecessarily (normal case), there is an additional flag for `cyberismo app` `--concurrent-apps` which causes the instance to watch changes in `.cards/local`. Module changes are not watched, as they should not be modified. File renames are ignored. 

When a file change is noticed, it is checked if a resource file was changed. If so, then resource is removed from cache and recreated. Then pushed into cache again.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -50,6 +50,7 @@ import { type Level } from 'pino';
 export interface CardsOptions {
   details?: boolean;
   forceStart?: boolean;
+  watchResourceChanges?: boolean;
   projectPath?: string;
   repeat?: number;
   showUse?: boolean;
@@ -123,7 +124,7 @@ export class Commands {
     const creatingNewProject = command === Cmd.create && args[0] === 'project';
     if (!creatingNewProject) {
       try {
-        await this.doSetProject(options.projectPath || '', options.logLevel);
+        await this.doSetProject(options);
       } catch (error) {
         return { statusCode: 400, message: errorFunction(error) };
       }
@@ -149,7 +150,8 @@ export class Commands {
   }
 
   // Handles initializing the project so that it can be used in the class.
-  private async doSetProject(path: string, level?: Level) {
+  private async doSetProject(options: CardsOptions) {
+    const path = options.projectPath || '';
     this.projectPath = resolveTilde(await this.setProjectPath(path));
     if (!Validate.validateFolder(this.projectPath)) {
       let errorMessage = '';
@@ -165,7 +167,10 @@ export class Commands {
       throw new Error(`Input validation error: cannot find project '${path}'`);
     }
 
-    this.commands = await CommandManager.getInstance(this.projectPath, level);
+    this.commands = await CommandManager.getInstance(this.projectPath, {
+      logLevel: options.logLevel,
+      watchResourceChanges: options.watchResourceChanges,
+    });
     if (!this.commands) {
       throw new Error('Cannot get instance of CommandManager');
     }
--- a/tools/data-handler/src/containers/project.ts
+++ b/tools/data-handler/src/containers/project.ts
@@ -64,6 +64,9 @@ import { ReportResource } from '../resources/report-resource.js';
 import { TemplateResource } from '../resources/template-resource.js';
 import { WorkflowResource } from '../resources/workflow-resource.js';
 
+import { ContentWatcher } from './project/project-content-watcher.js';
+import { pathToResourceName } from '../utils/resource-utils.js';
+
 // Re-export this, so that classes that use Project do not need to have separate import.
 export { ResourcesFrom };
 
@@ -75,12 +78,16 @@ export class Project extends CardContainer {
   private projectPaths: ProjectPaths;
   private settings: ProjectConfiguration;
   private validator: Validate;
+  private resourceWatcher: ContentWatcher | undefined;
 
   // Created resources are held in a cache.
   // In the cache, key is resource name, and data is resource metadata (as JSON).
   private createdResources = new Map<string, JSON>();
 
-  constructor(path: string) {
+  constructor(
+    path: string,
+    private watchResourceChanges?: boolean,
+  ) {
     super(path, '');
 
     this.settings = new ProjectConfiguration(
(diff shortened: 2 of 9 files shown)


tools/data-handler/src/containers/project.ts
```diff
@@ -93,6 +96,48 @@ export class Project extends CardContainer {
     // todo: implement project validation
     this.validator = Validate.getInstance();
     this.resources.collectLocalResources();
+
+    const ignoreRenameFileChanges = true;
+
+    // Watch changes in .cards if there are multiple instances of Project being
+    // run concurrently.
+    if (process.env.CONCURRENT_APPS === 'true') {
+      new ContentWatcher(
```
samutoljamo-pareto, line 105: I wonder if that is a memory leak, because of the watch event. When project is garbage collected, I don't think the contextwatcher is destroyed as it calls the watch method from node fs. Since the anynomous function given to contextwatcher holds a reference to project, project will never be garbage collected.

samutoljamo-pareto (commented):
Did not work on windows. I used 2 instances of cyberismo app with --concurrent-apps and after updating a linkType, did not update the link type even after refreshing. Started a third instance, which had the updated value.

tools/data-handler/src/containers/project.ts
Fuzzbender: Do you think if it is assigned then it would? (e.g. `const tadaa = new ContentWatcher(...)` )
This is mostly theoretical, as it would only affect if someone would use the second instance and only for that instance.

tools/data-handler/src/command-handler.ts
```diff
@@ -113,6 +114,9 @@ export class Commands {
     args: string[],
     options: CardsOptions,
   ): Promise<requestStatus> {
+    if (options.concurrentApps) {
+      process.env.CONCURRENT_APPS = 'true';
```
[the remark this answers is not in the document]
Fuzzbender: Yes. I was a bit lazy as I didn't want to push the value through the whole stack. But maybe I'll put into `ProjectOptions` or something like that.

Fuzzbender:
> Did not work on windows. I used 2 instances of cyberismo app with --concurrent-apps and after updating a linkType, did not update the link type even after refreshing. Started a third instance, which had the updated value.

The use case I was using was:
1. Launch app normally `cyberismo app` on `cyberismo-isms`. 
2. Navigate to Progress page and open that card.
3. Launch a second app; `cyberismo app --concurrent-apps` 
4. Edit one of the resources: 
  `cyberismo update cisms/cardTypes/register add customFields "{ \"name\": \"base/fieldTypes/identifier\" }"`
  `cyberismo update cisms/cardTypes/register add alwaysVisibleFields \"base/fieldTypes/identifier\"`
  
Immediately after #4 the page of first application is updated.

tools/data-handler/src/containers/project.ts
samutoljamo-pareto: That will also not work. We could use a signal for telling the watcher to stop listening: https://nodejs.org/docs/latest/api/fs.html#fspromiseswatchfilename-options . Macros still create a new project each time they execute so it is not only theoretical. But technically we could have some kind of cleanup function which sends the abort signal.

Fuzzbender: The macros own `Project` should be fixed separately. We should just have one Project instance; now `Project` has resources (=macros) that create projects in some cases. And these projects have macros, which in turn have projects, ...

I think I will implement the `ProjectOptions` that the CLI will pass to `Project` instances that it uses.
Then macros' `Projects` have no need for the value. And add the signal as well.

samutoljamo-pareto: Since we've got the class based structure, do you think we should start using some kind of library to do DI for us? So we don't get "constructor madness"

samutoljamo-pareto:
Do you mind if I make an e2e test for this?

tools/data-handler/src/containers/project.ts
Fuzzbender: I have to check that. In general, I'd like to get rid of dependencies. :)

Fuzzbender:
Changes: 
1. Renamed the option from `--concurrent-apps` to `--watch-resource-changes` 
2. No longer uses environmental variables, but pushes options through the stack
3. Should no longer cause memory leaks. Class itself can be closed; owner of the class calls this when that class is disposed of, and that class is disposed of when process is killed (interrupted or cancelled).

samutoljamo-pareto:
The above does not work, because `attachCommandManager` function creates an instance of command manager on its own. Either cli should inject it or give the flag also to the backend

tools/data-handler/src/containers/project/project-content-watcher.ts
```diff
@@ -28,14 +29,29 @@ export class ContentWatcher {
...
+    });
     this.watcher.unref();
   }
+
+  /**
+   * Close content watcher. Removes watchers.
+   */
+  public close(): void {
+    this.abortController.abort();
+    this.watcher?.close();
```
samutoljamo-pareto, line 55: Isn't this redundant as abort controller should stop the watcher

samutoljamo-pareto:
Also about the memory leaks: The issue is not at when the process is destroyed, but rather at when the instances should be destroyed. I think also it's a bad habit to have those process exit call there as it might create annoying bugs which cause the tests to succeed / they prevent other handlers from running. I think the `unref` function call was not idea as it makes sure the event loop won't be kept alive if there's no other activity. With that we do not need those calls as at that point memory will be freed anyway(not our issue).

The issue is that we prevent GC from freeing up memory during runtime itself. Of course, this will now not happen as we only have a single instance that creates the watcher. Except if we created new instances of commandmanager, but currently that is just theoretical. Wouldn't hurt to leave a comment to the `getInstance`-method saying that this leaks memory if called repeatedly(I tested manually that there is indeed a memory leak)

samutoljamo-pareto:
Tried calling the dispose method, but looks like the `process.on` methods also create a similar leak. Best to just remove them and use unref. With both dispose calls before creating new instances and no `process.on` calls, memory seems usage seems to be pretty constant

Fuzzbender:
> The above does not work, because `attachCommandManager` function creates an instance of command manager on its own. Either cli should inject it or give the flag also to the backend

I can pass the options to there as well; but the implementation shouldn't create a new instance of `CommandManager`. The new instance is currently created solely based on path used. 

I will also create a ticket about the CM memory leak (it is not related to this). I think I will just throw if instance is wanted with different path. Alternatively, getting instance is without path, and you can set it separately (and if path changes, it should re-create `Project`)

samutoljamo-pareto:
Actually, I think I looked wrong. Since cli is called first, it will call the getInstance method before app so app will end up using the same instance so it should create another instance. So if you remove the process.on methods + use unref, there should only be a "theoretical" memory leak, which won't happen in practice

PR #671, opened 2025-05-16
Change the tests to use git URLs for module imports

Fuzzbender:
Change the module imports to use git-import instead of file-import in various tests.

### DIFF
--- a/tools/app/cypress.config.ts
+++ b/tools/app/cypress.config.ts
@@ -26,7 +26,7 @@ export default defineConfig({
 
           // Create test project from module-base
           execSync(
-            'cd ../../.tmp&&cyberismo create project "Basic Acceptance Test" bat cyberismo-bat&&cd cyberismo-bat&&cyberismo import module ../module-base&&cyberismo create card base/templates/page',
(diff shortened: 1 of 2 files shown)


samutoljamo-pareto (approved):
Looks good, but could also changing the attachments to come from inside the project(copy from repo) or the internet so it works with containers

Fuzzbender:
> Looks good, but could also change the attachments to come from inside the project(copy from repo) or the internet so it works with containers

Can you point where it doesn't use attachment from the repo? 
At least the CLI test with attachments is using data from repo: https://github.com/CyberismoCom/cyberismo/blob/main/tools/cli/test/cli.test.ts#L238
Attachments do not support URLs AFAIK, so one would need to copy the file from the internet in the tests.

samutoljamo-pareto:
Exactly the line you pointed does use a attachment from the monorepo, but not from within the "test project". Only the project folder is mounted to the docker container, so it cannot use the relative paths to find the file.

Fuzzbender:
> Exactly the line you pointed does use an attachment from the monorepo, but not from within the "test project". Only the project folder(< project root>/.tmp/xxx) is mounted to the docker container, so it cannot use the relative paths to find the file.

The "test project" cannot contain any attachments since it is created a few steps earlier (on row: https://github.com/CyberismoCom/cyberismo/blob/main/tools/cli/test/cli.test.ts#L69). 
They would need to be created adhoc, or copied from somewhere through an URL. 

I'll skip modifying the attachment test for now.

samutoljamo-pareto:
They can just be copied, but I can add it to the CI PR

PR #674, opened 2025-05-21
INTDEV-830 create link in create cards macro

samutoljamo-pareto:
Links will be created for all top level cards and a message will be shown that tells how many links have been created or not(so no mention which as it would not fit to the notification). Maybe we'd need to create some kind possibility to open more detailed version of the notifications. Most common reason to fail would be attempting to create a link which does not support the given cardtypes.

### DIFF
--- a/tools/app/__tests__/utils.test.ts
+++ b/tools/app/__tests__/utils.test.ts
@@ -22,6 +22,7 @@ import {
   isCardTypePermittedForLinkType,
   isAlreadyLinked,
   canCreateLinkToCard,
+  parseNestedDataAttributes,
 } from '@/lib/utils';
 
 test('flattenTree works with test data', async () => {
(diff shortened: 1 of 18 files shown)


henryhaverinen:
I tried to test this feature, so I made a report that produces a `createCards` macro that would create the new card somewhere else, and it would use this feature to add a link to the card that had the `createCards` macro. See the branch feature/henryhaverinen/INTDEV-830 of module-secure-development-essentials. If you create a threat model, create some processes or data stores in the threat model, and then use the buttons to create security issues, you can see that they will be created under the process or data store, even though I tried to make the issues go under "List of identified issues".

However, this didn't work, because adding a `cardKey` parameter to the `createCards` macro does not seem to change the location where the new card will be created. I tried this also with a simpler example, just without wrapping `createCards` within a report.  Instead of using the provided `cardKey`,  the `createCards` macro  seems to always create the new cards under the card that had the `createCards` macro.

I think this is a bug. Maybe it would make sense to fix this bug at the same time with this change?

samutoljamo-pareto:
@henryhaverinen I think it's because those attribute names are lowercased. I'll fix it

samutoljamo-pareto:
Seems there is something wrong with the report itself. "listOfThreats" is probably empty, because it started to work after I hardcoded it to the report

Fuzzbender (approved):

PR #677, opened 2025-05-22
Use cyberismo-demo instead of cyberismo-docs for CLI unit tests

Fuzzbender:
Using `cyberismo-docs` is a bit rough; we sometimes need to push changes that won't work with current `main` (since project is used in CLI unit tests) to allow upcoming PRs build correctly. 

Using our official documentation build for this meant that sometimes the online documentation site would break until the breaking-PR would have been merged. 

Thus, using `cyberismo-demo` instead for tests. In the long run, CLI unit tests should use module and project from within the mono-repo itself.


tools/cli/test/cli.test.ts
```diff
@@ -6,10 +6,9 @@ import { exec, execSync } from 'child_process';
 import { log } from 'console';
 import { existsSync, rmSync } from 'fs';
 
-// CI should clone the module-base and cyberismo-docs repositories to these paths
-// Locally run test clone the repositories to these paths in before hook
+// CI should clone the module-base to this path
+// Locally run test clone the repository to this path in before hook
```
samutoljamo-pareto, line 10: "Locally clones the repository to this path in before hook"?

samutoljamo-pareto (approved):

PR #680, opened 2025-05-26
INTDEV-579 Allow renaming or removing workflow states with update

Fuzzbender:
This change enables operations like: 

State rename:
`cyberismo update cisms/workflows/page change states "{\"name\": \"Approved\", \"category\":\"closed\"}" "{\"name\": \"ApprovedNew\", \"category\":\"closed\"}"` 
--> Renames 'Approved' state to 'ApprovedNew' and updates all cards using the state.

State removal: 
`cyberismo update cisms/workflows/page remove states "Approved"`
--> Removes 'Approved' state. Cards using the state are not updated, as there is no replacement value given. It is assumed that the user knows what he/she is doing. User needs to update cards him/herself with valid value(s).

or with replacement value: 
`cyberismo update cisms/workflows/page remove states "Approved" "{\"name\": \"Deprecated\", \"category\": \"closed\"}"`
--> Removes 'Approved' state and replaces workflow state value in cards from 'Approved' to 'Deprecated'

Using the escaped JSON is a bit clumsy, but CLI use case is not very important. 
Programmatically the last operation is for example:
```
      const op = {
        name: 'change',
        target: { name: 'Approved', category: 'closed' },
        to: { name: 'Deprecated', category: 'closed' },
      } as RemoveOperation<WorkflowState>;
      workflow.update('states', op);
```

### DIFF
--- a/tools/data-handler/src/resources/file-resource.ts
+++ b/tools/data-handler/src/resources/file-resource.ts
@@ -19,6 +19,7 @@ import { readFile } from 'node:fs/promises';
 
 import type {
   Card,
+  FetchCardDetails,
   ResourceFolderType,
 } from '../interfaces/project-interfaces.js';
 import {
@@ -48,6 +49,7 @@ import {
 } from '../utils/resource-utils.js';
 import type { Resource } from '../interfaces/project-interfaces.js';
 import { sortCards } from '../utils/card-utils.js';
+import { Template } from '../containers/template.js';
 import { Validate } from '../commands/index.js';
 
 export {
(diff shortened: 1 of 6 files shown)


tools/data-handler/src/resources/file-resource.ts
```diff
@@ -102,6 +103,34 @@ export class FileResource extends ResourceObject {
     );
   }
 
+  // Collects cards that are using the 'cardTypeName'.
+  protected async collectCards(cardContent: object, cardTypeName: string) {
```
samutoljamo-pareto, line 107: Could we use fetchCardDetails rather tha object

tools/data-handler/src/resources/workflow-resource.ts
```diff
@@ -56,6 +84,94 @@ export class WorkflowResource extends FileResource {
     await this.write();
   }
 
+  // Handle change of workflow state.
+  private async handleStateChange<Type>(op: ChangeOperation<Type>) {
```
samutoljamo-pareto, line 88: Could most likely get rid of "as" keywords if `op` would be `ChangeOperation<Workflow>`

tools/data-handler/src/resources/workflow-resource.ts
```diff
@@ -56,6 +84,94 @@ export class WorkflowResource extends FileResource {
...
+    }
+    // Update all cards that use this state.
+    const toStateName = (op.to as MandatoryValue).name;
+
+    await this.updateCardStates(stateName, toStateName);
+  }
+
+  // Handle removal of workflow state.
+  // State can be removed with or without replacement.
+  private async handleStateRemoval<Type>(op: RemoveOperation<Type>) {
```
samutoljamo-pareto, line 120: Same here

samutoljamo-pareto (approved):

PR #686, opened 2025-05-28
INTDEV-860 Switch to simple-git when cloning modules

Fuzzbender:
This is drop-in replacement for `isomorphic-git`. 

No additional new features were intended to be implemented, but looks like SSH already works (importing needs a slight adjustment).

Should we somehow ensure that git client is installed? Maybe.

### DIFF
--- a/tools/data-handler/src/module-manager.ts
+++ b/tools/data-handler/src/module-manager.ts
@@ -11,16 +11,15 @@
   License along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
 
-import fs from 'node:fs';
 import { join } from 'node:path';
 import { mkdir, readdir, rm } from 'node:fs/promises';
 
-import git from 'isomorphic-git';
-import http from 'isomorphic-git/http/node/index.js';
+import { simpleGit, type SimpleGit } from 'simple-git';
 
 import { copyDir, deleteDir, pathExists } from './utils/file-utils.js';
 import type { Import } from './commands/index.js';
 import type {
+  Credentials,
   ModuleSetting,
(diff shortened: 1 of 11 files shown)


tools/data-handler/src/module-manager.ts
```diff
@@ -80,13 +65,9 @@ export class ModuleManager {
...
 
-    let repoUrl: URL;
-    try {
-      repoUrl = new URL(module.location);
-    } catch {
-      throw new Error(`Invalid repository URL: ${module.location}`);
-    }
+    const destinationPath = join(this.tempModulesDir, module.name);
 
+    let remote: string = module.location;
```
samutoljamo-pareto, line 70: I think `:string` here is not needed

tools/data-handler/src/module-manager.ts
```diff
@@ -97,37 +78,49 @@ export class ModuleManager {
           `... Using credentials '${process.env.CYBERISMO_GIT_USER}' for cloning '${module.name}'`,
         );
       }
-      repoUrl.username = process.env.CYBERISMO_GIT_USER;
-      repoUrl.password = process.env.CYBERISMO_GIT_TOKEN;
+      try {
+        const repoUrl = new URL(module.location);
+        const user = process.env.CYBERISMO_GIT_USER;
```
samutoljamo-pareto, line 83: Instead of putting these here, could we pass them around and take them from process.env inside the cli package? Module importing may be possible from app at some point and we cannot use env variables then

samutoljamo-pareto (approved):
I'll remove isomorphic git when antora is removed.

PR #687, opened 2025-05-28
INTDEV-871 remove antora

ghost:


samutoljamo-pareto:
Waiting until #683 is merged

samutoljamo-pareto:
Now the site export works just fine and UI also removes buttons etc that are cannot be used. Even if there would be something left that tries to do a post request, UI will simply show a notification: "only GET requests supported in export mode". This is done in a function that is used by all hooks, which contact the api.

To ease up development, there now is a `pnpm dev-export` script defined at top level. It will export whatever project is specified in `tools/backend/.env` and then use a development version of the app in export mode, which will use those generated jsons.

After the optimization suggested by @henryhaverinen , the bottleneck is that we have to evaluate macros of all pages. This is why generating `cyberismo-docs` takes less than a second whike `cyberismo-demo` takes ~15 seconds. 

There is still one issue though: it seems like with a large amount of cards(tried with 9k), memory usage climbs very fast until at some point the process exits because of an allocation error. This is very weird, because we only generate the cards once and we do not even call clingo. I'll try to see if I can find it. We only call: `showCardDetails`, `asciidoctor.convert` and `evaluatemacros` repeatedly. I took a headsnapshot and it looked like we had 3 million promises, not sure if that will lead to the issue.

tools/app/src/routes.ts
```diff
@@ -39,6 +65,7 @@ export const router = createBrowserRouter([
       },
       {
         path: '/cards/:key/edit',
+        loader: ({ params }) => createEditLoader(params.key!)(),
```
Fuzzbender, line 68: Can we cache this value to avoid reading the config file for each routing? 
The value is unlikely to change runtime.

tools/backend/src/app.ts
```diff
@@ -0,0 +1,57 @@
...
+      return c.text('Not Found', 400);
+    }
+    const file = await readFile(
+      path.join(import.meta.dirname, 'public', 'index.html'),
+    );
+    return c.html(file.toString());
+  });
+  // Error handling
+  app.onError((err, c) => {
+    console.error(err.stack);
```
Fuzzbender, line 52: Could we have logs file for backend as well? 
Any errors we get are now likely to get lost for static sites.

tools/backend/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import path from 'node:path';
+
+import fs, { readFile } from 'node:fs/promises';
```
Fuzzbender, line 16: If the whole `fs` needs to be imported, there is hardly need to separately import functions (note that there are more functions from `fs/promises` on line 20).

tools/backend/src/export.ts
```diff
@@ -0,0 +1,149 @@
...
+  projectPath: string,
+  dir?: string,
+  level?: 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'fatal',
+  onProgress?: (current?: number, total?: number) => void,
+) {
+  reset(); // Just in case
+  try {
+    dir = dir || 'static';
+
+    // replace with logger
```
Fuzzbender, line 108: Not sure what the comment refers to...

tools/backend/src/export.ts
```diff
@@ -0,0 +1,149 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2024
```
Fuzzbender, line 3: If this is new file, then year should be 2025

tools/backend/src/index.ts
```diff
@@ -1,66 +1,50 @@
+/**
```
Fuzzbender, line 1: Use the comment block version that won't spill out to 80+ chars (as you did above)

tools/backend/src/main.ts
```diff
@@ -1,7 +1,23 @@
+/**
```
Fuzzbender, line 1: Use the narrower copyright block (that won't have more than 80 chars per row)

tools/backend/src/utils.ts
```diff
@@ -0,0 +1,42 @@
+import path from 'node:path';
```
Fuzzbender, line 1: Add copyright block (for year 2025).

tools/backend/src/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+import { createServer } from 'node:net';
+
+export const staticFrontendDirRelative = path.relative(
+  process.cwd(),
+  path.resolve(import.meta.dirname, 'public'),
+);
+
+export async function findFreePort(
```
Fuzzbender, line 9: Would it make more sense to have port range `[min, max]` and the function would return any one value from the range. Now, it sort of works in that way, but it is `[min, attempts]`, where attempts is basically `max = min + attempts`.

tools/data-handler/src/commands/show.ts
```diff
@@ -256,8 +257,10 @@ export class Show {
    * Shows all cards (either template or project cards) from a project.
    * @returns cards list array
```
Fuzzbender, line 258: Add @param description

tools/backend/src/export.ts
samutoljamo-pareto: I think the toSsg required only a few methods so we could've gotten rid of fs, but it was removed in the new version

tools/backend/src/routes/cards.ts
```diff
@@ -70,6 +72,7 @@ router.get('/', async (c) => {
 async function getCardDetails(
   commands: CommandManager,
   key: string,
+  exportMode?: boolean,
```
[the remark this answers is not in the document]
samutoljamo-pareto: What about simply `static`?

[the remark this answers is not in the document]
samutoljamo-pareto: I used `staticMode`, works nicely with `mode: staticMode ? 'static' : 'inject',`

tools/cli/package.json
```diff
@@ -23,15 +23,19 @@
...
   "dependencies": {
-    "@inquirer/confirm": "^5.1.12",
-    "commander": "^14.0.0",
     "@cyberismo/backend": "workspace:*",
     "@cyberismo/data-handler": "workspace:*",
-    "dotenv": "^16.5.0"
+    "dotenv": "^16.5.0",
+    "@inquirer/confirm": "^5.1.12",
+    "cli-progress": "^3.12.0",
+    "commander": "^14.0.0"
```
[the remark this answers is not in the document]
samutoljamo-pareto: I think rebase messed it up, I'll reorder it

tools/backend/src/app.ts
Fuzzbender: Ok. Make a ticket about it and lets fix it at some point.
Having multiple log files might be useful, if the logs grow large and we know that only certain operations are interesting. Getting GB sized logs from customers can be difficult. Alternatively, log rotation should be considered.

samutoljamo-pareto: We don't need multiple files to filter the logs: e.g.
```
jq -c 'select(.queryName=="card")' .logs/cyberismo_data-handler.log | pnpm dlx pino-pretty
[13:06:47.633] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
[13:06:48.095] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
[13:06:48.525] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
[13:06:49.203] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
[13:06:49.613] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
[13:06:50.014] TRACE (1021848): Running query
    module: "calculate"
    queryName: "card"
```
I think rolling logs would be good, since rarely we have the need to look at logs over a long periods of time.

PR #688, opened 2025-05-30
Hotfix: run container on all pushes to main

ghost:
Forgot that the container build does not depend on other actions anymore. Prebuilds are not run if no changes happen to node clingo so container build did not run. I will rebase this on top of main, so only the `hotfix`-commit is added


samutoljamo-pareto:
Actually, it looks like the container was published. But the action is not visible.

PR #710, opened 2025-06-04
INTDEV-859 Move logs under project

samutoljamo-pareto:
Moved logs under the project to `<project_path>/.logs/dh.log`
By default, DH won't log to stdout, but the logs can be enabled with `-L debug|info|trace etc...`. Regardless of configuration, all logs(including trace) are logged to the file. The logs are in pino's json format, but they can be viewed like so: `cyberismo app -L trace | npx pino-pretty`

Also made a few small improvements to logging, but nothing major. Now we should start simply using the logger more. I removed the countless clingo traces, as there is no benefit to logging every single query / response given by clingo if there are no errors.

### DIFF
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -12,7 +12,7 @@
   License along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { Argument, Command } from 'commander';
+import { Argument, Command, Option } from 'commander';
 import confirm from '@inquirer/confirm';
 import {
   type CardsOptions,
@@ -172,7 +172,12 @@ const additionalHelpForRemove = `Sub-command help:
 program
   .name('cyberismo')
   .description(packageDef.description)
-  .version(packageDef.version);
+  .version(packageDef.version)
+  .addOption(
+    new Option('-L, --log-level <level>', 'Set the log level')
+      .choices(['trace', 'debug', 'info', 'warn', 'error', 'fatal'])
+      .default('fatal'),
+  );
 
 // Add card to a template
 program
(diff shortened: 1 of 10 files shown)


tools/cli/src/index.ts
```diff
@@ -172,7 +172,12 @@ const additionalHelpForRemove = `Sub-command help:
 program
   .name('cyberismo')
   .description(packageDef.description)
-  .version(packageDef.version);
+  .version(packageDef.version)
+  .addOption(
+    new Option('-L, --log-level <level>', 'Set the log level')
+      .choices(['trace', 'debug', 'info', 'warn', 'error', 'fatal'])
+      .default('fatal'),
```
Fuzzbender, line 179: Would 'error' be better default?

tools/data-handler/src/containers/project/project-paths.ts
```diff
@@ -81,7 +81,7 @@ export class ProjectPaths {
   }
 
   public get logPath(): string {
-    return join(this.resourcesFolder, 'dh.log');
+    return join(this.path, '.logs', 'dh.log');
```
Fuzzbender, line 84: Let's call the log file `cyberismo_data-handler.log`. In the future if we add separate log files, then call them `cyberismo_<tools subfolder name>.log`.

tools/cli/src/index.ts
samutoljamo-pareto: Idea here was that these are more like developer logs and in addition to that we have cli stdout outputs. CLI logs the errors it catches anyway. If the error is recoverable, we might wanna log it as error although we might not want to output it by default. Especially because the logs are in a structured format, they aren't very readable unless passed through the pino-logger

Fuzzbender: Ok. In the long run, the "CLI errors" need to be put in the log as well, since app users won't probably follow stdout:s while using the app.

PR #715, opened 2025-06-06
Improve unit tests

Fuzzbender:
When `Show` was changed to have two parameters for class construction, tests were not adjusted. Adding the missing `Calculate` parameter to all cases where `Show` is created.

Also, increase timeout for two git-related tests as these failed for me locally at least once.


samutoljamo-pareto (approved):
Macro tests(`test/macros/index.test.ts`) are also missing the Calculate class, could be fixed in this PR too

PR #719, opened 2025-06-11
Fix timeout issue on CI/Windows

Fuzzbender:
Since collecting prefixes uses "git clone", it can take a little time. If this timeouts (or any other exception), the `finally` part  is never called.  And thus the implementation jams.

Fixed, by moving the `collectModulePrefixes` under `try-catch`.

Also increasing one test's timeout to 15s, as it seem to take a lot of time occasionally for CI/Windows.

### DIFF
--- a/tools/data-handler/src/module-manager.ts
+++ b/tools/data-handler/src/module-manager.ts
@@ -365,7 +365,6 @@ export class ModuleManager {
 
     let uniqueModules: ModuleSetting[] = [];
     try {
-      // Collect prefixes from project's dependency modules.
       await this.collectModulePrefixes(modules, credentials);
 
       uniqueModules = this.removeDuplicates(this.modules);


Fuzzbender:
As I got lucky with CI build: Now it can be seen that timeout leads to normal CI test failure: https://github.com/CyberismoCom/cyberismo/actions/runs/15577526414
and no longer jams the action.

tools/data-handler/src/module-manager.ts
```diff
@@ -363,11 +363,11 @@ export class ModuleManager {
 
     const dotInterval = start();
 
-    // Collect prefixes from project's dependency modules.
-    await this.collectModulePrefixes(modules, credentials);
-
     let uniqueModules: ModuleSetting[] = [];
     try {
+      // Collect prefixes from project's dependency modules.
```
samutoljamo-pareto, line 368: this one

Fuzzbender: I'll remove it; it was there for some reason in the original code.

PR #722, opened 2025-06-11
INTDEV-872 INTDEV-880 INTDEV-858 Improve error handling in macros and node-clingo

ghost:
Ended up doing these tickets in the same PR, since they were very close to each other.

Now we're able to get errors from clingo as well as the warnings. To report/graph macro I added handling of the Clingo error so that it can also add its context to the error.

For macros I ended up adding an error field so that error handling can be done more simply. If a dependency throws an error, then that error is shown. Some examples for the errors: 

(report cisms/reports/dashboard2 uses invalid clingo syntax)
image
image

(report cisms/reports/dashboard2 uses valid syntax, but there are 2 reports so the the dashboard is given as input to the macro --> invalid json)
image

### DIFF
--- a/tools/data-handler/src/macros/base-macro.ts
+++ b/tools/data-handler/src/macros/base-macro.ts
@@ -91,6 +91,16 @@ abstract class BaseMacro {
     };
   }
 
+  private findTask(globalId: string, localId: number) {
+    const task = this.tasks.find(globalId, localId);
+    if (!task) {
+      this.logger.warn(
+        `Task not found for global id ${globalId}, local id ${localId}.`,
+      );
+    }
+    return task;
+  }
+
   /**
    * Function responsible for starting the promise and storing it along with its localId.
    */
@@ -125,7 +135,7 @@ abstract class BaseMacro {
       .then(() => {
         for (const dependency of dependencies) {
           if (dependency.error) {
-            const task = this.tasks.find(this.globalId, localId);
+            const task = this.findTask(this.globalId, localId);
             if (task) {
               // There could be a better way, but multi-nested macros are rare
               task.error = new MacroError(
@@ -138,12 +148,7 @@ abstract class BaseMacro {
                   parameters: dependency.parameters,
                 },
               );
-              return;
             }
-            // Should never happen
-            this.logger.warn(
-              `Task was not found for dependency: ${dependency.placeholder} (globalId: ${this.globalId}, localId: ${localId})`,
-            );
             return;
           }
           input = input.replace(
@@ -154,7 +159,7 @@ abstract class BaseMacro {
           try {
             JSON.parse(input);
           } catch {
-            const task = this.tasks.find(this.globalId, localId);
+            const task = this.findTask(this.globalId, localId);
             if (task) {
               task.error = new MacroError(
                 'Invalid JSON produced by macro dependency',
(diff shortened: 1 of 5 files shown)


tools/data-handler/src/macros/base-macro.ts
```diff
@@ -111,20 +121,62 @@ abstract class BaseMacro {
     const dependencies = this.findDependencies(rawInput);
 
     // Create a promise to resolve dependencies, execute the macro, and handle the results
-    const promise = Promise.all(dependencies.map((dep) => dep.promise))
+    const promise = Promise.allSettled(dependencies.map((dep) => dep.promise))
       .then(() => {
         for (const dependency of dependencies) {
+          if (dependency.error) {
+            const task = this.tasks.find(this.globalId, localId);
```
Fuzzbender, line 128: There is some code repetition 128-147 & 157-174.
See if you could combine these to a `private` method

tools/data-handler/src/macros/base-macro.ts
```diff
@@ -134,27 +186,45 @@ abstract class BaseMacro {
...
-          console.error(
+          this.logger.warn(
             `Task not found after execution: macro ${this.metadata.name}, local id ${localId}.`,
           );
         }
       })
       .catch((err) => {
+        //
+        if (!(err instanceof Error)) {
+          err = new Error('Unknown error');
```
Fuzzbender, line 205: You could log the full error here. Just in case we get to some unknown error, it is easier to investigate if the original error is not totally lost.

tools/data-handler/src/macros/base-macro.ts
```diff
@@ -111,20 +121,62 @@ abstract class BaseMacro {
...
+                dependency.error.context.parameters,
+                {
+                  macroName: dependency.macro,
+                  parameters: dependency.parameters,
+                },
+              );
+              return;
+            }
+            // Should never happen
+            this.logger.warn(
```
Fuzzbender, line 144: Shouldn't all of the macro errors be logged as errors and not warnings?
It usually should indicate that there is something wrong with your project configuration if you receive macro errors.

tools/node-clingo/src/binding.cc
```diff
@@ -52,7 +61,31 @@ void handle_clingo_error(const Napi::Env &env)
...
     if (clingo_error_code() != 0)
     {
-        throw Napi::Error::New(env, clingo_error_message());
+        Napi::Error error = Napi::Error::New(env, clingo_error_message());
+        
+        // Create an object to hold error details
+        Napi::Object errorObj = Napi::Object::New(env);
+        
+        // Add errors and warnings from g_errorMessages
+        Napi::Array errors = Napi::Array::New(env);
```
Fuzzbender, line 70: Rows 70-88 &  496-511 repeat the same code; make a routine that is used in both to avoid code repetition.

tools/node-clingo/src/binding.cc
```diff
@@ -38,7 +47,7 @@ std::unordered_map<std::string, std::string> g_basePrograms;
  */
 void silent_logger(clingo_warning_t code, char const *message, void *data)
```
samutoljamo-pareto, line 48: I'll rename this and change the comment, since it's not a silent logger anymore

PR #723, opened 2025-06-11
Release

samutoljamo-pareto:
This should trigger CI to create a new PR once this is merged to main.

### DIFF
--- a/.changeset/eleven-knives-cough.md
+++ b/.changeset/eleven-knives-cough.md
@@ -1,0 +1,9 @@
+---
+'@cyberismo/node-clingo': minor
+'@cyberismo/assets': minor
+'@cyberismo/backend': minor
+'@cyberismo/cli': minor
+'@cyberismo/data-handler': minor
+---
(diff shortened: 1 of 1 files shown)


.changeset/eleven-knives-cough.md
Fuzzbender, line 1: Is the filename hard coded? The name looks a bit "special". Can we change the name?

Fuzzbender (approved):

Fuzzbender:
> The CI would not really need to run when we make these "publish PRs"

Agreed. This might be a solution: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs

.changeset/eleven-knives-cough.md
samutoljamo-pareto: It's randomized. I think the file name does not matter. I guess that is because it allows large teams to concurrently do releases. In other words, not much value to us

PR #725, opened 2025-06-12
Remove changesets, simplify

samutoljamo-pareto:
After using changesets a bit, I realized we do not really need it and we ran into some problems, so let's simplify:
1. Release if version has not been released in main(`pnpm publish -r` cannot "override" a previous release)
2. When you want to publish(everything else except `node-clingo`), just run `pnpm bump-version patch|minor|major` in root of the repo, commit the change and create a PR
3. For `node-clingo`, just manually increase the version


Fuzzbender (approved):
Looks good to me.

PR #729, opened 2025-06-12
Skip git module cloning test on CI/Windows due to fluctuation

Fuzzbender:
Let's turn the test off on CI/Windows since it seems to be rather troublesome.


samutoljamo-pareto (approved):
This is fine, but I think there is also a skip function, which should exist in both `it` and `this`. So I think instead of returning you could call this.skip() or create this type of function `const itIfNotWindows = process.platform === "win32" ? it.skip : it` and use that instead of `it`

PR #730, opened 2025-06-12
INTDEV-875 Misleading error message when configuration file is not valid

Fuzzbender:
Use the error message from the exception instead of trying to come up with a good one.


tools/data-handler/src/project-settings.ts
```diff
@@ -60,15 +61,11 @@ export class ProjectConfiguration implements ProjectSettings {
     let settings;
     try {
       settings = readJsonFileSync(this.settingPath) as ProjectConfiguration;
-    } catch {
-      throw new Error(
-        `Invalid path '${this.settingPath}' to configuration file`,
-      );
+    } catch (error) {
+      throw new Error(errorFunction(error));
```
samutoljamo-pareto, line 65: Is the error function really needed? I think it's generally bad to wrap errors to a more simplified version unless there's a use-case(e.g. providing more context). Then you can call that function at higher level(currently handled in commandHandler/backend). Thus we would not really need the try catch.

samutoljamo-pareto (approved):
Left a minor comment

PR #734, opened 2025-06-17
INTDEV-881 Simplify query language

samutoljamo-pareto:


tools/app/src/components/ContentArea.tsx
```diff
@@ -410,8 +410,8 @@ const PolicyChecks = ({
   const router = useAppRouter();
 
   if (
-    policyChecks.successes.length === 0 &&
-    policyChecks.failures.length === 0
+    (policyChecks.successes?.length ?? 0) === 0 &&
```
Fuzzbender, line 413: I think making the return arrays optional looks a bit ugly on app. Now it needs to check each value and use empty array if something is missing. How about that DH/backend would put all arrays in place (even If they are empty) for app, so that app could count on that arrays are there, but they can be empty

henryhaverinen: We could add a new predicate for defining that there should be a certain child result collection. It could be called `childResultCollection(ParentKey, Collection)` as we're not usually using verbs in "declarative" predicates and we're using the term collection already.

tools/data-handler/src/permissions/action-guard.ts
```diff
@@ -50,27 +50,27 @@ export class ActionGuard {
     }
     const res = cards[0];
     if (action === 'editContent') {
-      return checkOperation(res.deniedOperations.editContent);
+      return checkOperation(res.deniedOperations.editContent ?? []);
```
samutoljamo-pareto, line 53: These could still be deleted

henryhaverinen:
I noticed that in the cyberismo-demo project card demo_6xk31ver, the report about policy checks does not show any results with this change, but it showed some results previously. I think this is expected, as the logic to include policyChecks in the result objects has now been moved from the query language to the card query. Unfortunately the query rules to include policy checks in card.lp are not super simple to replicate in reports.

So this is a content breaking change, and we have to try and find reports that make assumptions about policyChecks, links or other removed features of the result objects that you used to get with a `selectAll`.

PR #735, opened 2025-06-17
INTDEV-929 Fix publish

ghost:
**Infinite loop on linux**: For some reason, putting on argument at the end of shebang does not work on linux. E.g. https://github.com/TryGhost/gscan/issues/307  https://github.com/nodejs/node/issues/32876 . We are already using `process.removeAllListeners('warning')`, which disables all warnings, so we should simply remove the argument and it should work on all platforms after.

**only-allow pnpm**: This caused some issues, because it also ran as it was specified as a pre-install script. We want to allow using `npm install -g @cyberismo/cli` or `npx @cyberismo/cli`, so let's remove `only-allow` from packages. Only-allow is still present at the root of the repo.

### DIFF
--- a/tools/node-clingo/test/solve.test.ts
+++ b/tools/node-clingo/test/solve.test.ts
@@ -10,7 +10,6 @@ describe('Clingo solver', () => {
   it('should solve a simple logic program', async () => {
     const program = 'a. b. c(1). c(2).';
     const result = await solve(program);
-    console.log(result);
 
     expect(result).toBeDefined();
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto:
Added 2 more fixes:
* --ignore-scripts was not used, resulting in an error in the publish CI if you're attempting to publish node-clingo, because it tries to build the node-clingo package without clingo. However, it is not needed to do the release to npm
* Removed some console logs from node clingo tests

PR #738, opened 2025-06-18
INTDEV-883 Configuration editor

samutoljamo-pareto:
Added configuration editor's navigation. Currently clicking does nothing. Everything also works in export site.
Note: Most of the changes are related to restructuring as I wanted both the /cards and /configuration to share code when possible.


henryhaverinen (changes requested):
This is looking very promising. I'd like to see the following changes:

- The navigation hierarchy should use the display name, whenever it is available. For example, "Annual task" instead of "base/templates/annualTask"
- If the display name is not available, then navigation should show the last part of the technical name (identifier), for example "evidence.lp" instead of eucra/calculations/evidence.lp, and "annualTask" instead of base/cardTypes/annualTask (if display name is not available for this card type).
- Calculations seems to be missing for the modules. Calculations is shown for the local project
- The title of the navigation tree could be for example Configuration - {project name}

henryhaverinen:
For some reason, .DS_Store is shown in the hierarchy:

image

henryhaverinen:
This implementation shows the top level resources only. I think it is a good first step, but we don't have stories in the backlog for the more detailed navigation. So you could consider the following improvements next:

- Showing "Parameters schema"​, "Query", and "Report template" under each report
- Showing "Model logic program" under each graph model
- Showing "View logic program" under each graph view
- Showing the template cards hierarchy under each template, ordered by rank

samutoljamo-pareto:
> For some reason, .DS_Store is shown in the hierarchy:
> 
> image

Could create a seperate ticket, as it is a bug in DH

Fuzzbender:
> For some reason, .DS_Store is shown in the hierarchy:
> 
> image

This is DH's and module's problem. 
1) `.DS_Store` shouldn't be pushed to the project
2) validation should complain about it
3) it should definitely NOT be shown as a resource

samutoljamo-pareto:
Ready to merge, I'll ask Sami on monday if he wants to review this or could we put this to main directly. This change is so big that it would be better to get to main rather than put it to temp-main

PR #739, opened 2025-06-19
INTDEV-895 adjust escaping

samutoljamo-pareto:
While looking at INTDEV-895, noticed that a couple of extra characters were being escaped. The actual bug was in the project itself. I also added a few tests


Fuzzbender:
Looks good to me.

Fuzzbender (approved):

PR #742, opened 2025-06-19
INTDEV-934 Only accept JSON and LP files as potential resource files

Fuzzbender:
Instead of deny-listing all sorts of file names, only accept JSON and LP files as potential resource files.

### DIFF
--- a/tools/data-handler/src/containers/project/resource-collector.ts
+++ b/tools/data-handler/src/containers/project/resource-collector.ts
@@ -14,9 +14,8 @@
 
 import { type Dirent, readdirSync } from 'node:fs';
 import { readdir } from 'node:fs/promises';
-import { join } from 'node:path';
+import { extname, join } from 'node:path';
 
-import { CardContainer } from '../card-container.js';
 import type { Project } from '../project.js';
 import type { ProjectPaths } from './project-paths.js';
 import type {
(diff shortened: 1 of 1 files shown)


henryhaverinen (approved):
I don't suppose there's any need to list .adoc and .hbs as allowed extensions?

Fuzzbender:
> I don't suppose there's any need to list .adoc and .hbs as allowed extensions?

No this is only when "collecting" files; the resources themselves then check their files. 
If there is any file that matches the rule, in an appropriate folder makes that file ~ a resource. 
Thus, only single file is needed. Generally, resources all have JSON file, but calculations are ~resources, therefore LP file is also accepted.

tools/data-handler/src/containers/project/resource-collector.ts
```diff
@@ -232,10 +232,9 @@ export class ResourceCollector {
     resources.push(
       ...entries
         .filter((entry) => {
+          const extension = extname(entry.name);
           return (
-            entry.isFile() &&
-            entry.name !== '.gitkeep' &&
-            entry.name !== CardContainer.schemaContentFile
+            (entry.isFile() && extension === '.json') || extension === '.lp'
```
samutoljamo-pareto, line 237: I think the parenthesis should be in the OR statement. `entry.isFile() && (extension === '.json' || extension === '.lp')`. Alternatively we could define a named constant somewhere(and use `whilelist.includes(extension)`) so it is more easy to change it.

samutoljamo-pareto (approved):

PR #760, opened 2025-07-07
Bugfix/henryhaverinen/intdev 951

henryhaverinen:


samutoljamo:
Noticed that too, we need to define the fields manually as we can't select them directly anymore since we use Card,Field as the key rather than Field

samutoljamo-pareto (approved):
Looks great

PR #763, opened 2025-07-18
INTDEV-949 Fix: raw macro does not work for handlebars helpers / syntax

samutoljamo-pareto:
Only macros were being escaped. Instead of trying to implement raw macro within handlebars, I used a regex to process the content before passing it forward. Thus, anything inside raw macros will be rendered as is

### DIFF
--- a/tools/data-handler/src/macros/index.ts
+++ b/tools/data-handler/src/macros/index.ts
@@ -34,6 +34,24 @@ const CURLY_LEFT = '&#123;';
 const CURLY_RIGHT = '&#125;';
 
 /**
+ * Pre-processes the content to handle {{#raw}} blocks by escaping all handlebars syntax inside them
+ * @param content The template content to process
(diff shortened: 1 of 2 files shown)


tools/data-handler/src/macros/index.ts
```diff
@@ -33,6 +33,80 @@ import { ClingoError } from '@cyberismo/node-clingo';
...
+    }
+    return lineNum;
+  };
+
+  while (i < content.length) {
+    // Check for {{#raw}}
+    if (matchesAt(i, '{{#raw}}')) {
+      const openingLine = getLineNumber(i);
+      // Find the matching {{/raw}} - no nesting allowed
+      let j = i + 8;
```
Fuzzbender, line 71: I assume that the magic 8 is the length of `{{#raw}}`.
To make the code more easier to read, make string constants from re-used strings, and use their lengths instead of magic numbers.

PR #771, opened 2025-07-23
INTDEV-922 implement SVG score card generation

samutoljamo-pareto:
Score cards were using html syntax and as part of wanting to use `asciidoctor-pdf`, we cannot use html based components or we need have 2 separate implementations. In this case, I used svg for implementing the score card. Some restrictions:
* Rendering now happens in DH instead of App. Generally this is a bad idea. Might be smart to create some kind of common package that shared between DH and App, which would allow the app to render it. Development would be nicer if we used e.g. react to make the template instead of actually string(it can still be rendered on the server)
* Layout depends on estimating size of text(we could get a more accurate estimation by actually rendering the svg and checking its size, but I don't think that's needed)
* The svg is not responsive, but app does not work on mobile anyway
* No good way to go around these issues, unless we either:
1. Utilize web technologies for the pdf and use html(no straightforward tool exists, since we did not want to use `asciidoctor-web-pdf`, since it has stopped  receiving updates)
2. Use svg for exports but html for the site / exported site: results in small differences in the pdf version.

I think we can for now go with the svg version and use a separate implementation for app if needed later. 

image

### DIFF
--- a/tools/data-handler/src/svg/scoreCard.ts
+++ b/tools/data-handler/src/svg/scoreCard.ts
@@ -78,9 +78,9 @@ export function scoreCard(options: ScoreCardOptions): string {
   const svgContent = `<svg class="card" width="${width}" height="${height}" viewBox="0 0 ${width} ${height}" xmlns="http://www.w3.org/2000/svg">
     <rect rx="8" ry="8" fill="#fff" stroke="#dfe4ea" stroke-width="3" width="${width}" height="${height}"/>
     <g text-anchor="middle">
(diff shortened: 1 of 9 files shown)


tools/data-handler/src/svg/scoreCard.ts
```diff
@@ -0,0 +1,88 @@
...
+const LINE_GAP_CAPTION = 16;
+
+/**
+ * Options for the score card
+ * @param title - The title of the score card
+ * @param value - The value of the score card
+ * @param unit - The unit of the score card
+ * @param legend - The legend of the score card
+ */
+export interface ScoreCardOptions {
```
Fuzzbender, line 35: It would be nicer to have the non-optional parameters first in the interface.

PR #773, opened 2025-07-24
INTDEV-561 Pie chart / charting

samutoljamo-pareto:
I tried different charting libraries, but Vega seemed to be the best one for us. Vega charts are specified using json and they've also provided a json schema for that. Vega charts can be also rendered with svg without web technologies(For most charting libraries, this was hard). I also added Vega-Lite, which provides a shorter way to specify Vega charts in common scenarios(There is quite many of the: https://vega.github.io/vega-lite/examples/). Vega-Lite is "compiled" to Vega.

### Example Vega-Lite pie chart
```
{
  "$schema": "https://vega.github.io/schema/vega-lite/v6.json",
  "description": "A simple pie chart with embedded data.",
  "data": {
    "values": [
      {"category": 1, "value": 4},
      {"category": 2, "value": 6},
      {"category": 3, "value": 10},
      {"category": 4, "value": 3},
      {"category": 5, "value": 7},
      {"category": 6, "value": 8}
    ]
  },
  "mark": "arc",
  "encoding": {
    "theta": {"field": "value", "type": "quantitative"},
    "color": {"field": "category", "type": "nominal"}
  }
}
```

### Example Vega pie chart
```
{
  "$schema": "https://vega.github.io/schema/vega/v6.json",
  "description": "A basic pie chart example.",
  "width": 200,
  "height": 200,
  "autosize": "none",

  "signals": [
    {
      "name": "startAngle", "value": 0,
      "bind": {"input": "range", "min": 0, "max": 6.29, "step": 0.01}
    },
    {
      "name": "endAngle", "value": 6.29,
      "bind": {"input": "range", "min": 0, "max": 6.29, "step": 0.01}
    },
    {
      "name": "padAngle", "value": 0,
      "bind": {"input": "range", "min": 0, "max": 0.1}
    },
    {
      "name": "innerRadius", "value": 0,
      "bind": {"input": "range", "min": 0, "max": 90, "step": 1}
    },
    {
      "name": "cornerRadius", "value": 0,
      "bind": {"input": "range", "min": 0, "max": 10, "step": 0.5}
    },
    {
      "name": "sort", "value": false,
      "bind": {"input": "checkbox"}
    }
  ],

  "data": [
    {
      "name": "table",
      "values": [
        {"id": 1, "field": 4},
        {"id": 2, "field": 6},
        {"id": 3, "field": 10},
        {"id": 4, "field": 3},
        {"id": 5, "field": 7},
        {"id": 6, "field": 8}
      ],
      "transform": [
        {
          "type": "pie",
          "field": "field",
          "startAngle": {"signal": "startAngle"},
          "endAngle": {"signal": "endAngle"},
          "sort": {"signal": "sort"}
        }
      ]
    }
  ],

  "scales": [
    {
      "name": "color",
      "type": "ordinal",
      "domain": {"data": "table", "field": "id"},
      "range": {"scheme": "category20"}
    }
  ],

  "marks": [
    {
      "type": "arc",
      "from": {"data": "table"},
      "encode": {
        "enter": {
          "fill": {"scale": "color", "field": "id"},
          "x": {"signal": "width / 2"},
          "y": {"signal": "height / 2"}
        },
        "update": {
          "startAngle": {"field": "startAngle"},
          "endAngle": {"field": "endAngle"},
          "padAngle": {"signal": "padAngle"},
          "innerRadius": {"signal": "innerRadius"},
          "outerRadius": {"signal": "width / 2"},
          "cornerRadius": {"signal": "cornerRadius"}
        }
      }
    }
  ]
}
```
### Available macros
You can use vega directly:
```
{{#vega}}
"spec": {...}
{{/vega}}
```
Or specify charts using Vega-Lite
```
{{#vegaLite}}
"spec": {...}
{{/vegaLite}}
```
I did not yet create the pie chart macro yet. I think we could implement them as reports if we need to provide a higher level abstraction than Vega Lite is instead of hard coding them.

DH renders the charts for static export, but app renders the chart based on the spec using canvas. Although app does not use svg renderer, based on my tests they looked exactly the same. And if we want, we can enable svg renderer in the app too. With the same spec, we can also make charts interactive in app while making a static export for the pdf:
```
{{#vegaLite}}
  "spec": {
    "description": "Pie Chart with percentage_tooltip",
    "width": 100,
    "height": 100,
    "data": {
      "values": [
        {
          "category": 1,
          "value": 4
        },
        {
          "category": 2,
          "value": 6
        },
        {
          "category": 3,
          "value": 10
        },
        {
          "category": 4,
          "value": 3
        },
        {
          "category": 5,
          "value": 7
        },
        {
          "category": 6,
          "value": 8
        }
      ]
    },
    "mark": {
      "type": "arc"
    },
    "encoding": {
      "theta": {
        "field": "value",
        "type": "quantitative",
        "stack": "normalize"
      },
      "color": {
        "field": "category",
        "type": "nominal"
      },
      "tooltip": [
        {
          "field": "category",
          "type": "nominal"
        },
        {
          "field": "value",
          "type": "quantitative"
        }
      ]
    }
  }
{{/vegaLite}}
```


tools/data-handler/src/macros/index.ts
```diff
@@ -438,3 +406,18 @@ export function createImage(image: string, controls: boolean = true) {
...
+/**
+ * Creates a Handlebars macro block string with the given macro name and options.
+ *
+ * @param macro - The name of the macro to create (e.g., 'scoreCard', 'include').
+ * @param options - The options object to be stringified and inserted as macro content.
+ * @returns The Handlebars macro block as a string, e.g. {{#macro}}...{{/macro}}
+ */
+export function createMacro(macro: MacroName, options: unknown) {
+  let optionsString = JSON.stringify(options, null, 0);
+  if(optionsString.length > 1) {
```
Fuzzbender, line 419: Add whitespace before opening parentheses. 
We should change the linter so that it would complain about this.

Fuzzbender (approved):

tools/data-handler/src/macros/index.ts
samutoljamo-pareto: Yes and when that is done, linter should be also configured for backend. Currently it is not checked.

samutoljamo-pareto: Actually that was because I had prettier accidentally turned off. After running prettier that was fixed.

PR #774, opened 2025-07-25
INTDEV-747

samutoljamo-pareto:
I used link instead of xref so I changed link to xref

### DIFF
--- a/tools/data-handler/src/macros/xref/index.ts
+++ b/tools/data-handler/src/macros/xref/index.ts
@@ -50,7 +50,8 @@ export default class XrefMacro extends BaseMacro {
     }
 
     // Generate AsciiDoc link with proper React Router URL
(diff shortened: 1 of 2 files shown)


tools/data-handler/src/macros/xref/index.ts
```diff
@@ -50,7 +50,8 @@ export default class XrefMacro extends BaseMacro {
     }
 
     // Generate AsciiDoc link with proper React Router URL
-    return `link:/cards/${options.cardKey}[${card.metadata.title}]`;
+    return `xref:${options.cardKey}.adoc[${card.metadata.title}]`;
+    // xref:docs_qp3vhh4t.adoc[Link Title]
```
Fuzzbender, line 54: Is the comment line needed?

PR #780, opened 2025-07-29
INTDEV-780 Add resource prefix/type/identifier functions

samutoljamo-pareto:
Added `resourcePrefix`, `resourceType` and `resourceIdentifier` functions


samutoljamo-pareto:
I won't merge this to temp-main, because it would trigger node-clingo CIs for all future PRs

Fuzzbender (approved):

PR #782, opened 2025-07-31
INTDEV-924 Add context to logic programs

samutoljamo-pareto:
We add context by setting each logic program for each resource separately. To make adding logic program to solves easier, I created a system where:
* Each program has a key(e.g. `base/fieldTypes/myFieldType`
* Each program may have flags(e.g. `resources` or `main`)

When you want to solve something, you may use flags or keys to add logic programs. Currently we only have flag `main`, which adds every single logic program there is.

### Errors
For example, if there is an exception in some calculation:
image

### Exceptions
Our exception handling was quite complex as we had NAPI on no-exception mode while we had c++ exceptions enabled. After upgrading NAPI to a newer version, I started using NAPIs `except_all` target, which catches std exceptions too so we do not have to add the boilerplate try catch for all functions.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -348,8 +348,8 @@ importers:
   tools/node-clingo:
     dependencies:
       node-addon-api:
-        specifier: ^7.1.0
-        version: 7.1.1
+        specifier: ^8.5.0
+        version: 8.5.0
       node-gyp-build:
         specifier: ^4.8.4
         version: 4.8.4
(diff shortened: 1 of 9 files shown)


Fuzzbender:
Looks like there is some other commits in this PR as well. 
I'll review just the "INTDEV-924" commit and ones after it.

samutoljamo-pareto:
Fixed now

Fuzzbender:
Looks good to me. The only thing I am on-fence is the term `flag`. 

How about calling it a `category` instead? Flag is usually an instruction for program(s).
If you concur with this, then `main` should probably be `all`.

PR #792, opened 2025-08-04
INTDEV-887 INTDEV-888 add basic editors and viewers

samutoljamo-pareto:
I added very basic viewers for all pages and an editor for file based content. It's not yet possible to modify resources themselves.

On DH, I added the possibility of editing and showing files for folder based resources. This means that all the files under reports, graphviews and graphmodels are treated as just files.

image

image
image

image
image


tools/backend/src/domain/resources/service.ts
```diff
@@ -17,6 +17,10 @@ import {
   type ResourceFolderType,
 } from '@cyberismo/data-handler/interfaces/project-interfaces';
 import { CommandManager } from '@cyberismo/data-handler';
+import {
+  resourceName,
+  isResourceFolderType,
+} from '../../../../data-handler/dist/utils/resource-utils.js';
```
Fuzzbender, line 23: imports from `dist` folder, should probably import from sources directly. 
Alternatively, add these to common exports in DH and import them on row 19

tools/data-handler/src/resources/folder-resource.ts
```diff
@@ -98,6 +106,54 @@ export class FolderResource extends FileResource {
...
+    const filePath = join(this.internalFolder, fileName);
+    return readFile(filePath, 'utf8');
+  }
+
+  /**
+   * Shows all file names in the resource.
+   * @returns all file names in the resource.
+   */
+  public async showFileNames(): Promise<string[]> {
+    return readdir(this.internalFolder);
```
Fuzzbender, line 124: You could filter to show only allow-listed resource files (.lp, .hb, ...).

tools/data-handler/src/utils/resource-utils.ts
```diff
@@ -29,7 +29,11 @@ const TYPE_INDEX = 1;
 const IDENTIFIER_INDEX = 2;
 // Valid resource name has three parts
 const RESOURCE_NAME_PARTS = 3;
+const RESOURCE_FOLDER_TYPES = ['graphModels', 'graphViews', 'reports'];
```
Fuzzbender, line 32: Template is also a folder resource.

tools/data-handler/src/commands/edit.ts
```diff
@@ -83,10 +87,26 @@ export class Edit {
...
-    if (!cardPath) {
-      throw new Error(`Card '${cardKey}' does not exist in the project`);
+    const prefix = getModuleNameFromCardKey(cardKey);
+    if (prefix === this.project.projectPrefix) {
+      // Determine the card path
+      const cardPath = this.project.pathToCard(cardKey);
+      if (!cardPath) {
+        throw new Error(`Card '${cardKey}' does not exist in the project`);
+      }
+    } else {
```
Fuzzbender, line 97: If it is not a project card, you cannot edit it. It is not allowed to edit module cards.

tools/data-handler/src/containers/template.ts
```diff
@@ -583,4 +584,27 @@ export class Template extends CardContainer {
...
     return this.showCards(this.templateCardsPath);
   }
+
+  /**
+   * Updates card content.
+   * @param cardKey card's ID that is updated.
+   * @param changedContent changed content
+   */
+  public async updateCardContent(cardKey: string, changedContent: string) {
+    return super.updateCardContent(cardKey, changedContent);
```
Fuzzbender, line 594: If not in `Edit`, then here it should check that this is not a module template. 
If it is, then editing should fail. (Similarly for the other new function).

tools/data-handler/src/resources/folder-resource.ts
```diff
@@ -98,6 +106,54 @@ export class FolderResource extends FileResource {
...
+      throw new Error(`File '${fileName}' is not in the resource`);
+    }
+
+    // This makes sure that the file is in the resource folder.
+    if (basename(normalizedFilePath) !== fileName) {
+      throw new Error(`File '${fileName}' is not in the resource`);
+    }
+
+    // We need to check if a file exists because we don't want to allow creating a new file.
+    const fileExists = await access(filePath)
```
Fuzzbender, line 148: Don't. Instead just call `writeFile` with `flag: 'wx'`. 
Promise fails if file does not exist.

Fuzzbender (approved):
Looks good to me, some minor comments.

tools/data-handler/src/resources/folder-resource.ts
Fuzzbender: Having `access()` and some other file API call introduces a race condition.

You can just skip the `access()` call and use `writeFile` with `flag: 'r+'` if you want to modify existing file (sorry; I initially recommended using `wx` which would only for new files, not existing). 

From documentation:
'r+': Open file for reading and writing. **An exception occurs if the file does not exist.**

samutoljamo-pareto: Thanks, did not `r+` existed so I thought the race condition there was acceptable as it occurring would be extremely unlikely(happens only if user manually edits file system). I'll change it to `r+`

tools/data-handler/src/commands/edit.ts
```diff
@@ -83,10 +87,26 @@ export class Edit {
...
+      for (const template of await this.project.templates(ResourcesFrom.all)) {
```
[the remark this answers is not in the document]
samutoljamo-pareto: Yes

PR #794, opened 2025-08-05
INTDEV-873 Improve SVG viewer link functionality

CyberJML:
Improved way to use links. Try it out. This should now
- Not activate links if any dragging is detected
- Navigate to links if there's no dragging detected

### DIFF
--- a/tools/app/src/components/modals/svgViewerModal.tsx
+++ b/tools/app/src/components/modals/svgViewerModal.tsx
@@ -9,6 +9,7 @@ import {
 } from '@mui/material';
 import { Add, Remove, Autorenew, Close } from '@mui/icons-material';
 import { Link } from 'react-router';
+import { useTranslation } from 'react-i18next';
 
 export interface SvgViewerModalProps {
   open: boolean;
@@ -58,6 +59,7 @@ const SvgViewerModal: React.FC<SvgViewerModalProps> = ({
   padding = 32,
 }) => {
   useBackButtonModal(open, onClose);
+  const { t } = useTranslation();
 
   const [zoom, setZoom] = useState(1);
   const [naturalSize, setNaturalSize] = useState<Size | null>(null);
@@ -160,7 +162,7 @@ const SvgViewerModal: React.FC<SvgViewerModalProps> = ({
     let moved = false;
 
     const clickBlocker = (e: MouseEvent) => {
-      if (!e.ctrlKey && !e.metaKey) {
+      if (moved) {
         e.preventDefault();
         e.stopPropagation();
       }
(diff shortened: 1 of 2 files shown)


Fuzzbender (approved):
Looks good to me.

If this resolves / changes behaviour that is described in https://cyberismo.atlassian.net/browse/INTDEV-873, please modify the PR title to contain the JIRA ticket ID. Also amend the commit message with the JIRA ticket ID.

tools/app/src/components/modals/svgViewerModal.tsx
```diff
@@ -282,8 +261,7 @@ const SvgViewerModal: React.FC<SvgViewerModalProps> = ({
             userSelect: 'none',
           }}
         >
-          Pinch or Shift+Wheel = Zoom • Scroll/Drag = Pan • Ctrl/Cmd+Click =
-          Follow link
+          Pinch or Shift+Wheel = Zoom • Scroll/Drag = Pan
```
samutoljamo-pareto, line 264: Could use the `useTranslation` hook for this

samutoljamo-pareto (approved):
Looks good to me, but could @Fuzzbender test if the issue on mac went away?

Fuzzbender:
> Looks good to me, but could @Fuzzbender test if the issue on mac went away?

Seems to work in Safari. The SVG link works similarly in both "normal" and "fullscreen" mode. 
Clicking a link opens the link in the same tab. 
CMD + Link opens in new tab. 
Shift/Option + Link downloads the linked content. 
Right click opens context menu.

tools/app/src/components/modals/svgViewerModal.tsx
CyberJML: Makes sense, I will

CyberJML:
> > Looks good to me, but could @Fuzzbender test if the issue on mac went away?
> 
> Seems to work in Safari. The SVG link works similarly in both "normal" and "fullscreen" mode. Clicking a link opens the link in the same tab. CMD + Link opens in new tab. Shift/Option + Link downloads the linked content. Right click opens context menu.

One thing you should also test with MAC is to drag and end drag over the link. This was major problem at the start and the reason why I originally added the ctrl key to follow the link. Start drag anywhere, and end it over the link -> Link should not be followed

Fuzzbender:
> > > Looks good to me, but could @Fuzzbender test if the issue on mac went away?
> > 
> > 
> > Seems to work in Safari. The SVG link works similarly in both "normal" and "fullscreen" mode. Clicking a link opens the link in the same tab. CMD + Link opens in new tab. Shift/Option + Link downloads the linked content. Right click opens context menu.
> 
> One thing you should also test with MAC is to drag and end drag over the link. This was major problem at the start and the reason why I originally added the ctrl key to follow the link. Start drag anywhere, and end it over the link -> Link should not be followed

I was not able to reproduce any issues when dragging links (either starting from link and drag away; or drag ending on top of a link).

PR #796, opened 2025-08-06
Skip fluctuating module-manager test in Windows

Fuzzbender:
Seems that this test fails from time to time on Windows (maybe once every 20 runs or so). 
Let's skip it for now.

### DIFF
--- a/tools/data-handler/test/module-manager.test.ts
+++ b/tools/data-handler/test/module-manager.test.ts
@@ -13,6 +13,7 @@ import { fileURLToPath } from 'node:url';
 import { CommandManager } from '../src/command-manager.js';
 
 describe('module-manager', () => {
+  const skipTest = process.env.GITHUB_ACTIONS && os.platform() === 'win32';
   const baseDir = dirname(fileURLToPath(import.meta.url));
   const testDir = join(baseDir, 'tmp-module-manager-tests');
   const decisionRecordsPath = join(testDir, 'valid/decision-records');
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/module-manager.ts
```diff
@@ -231,6 +228,19 @@ export class ModuleManager {
       : new Set();
   }
 
+  // Increase timeout for CI environments and add platform-specific adjustments
+  private gitTimeout(): number {
+    const baseTimeout = 15000;
+    const isCI = process.env.GITHUB_ACTIONS;
```
samutoljamo-pareto, line 234: I think `process.env.CI` would be more accurate

tools/data-handler/test/module-manager.test.ts
```diff
@@ -88,19 +88,25 @@ describe('module-manager', () => {
...
     ).to.be.rejectedWith(`Input validation error: cannot find project`);
   });
   it('try to import from incorrect git path', async () => {
-    const gitModule = 'https://github.com/CyberismoCom/i-do-not-exist.git';
-    const result = await expect(
-      commands.importCmd.importModule(gitModule, commands.project.basePath),
-    ).to.be.rejected;
-    expect(result.message).to.include('Failed to clone module');
+    if (skipTest) {
+      skip(
```
samutoljamo-pareto, line 100: Noticed that this `skip` function was imported from `node:test`(node's own test runner) although we use mocha, which also has the function(see https://mochajs.org/#inclusive-tests). Instead of `skip`, we should be able to just call `this.skip()`. ´

samutoljamo-pareto (approved):
Looks good to me, left a few comments

tools/data-handler/test/module-manager.test.ts
Fuzzbender: Yes. Though it requires **not** to use arrow function, but it is ok.

Fuzzbender:
There are still some issues with "SimpleGit + Git + CI + Windows". 
I'll merge these, and try to think why it still fails from time to time.

PR #801, opened 2025-08-12
INTDEV-954 Enable linting for 'backend' and 'node-clingo'

Fuzzbender:
Enable listing for 'node-clingo' and 'backend' sub-projects.

It might be good idea to define the interfaces for the backend methods instead of using `unknown`.

### DIFF
--- a/tools/backend/eslint.config.js
+++ b/tools/backend/eslint.config.js
@@ -1,0 +1,19 @@
+// @ts-check
+
+import eslint from '@eslint/js';
+import tseslint from 'typescript-eslint';
+import eslintConfigPrettier from 'eslint-config-prettier';
+
+export default [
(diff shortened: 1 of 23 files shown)


samutoljamo-pareto (approved):
Seems good, but there might be a few conflicts since #792 is merged. And for the backend types, the reason it mainly uses unknown is because the types are stored on the frontend. There are at least two solutions:
1. Introduce a common package that would contain type definitions. It could also store other helpers from DH. For example, currently the metadata files from DH are imported to app from DH, but that is a bit risky, DH uses a lot of node features and doing the wrong import will break the app, since node imports are not available (That's the reason why the metadata are separated)
2. https://hono.dev/docs/guides/rpc

PR #802, opened 2025-08-13
INTDEV-885 allow deleting resources

ghost:
This allows deleting all resources, except calculations. Error messages are shown if there is one. In practice, most delete operations fail because they are used somewhere and thus cannot be deleted.

I did not create a workaround for calculations, but rather used the resource API. Once the calculations are implemented as a resource, deleting the calculations will work too.

I also added `zod` to the backend, since it works very well with typescript and hono. It allows parsing/validating unknown objects. It does support validating more complex rules than just types and it is able to provide the typescript type from the definition directly. https://zod.dev/basics

### DIFF
--- a/tools/backend/src/domain/resources/service.ts
+++ b/tools/backend/src/domain/resources/service.ts
@@ -340,9 +340,7 @@ async function groupResourcesByPrefix(
 /**
  * Delete a resource.
  * @param commands Command manager.
- * @param module Name of the module.
- * @param type Name of the type.
- * @param resource Name of the resource.
+ * @param resource Resource to delete.
  */
 export async function deleteResource(
   commands: CommandManager,
--- a/tools/backend/src/domain/resources/schema.ts
+++ b/tools/backend/src/domain/resources/schema.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod';
 
 export const resourceParamsSchema = z.object({
+  identifier: z.string(),
   prefix: z.string(),
   type: z.enum([
     'graphModels',
@@ -13,7 +14,6 @@ export const resourceParamsSchema = z.object({
     'fieldTypes',
     'linkTypes',
   ]),
-  identifier: z.string(),
 });
 
 export type ResourceParams = z.infer<typeof resourceParamsSchema>;


tools/app/src/components/context-menus/CardContextMenu.tsx
```diff
@@ -22,17 +22,21 @@ import {
...
 } from '@mui/joy';
 import { useTranslation } from 'react-i18next';
-import { useModals } from '../lib/utils';
-import { MoveCardModal, DeleteModal, AddAttachmentModal } from './modals';
-import { useAppSelector, useAppRouter } from '../lib/hooks';
-import { useCard } from '../lib/api';
-import { useParentCard } from '../lib/hooks';
+import { useModals } from '@/lib/utils';
+import {
+  MoveCardModal,
```
Fuzzbender, line 27: Just general comment that imports from one source could be in alphabetical order. 
No need to do anything in this PR, but we should have a tool (ESlint plugin?) that does these automatically, so that they are always in the same way across the project.

tools/backend/src/domain/resources/service.ts
```diff
@@ -298,6 +336,27 @@ async function groupResourcesByPrefix(
 
   return { rootResources, moduleResources };
 }
+
+/**
+ * Delete a resource.
+ * @param commands Command manager.
```
Fuzzbender, line 342: The documented params do not match the function signature.

samutoljamo-pareto:
Yes, there is a notification, which says whatever DH says. When the resource is in use, you get this: `Cannot delete resource ${resourceNameToString(this.resourceName)}. It is used by: ${usedIn.join(', ')}`

tools/app/src/components/context-menus/ConfigContextMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function ConfigContextMenu({ node, enabled }: ConfigContextMenuProps) {
+  const { t } = useTranslation();
+  const { modalOpen, openModal, closeModal } = useModals({
+    delete: false,
+  });
+
+  return (
+    <>
```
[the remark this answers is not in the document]
samutoljamo-pareto: Required here

PR #803, opened 2025-08-14
INTDEV-955, INTDEV-933 C++ code formatting and analysis

Fuzzbender:
### C++ code analysis ###

C++ code in `tools/node-clingo` is now analysed by the GitHub. It also checks that code is formatted. 

### C++ code formatting ###

C++ code formatting is now using Clang formatter. The rules are supporting older version of Clang formatter, to make the installation simpler for all platforms. 

I tried to avoid formatting the existing code as little as possible.
The base C++ style is now based on LLVM (I think it was closest for existing code base).

#### Deviations ####
However I made two small styling decisions : 

1) Pointer and reference are part of the type, and not name
E.g.  `int* pointer;` instead of `int *pointer;`
IMHO it makes no sense to attach the information to the variable name,

2) Enforced whitespace before a control command (`if`, `switch`, `for` ...)
E.g.  `if (test)` instead of `if(test)`
IMHO improves code readability. 

On other cases, even if I had thought that something else is my preference, I opted to keep the existing code.

There were some inconsistencies in the current code, so multiple ways of doing certain code style couldn't be supported (for example, how `switch-case` is intended). I just made the call which way is now auto-formatted.

### New PNPM tasks ###

The new tasks for `tools/node-clingo` are: 
 * `format:cpp` - Format C++ files in place using Clang-formatter
 * `format:cpp:check` - Check formatting - used by developers
 * `format:cpp:optional` - Optional formatting - used by prebuildify
 * `format:cpp:check:optional` - Optional format check - used by build

The optional tasks are to avoid making the clang-format required by the builds.

### Additionally ### 

Suggesting developers to install precommit hook to auto-format C++ code. Then it should be harder to push non-formatted code to git (and then get an error from C++ code check). Additionally, recommend to install the clang-formatter to your favourite editor. 

FYI: Pre-commit sample that enforces that formatting needs be correct before a commit:
```
#!/bin/sh
#
# Sample pre-commit hook for enforcing clang-format on C++ files
# This hook prevents commits that contain unformatted C++ code in tools/node-clingo
#
# To install this hook, copy it to .git/hooks/pre-commit and make it executable:
#   cp <source location>/pre-commit.sample .git/hooks/pre-commit
#   chmod +x .git/hooks/pre-commit
#

# Check if clang-format is available
if ! command -v clang-format >/dev/null 2>&1; then
    echo "⚠️  clang-format is not installed or not in PATH"
    echo "⚠️  Skipping C++ format check"
    echo ""
    echo "To enable C++ format checking, install clang-format:"
    echo "  macOS: brew install clang-format"
    echo "  Ubuntu: sudo apt-get install clang-format"
    echo "  Windows: choco install llvm"
    echo ""
    # Skip format check but allow commit
    exit 0
fi

# Get list of staged C++ files in tools/node-clingo
CPP_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -E "tools/node-clingo/src/.*\.(cc|cpp|cxx|h|hpp)$")

if [ -z "$CPP_FILES" ]; then
    # No C++ files to check
    exit 0
fi

echo "Checking C++ code formatting with clang-format..."

# Check if any C++ files need formatting
NEEDS_FORMAT=false
for file in $CPP_FILES; do
    if [ -f "$file" ]; then
        # Check if file needs formatting (compare with clang-format output)
        if ! clang-format --dry-run --Werror "$file" >/dev/null 2>&1; then
            echo "❌ $file is not properly formatted"
            NEEDS_FORMAT=true
        else
            echo "✅ $file"
        fi
    fi
done

if [ "$NEEDS_FORMAT" = true ]; then
    echo ""
    echo "❌ Some C++ files are not properly formatted!"
    echo ""
    echo "To fix the formatting, run:"
    echo "  cd tools/node-clingo && npm run format:cpp"
    echo ""
    echo "Or format individual files:"
    echo "  clang-format -i <filename>"
    echo ""
    echo "Then stage and commit your changes again."
    exit 1
fi

echo "✅ All C++ files are properly formatted"
exit 0
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,42 @@
+# Clang-Format configuration for Cyberismo C++ code
+#
+# Note: clang-format is optional for CI builds - if not available, formatting checks are skipped
+# Developers should install clang-format for automatic code formatting.
+
+# Use LLVM; try to avoid any exceptions.
+BasedOnStyle: LLVM
+Language: Cpp
+ColumnLimit: 120
+
+# Indentation
+IndentWidth: 4
+TabWidth: 4
+UseTab: Never
+NamespaceIndentation: All
+IndentCaseLabels: true
+PenaltyBreakComment: 1000
+ReflowComments: false
+
+# Braces
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterControlStatement: Always
+  AfterEnum: true
+  AfterStruct: true
+  AfterFunction: true
+  AfterNamespace: true
+  BeforeElse: true
+  BeforeCatch: true
+
+# Enforce parameters on own rows, function return value on same row
+AllowAllParametersOfDeclarationOnNextLine: false
+AlwaysBreakAfterReturnType: None
+PenaltyReturnTypeOnItsOwnLine: 1000
+BinPackArguments: false
+BinPackParameters: false
+
+# Put pointers and references to types
+PointerAlignment: Left
+
+# When bracket opens, align arguments to new lines
+AlignAfterOpenBracket: AlwaysBreak
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,6 @@
 # Clang-Format configuration for Cyberismo C++ code
 #
-# Note: clang-format is optional for CI builds - if not available, formatting checks are skipped
+# Note that formatting checks for CI builds are skipped if clang-format cannot be found.
 # Developers should install clang-format for automatic code formatting.
 
 # Use LLVM; try to avoid any exceptions.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -4,11 +4,6 @@
 # You may wish to alter this file to override the set of languages analyzed,
 # or to provide custom queries or build logic.
 #
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL Advanced"
 
 on:
@@ -47,6 +42,8 @@ jobs:
           build-mode: none
         - language: javascript-typescript
           build-mode: none
+        - language: c-cpp
+          build-mode: manual
         # CodeQL supports the following values keywords for 'language': 'actions', 'c-cpp', 'csharp', 'go', 'java-kotlin', 'javascript-typescript', 'python', 'ruby', 'swift'
         # Use `c-cpp` to analyze code written in C, C++ or both
         # Use 'java-kotlin' to analyze code written in Java, Kotlin or both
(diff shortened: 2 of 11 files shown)


.github/workflows/cpp-quality.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "  cd tools/node-clingo && npm run format:cpp"
+          echo ""
+          echo "Or format individual files:"
+          echo "  clang-format -i <filename>"
+          exit 1
+        fi
+
+        echo "All C++ files are properly formatted"
+
+  build-test:
```
samutoljamo-pareto, line 71: Do we need this? It uses `npm install`, which does not build it, but rather uses a prebuilt one if it exists. Our other CIs also use the install command while container CIs build it always so those CIs cover this already

tools/node-clingo/package.json
```diff
@@ -4,11 +4,15 @@
   "description": "Node.js bindings for Clingo answer set solver",
   "scripts": {
     "install": "node scripts/download-prebuild.js || node-gyp-build",
-    "build": "tsc -p tsconfig.build.json",
-    "build-prebuildify": "shx rm -rf prebuilds && prebuildify --napi",
+    "build": "npm run format:cpp:check:optional && tsc -p tsconfig.build.json",
```
samutoljamo-pareto, line 7: replace `npm run` with `pnpm`(also in `build-prebuildify`)

tools/node-clingo/package.json
```diff
@@ -4,11 +4,15 @@
...
+    "build-prebuildify": "npm run format:cpp:optional && shx rm -rf prebuilds && prebuildify --napi",
     "dev": "tsc -p tsconfig.build.json --watch",
     "test": "vitest run",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "format:cpp": "clang-format -i src/*.cc src/*.h",
+    "format:cpp:check": "clang-format --dry-run --Werror src/*.cc src/*.h",
+    "format:cpp:optional": "command -v clang-format >/dev/null 2>&1 && npm run format:cpp || echo 'clang-format not available, skipping C++ formatting'",
```
samutoljamo-pareto, line 14: Same here

.clang-format
```diff
@@ -0,0 +1,42 @@
+# Clang-Format configuration for Cyberismo C++ code
+#
+# Note: clang-format is optional for CI builds - if not available, formatting checks are skipped
```
samutoljamo-pareto, line 3: Doesn't the CI fail if format is invalid?

.github/workflows/codeql.yml
```diff
@@ -59,40 +56,53 @@ jobs:
...
     # "We were unable to automatically build your code", modify the matrix above
     # to set the build mode to "manual" for that language. Then modify this step
     # to build your code.
-    # ℹ️ Command-line programs to run using the OS shell.
-    # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
-    - if: matrix.build-mode == 'manual'
+    # Command-line programs to run using the OS shell.
+    # See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+    # Manual build for C++ (node-clingo native addon)
+    - name: Build C++ code
```
samutoljamo-pareto, line 98: Does code ql require this? Reading the docs, I did get the feeling that building is required, but on the other hand, the CodeQL pipeline seems to only check source files: https://github.com/CyberismoCom/cyberismo/actions/runs/16965337290/job/48087730027?pr=803

samutoljamo-pareto (approved):
Looks good, left a few comments

.clang-format
Fuzzbender: "optional" in sense that if it cannot find clang-format, formatting checks are not skipped. 
I didn't want to force installation of clang-format everywhere where we build the stack.

.github/workflows/codeql.yml
Fuzzbender: I think it needs to build it, to ensure that it can find all the interfaces and their implementations as such.
There is a preview that is not requiring a build for C/C++: 

https://github.blog/changelog/2025-06-03-codeql-can-be-enabled-at-scale-on-c-c-repositories-in-public-preview-using-build-free-scanning/

Once that is released, we can remove the build

PR #805, opened 2025-08-15
SECURITY FIX: Remove race condition possibility when creating preferences file

Fuzzbender:
Use 'wx' flag when writing preferences file. This will ensure that the error is thrown and exception is handled, if the file exists.

This will fix: https://github.com/CyberismoCom/cyberismo/security/code-scanning/108

### DIFF
--- a/tools/data-handler/src/utils/user-preferences.ts
+++ b/tools/data-handler/src/utils/user-preferences.ts
@@ -9,10 +9,11 @@
     You should have received a copy of the GNU Affero General Public
     License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
+import { dirname } from 'node:path';
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 
-import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/utils/user-preferences.ts
```diff
@@ -95,17 +95,23 @@ export class UserPreferences {
...
+      }
+
+      // Try to create file exclusively using 'wx' flag
+      // This will fail atomically if file already exists
+      writeFileSync(this.prefsFilePath, formatJson(UserPreferences.defaults), {
+        flag: 'wx',
+      });
+    } catch (error: any) {
+      // If file already exists (EEXIST), that's fine - we'll use the existing file
+      if (error.code !== 'EEXIST') {
```
samutoljamo-pareto, line 114: Could put a `logger.warn` statement in else. I think it's good practice when ignoring exceptions

samutoljamo-pareto: Also I think you could use `error?.code` so it won't fail if error does not have `code`

samutoljamo-pareto (approved):

PR #806, opened 2025-08-15
INTDEV-913 INTDEV-914 View logic program context menu

ghost:
This adds "View logic program" to context menu in both configuration and card view.

### DIFF
--- a/tools/data-handler/src/containers/project/calculation-engine.ts
+++ b/tools/data-handler/src/containers/project/calculation-engine.ts
@@ -420,7 +420,6 @@ export class CalculationEngine {
 
   /**
    * Gets the logic program content for a specific resource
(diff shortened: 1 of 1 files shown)


tools/data-handler/src/commands/show.ts
```diff
@@ -374,6 +385,17 @@ export class Show {
     return Array.from(new Set(labels));
   }
 
+  /**
+   * Shows the content of a logic program.
+   * @param resource Name of the resource.
+   * @returns the content of the logic program.
+   */
+  public async showLogicProgram(resource: ResourceName) {
+    const logicProgram =
```
Fuzzbender, line 394: Could just `return this.project.calculationEngine.resourceLogicProgram(resource);`

The only excuse not to do that is debugging and this is only called from `showCardLogicProgram` which stores the return value anyway to local variable.

PR #807, opened 2025-08-15
Remove unused function from scripts

Fuzzbender:
As found out by code scanning the function in scripts is not used anywhere: https://github.com/CyberismoCom/cyberismo/security/code-scanning/93

Remove the unused function.

### DIFF
--- a/scripts/install-dev-packages.js
+++ b/scripts/install-dev-packages.js
@@ -152,24 +152,6 @@ function installClingo(condaCommand) {
 }
 
 /**
- * Check if an executable is available in the PATH.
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto:
Does anyone use that script? @CyberJML @samilempinen 

I don't think it's used by CI or installers, so unless someone needs it, it could be deleted. We could document, what needs to be installed for development(for native building + ruby/asciidoctor in addition to clingo)

samutoljamo-pareto:
Closing PR since #836 removed the script

PR #814, opened 2025-08-19
Use node v22 consistently in workflows

Fuzzbender:
New change was using node v.18.

Removed the `build: manual` (and whole build step for C++ code), since it would require that the code is re-built for each change. Moved to `build: none` (=statistical analysis) to avoid this. This came up already in the PR for C++ analysis. 
This change also removed nodeJS v.18 usage.

Also noticed that the package.json was using wrong case when referring to the repo. The actual name is "CyberismoCom/...", not "cyberismocom/...". It should work with either case but better use the real one.

As a future improvement it might make sense to put the node version to a environmental variable and use that everywhere, to ensure that all the workflows are using the same version (and need to update just one place when updating nodeJS). Something like: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -43,7 +43,7 @@ jobs:
         - language: javascript-typescript
           build-mode: none
         - language: c-cpp
-          build-mode: manual
+          build-mode: none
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (approved):
We could add the variable when we migrate to nodeJS v24

Fuzzbender:
> We could add the variable when we migrate to nodeJS v24

I'll write a story about it, so we don't forget.

PR #816, opened 2025-08-19
INTDEV-884 Create resources

samutoljamo-pareto:
This PR will allow creating resources. What was mentioned in specs but is not part of this:
* Display names: Almost all resources have display names and they could be asked while creating a resource. I thought this could be its own PR, since it requires lots of changes to the API
* Creating template cards: Currently we can only create cards from templates, so I thought I'll do it in a separate PR as this is already quite big
* Automatically checking whether a resource already exists with the given identifier instead of allowing the user to try to create one and fail
* Populate display name(as it does not exist)


tools/app/src/components/modals/resource-forms/FieldTypeForm.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { DataType } from '@cyberismo/data-handler/interfaces/resource-interfaces';
+import { ResourceFormProps } from './BaseResourceModal';
+import ProjectIdentifier from './Identifier';
+
+interface FieldTypeFormData {
+  identifier: string;
+  dataType: DataType;
+}
+
+const dataTypeOptions: Array<{ value: DataType; label: string }> = [
```
Fuzzbender, line 35: The label names should probably support translations (ie. `t` function).

tools/app/src/components/modals/resource-forms/Identifier.tsx
```diff
@@ -0,0 +1,21 @@
...
+      }}
+    >
+      {prefix} / {type} /
+    </Typography>
+  );
+}
+
+export default function ProjectIdentifier({ type }: { type: string }) {
+  const { project } = useProject();
+  return <Identifier prefix={project?.prefix || ''} type={type} />;
```
Fuzzbender, line 20: How can there be a `project` but no `prefix`?

tools/app/src/components/modals/resource-forms/LinkTypeForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation.
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { useForm, Controller } from 'react-hook-form';
```
Fuzzbender, line 14: Couldn't the resource modals share some of the functionality? 
It looks a bit much that each resource type has its own modal, when they are 90% the same.

tools/app/src/lib/constants.ts
```diff
@@ -11,3 +11,18 @@
 */
 
 export const MAX_RECENTS_STORED = 20;
+
+// These are resources that can be created from the toolbar
+export const RESOURCES = [
```
Fuzzbender, line 16: Could the `resourceNodes` in `tools/app/src/lib/api` and this array be somehow merged or shared?
Now both list all the resource types again. 
There is a similar problem in `data-handler`.  
Once we add new resource types, there are a lot of places that need to be adjusted.

tools/app/src/lib/definitions.ts
```diff
@@ -81,6 +82,53 @@ export interface CardView {
   timestamp: string;
 }
 
+// Data for creating a new field type
+export type CreateFieldTypeData = {
```
Fuzzbender, line 86: How about putting the `identifier` to "base resource type" and then inherit all the resource types from that (except Calculation). That way, if we need to adjust all resources, only the base need to be adjusted.

Fuzzbender (approved):
Looks good to me. Some questions and minor comments.

tools/app/src/components/modals/resource-forms/Identifier.tsx
samutoljamo-pareto: That's not possible, but there can be no `project`(before it's loaded) and that's what `project?.prefix` implies

tools/app/src/components/modals/resource-forms/LinkTypeForm.tsx
samutoljamo-pareto: They are currently very similar, but I don't think they will be. Though I could add "BaseForm" that will contain functionality related to the form / update the button since they will always be the same. I'd like to keep them in separate components though

tools/backend/src/domain/cardTypes/schema.ts
```diff
@@ -0,0 +1,18 @@
...
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+import { z } from 'zod';
+
+export const createCardTypeSchema = z.object({
+  identifier: z.string().min(1),
```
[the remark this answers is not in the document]
Fuzzbender: We already have schemas in JSONs so it makes more sense to use them as-is. Let's not re-produce the JSON schema with `zod`, since then if we need to adjust the schema we would need to adjust the binary as well. Currently in DH (in some cases), you can update the schema without making changes to binary which is nice.
Zod seemingly cannot use directly use JSON Schemas at all, which is weird design decision. 

In some cases it would make sense to use zod in DH; when there is a non-file based validation; such as `isValidProjectName` (and similar functions).  I'll chalk up a story.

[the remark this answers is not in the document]
samutoljamo-pareto: Zod's main selling point is the typescript support that nowadays is also in yup. It's hard to infer types from json schema so that has not been added. Though there is this code generator: https://www.npmjs.com/package/json-schema-to-typescript

PR #817, opened 2025-08-19
INTDEV-958 Use a global validator rather than recreating it

ghost:
The vega schema is large and recreating it for each macro caused validation to slow down


Fuzzbender:
Tests need some additional attention, but otherwise looks good to me.

PR #818, opened 2025-08-21
INTDEV-582 Change 'workflow' in card type

Fuzzbender:
Allow `update` command to change `workflow` in a card type. 
This causes all cards that are using the card type to be changed to a new state according to the state mapping table provided. 

The implementation uses a new parameter on CLI that is provided with a flagged parameter to a local file. The file should have just single property `stateMapping` object having keys for each current state and value for those to match new states. 
Screenshot 2025-08-22 at 9 35 25

Before changing the `workflow` value, it is checked that all states in the current workflow are mapped to new states, and that there are no invalid new states (that do not exist in the upcoming workflow). 

**Additionally** 

1) Changed `update` implementation so that if not a JSON value is given as `target`, it defaults to string (previously it would have just failed). This allows giving the parameter as name `demo/workflows/test` instead of JSON `{ "name": "demo/workflows/test" }. This makes it a lot smoother to use on CLI (where you additionally need to escape the JSON double quotes).
2) Added `logger` for all resources to the base class. Not sure if it could be aliased, as the current `AbstractResource.logger` looks a bit daft. Using just `Resource.logger` would be better. 
3) Fixed the command handler update tests so that they are not reliant on earlier tests.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -63,6 +63,7 @@ export interface CardsOptions {
   date?: string;
   version?: string;
   revremark?: string;
+  mappingFile?: string;
 }
 
 // Commands that this class supports.
@@ -356,12 +357,53 @@ export class Commands {
           parsedValue = value;
         }
 
+        // Handle mapping file for workflow changes
+        let mappingTable: { stateMapping: Record<string, string> } | undefined;
+        if (
+          options.mappingFile &&
+          operation === 'change' &&
+          key === 'workflow'
+        ) {
+          try {
+            const mappingData = await readJsonFile(
+              resolveTilde(options.mappingFile),
+            );
+            if (
+              mappingData &&
+              typeof mappingData === 'object' &&
+              'stateMapping' in mappingData
+            ) {
+              mappingTable = mappingData as {
+                stateMapping: Record<string, string>;
+              };
+            } else {
+              throw new Error(
+                'Mapping file must contain a "stateMapping" object',
+              );
+            }
+          } catch (error) {
+            throw new Error(
+              `Failed to read mapping file: ${errorFunction(error)}`,
+            );
+          }
+        }
+
+        let parsedNewValue = newValue;
+        if (newValue) {
+          try {
+            parsedNewValue = JSON.parse(newValue);
+          } catch {
+            parsedNewValue = newValue;
+          }
+        }
+
         await this.commands?.updateCmd.updateValue(
           resource,
           operation as UpdateOperations,
           key,
           parsedValue,
-          newValue ? JSON.parse(newValue) : undefined,
+          parsedNewValue,
+          mappingTable,
         );
       } else if (command === Cmd.updateModules) {
         const [module] = args;
--- a/tools/data-handler/src/resources/card-type-resource.ts
+++ b/tools/data-handler/src/resources/card-type-resource.ts
@@ -12,6 +12,7 @@
   License along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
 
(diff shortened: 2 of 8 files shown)


tools/data-handler/src/resources/card-type-resource.ts
```diff
@@ -315,6 +360,57 @@ export class CardTypeResource extends FileResource {
...
+
+    const currentWorkflowStates = currentWorkflow.states.map(
+      (state) => state.name,
+    );
+    const mappedSourceStates = Object.keys(stateMapping);
+    const unmappedCurrentStates = currentWorkflowStates.filter(
+      (stateName) => !mappedSourceStates.includes(stateName),
+    );
+
+    if (unmappedCurrentStates.length > 0) {
```
samutoljamo-pareto, line 392: Could also be `if(currentWorkflowStates.some(stateName => !mappedSourceStates.includes(stateName))) {...}`

tools/data-handler/src/resources/card-type-resource.ts
```diff
@@ -315,6 +360,57 @@ export class CardTypeResource extends FileResource {
...
+    }
+
+    // Also verify that all target states exist in the new workflow
+    const newWorkflowStates = newWorkflow.states.map((state) => state.name);
+    const mappedTargetStates = Object.values(stateMapping);
+    const invalidTargetStates = mappedTargetStates.filter(
+      (stateName) => !newWorkflowStates.includes(stateName),
+    );
+
+    if (invalidTargetStates.length > 0) {
```
samutoljamo-pareto, line 407: Same here

tools/data-handler/src/resources/resource-object.ts
```diff
@@ -68,6 +70,11 @@ export type Operation<T> =
  * Abstract class for resources.
  */
 export abstract class AbstractResource {
+  protected static get logger() {
+    return getChildLogger({
```
samutoljamo-pareto, line 74: If we had an abstract `getType`-method, we could create a non static logger that is part of each class and contains a reference to resource type

tools/data-handler/src/command-handler.ts
```diff
@@ -356,12 +357,53 @@ export class Commands {
...
+        if (
+          options.mappingFile &&
+          operation === 'change' &&
+          key === 'workflow'
+        ) {
+          try {
+            const mappingData = await readJsonFile(
+              resolveTilde(options.mappingFile),
+            );
+            if (
```
samutoljamo-pareto, line 371: Could use a zod schema, but not in this PR

samutoljamo-pareto (approved):
Looks good to me!

tools/data-handler/src/resources/card-type-resource.ts
Fuzzbender: I think I'll leave it as-is. With `some` the code would need to anyway then collect the states that do not match in order to show them in the error message. Collecting them in `some` callback does not make the code easier to read.

Fuzzbender: If you would throw from first missing state then it would be fine. But it would be a bit worse for the user (each state is complained separately; after first missing state is fixed, a new state nag is shown etc).

tools/data-handler/src/resources/resource-object.ts
Fuzzbender: Good comment. 

I added two new abstract getters to base-base class (`AbstractResource`), one for type, one for logger instance by name (this one is not real getter).

Then to base classes (`FileResource` and `FolderResource`) a new protected method that returns the `Logger` instance by using name. 

This way the inherited classes can just call `logger.info(...)` without needing to define anything, or import anything extra,

PR #819, opened 2025-08-21
INTDEV-886 Show validation errors for resources

ghost:
image

Validation errors for resources are shown on the UI. I think it would be better if we did not rely on string based errors, but I did not do that refactoring here in this PR

### DIFF
--- a/tools/data-handler/src/commands/validate.ts
+++ b/tools/data-handler/src/commands/validate.ts
@@ -931,10 +931,9 @@ export class Validate {
   /**
    * Validates a single resource.
    * @param resource Resource to validate
-   * @param resourceType Type of resource (e.g., 'cardType', 'fieldType', 'workflow', etc.)
    * @returns string containing all validation errors
    */
-  public async validateSingleResource(
+  public async validateResource(
     resourceName: ResourceName,
     project: Project,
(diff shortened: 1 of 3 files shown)


Fuzzbender:
> Since the resource classes are stateful, I wonder if we could keep track of it? Currently, if we call the validate method on a resource object, there is a chance that reading has failed(for example, if file does not exist). However, since there no way to know if resource has not been yet initialized or if reading has failed, it will still try to validate whatever is in the content-property

If it is seen as useful, there could be a state (initialised, readFromDisk, validated, ...). Alternatively, all but EEXISTS could be re-thrown from `initialize()`.

tools/data-handler/src/commands/validate.ts
```diff
@@ -928,6 +928,25 @@ export class Validate {
     return validationErrors.join('\n');
   }
 
+  /**
+   * Validates a single resource.
+   * @param resource Resource to validate
+   * @param resourceType Type of resource (e.g., 'cardType', 'fieldType', 'workflow', etc.)
```
Fuzzbender, line 934: This parameter does not exist in the function signature.

Fuzzbender (approved):
Looks good to me

PR #823, opened 2025-08-25
INTDEV-947 use show more directly

samutoljamo-pareto:
This also seems to fix INTDEV-956. After using normal fields for displaying more complex content like enums, notifications etc, clingo seemed to slow down significantly. I think one of the main issues is that we were using broad select facts, which caused unnecessary logic that had to be handled by clingo. Select does make sense for accessing properties that are part of the data, but having to use select for fields that only part of a specific query is unnecessarily complex. Thus, I added an extension to `showField`:
`showField(Key, Field, Value, DataType)` allows skipping select checks and results in a `#show field(Key, Field, Value, DataType)`. Similarly to `fields`, `showFields` allows specifying multiple fields directly. I recommend that `showField` should replace all cases, where a `field` must be created on the fly.

Note: This is a breaking change. A few months ago, `field(Key, Field, Value, DataType).` was added, but it has been now removed. That used to create a normal `field`, `dataType` and `showField` facts, so it did exactly what `showField` now does. I've thought this field fact has not been used yet by anyone else except me, but it would be probably good to check existing reports anyway.

### DIFF
--- a/tools/assets/src/calculations/common/queryLanguage.lp
+++ b/tools/assets/src/calculations/common/queryLanguage.lp
@@ -60,7 +60,23 @@ showField(Key, Field) :-
 % showField is also defined if showAll is defined so there is only a single show predicate for each result
 showField(Key, Field) :- showAll(Key), field(Key, Field, _).
 
-% Fields
+% Allow specifying up to 3 fields at the same time for showField, similar to field
+% For specifying 2 fields
+resultField(Key, Field, Value, DataType) :-
+    resultFields(Key, Field, Value, DataType, _, _, _).
+
+resultField(Key, Field, Value, DataType) :-
+    resultFields(Key, _, _, _, Field, Value, DataType).
+
+% For specifying 3 fields
+resultField(Key, Field, Value, DataType) :-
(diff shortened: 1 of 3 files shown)


Fuzzbender (approved):

samutoljamo-pareto:
At least personally I feel like `selectField` is a bit confusing, since `select` elsewhere means that it will display something that is already defined. I think the only issue with `showField` is that it is already used in the query language. What do you think of `resultField`? That does not exist in the query language and quite well represents that will be part of a `result` regardless of whether it is selected or not.

Fuzzbender:
> At least personally I feel like `selectField` is a bit confusing, since `select` elsewhere means that it will display something that is already defined. I think the only issue with `showField` is that it is already used in the query language. What do you think of `resultField`? That does not exist in the query language and quite well represents that will be part of a `result` regardless of whether it is selected or not.

`resultField` would be good one. Or if we want to have new term ... `pickField`.

henryhaverinen:
I think `resultField` is a great name. It refers to existing concepts and it is easy to understand

PR #824, opened 2025-08-25
INTDEV-927 Remove base module dependency and include module-test

samutoljamo-pareto:
I left import from git tests to DH, since they only call `showModules` after importing. These should not fail when data breaks(Unless the data has changed so substantially that `showModules` or importing itself fails). For now I removed the git import from CLI(replaced with a local import), do you think that's okay or should we try to keep it too?

Also: I removed base-module from module-test


tools/cli/test/cli.test.ts
```diff
@@ -2,11 +2,10 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import { expect, use } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-import { exec, execSync } from 'child_process';
+import { exec } from 'child_process';
 import { log } from 'console';
-import { existsSync, rmSync } from 'fs';
+import { rmSync } from 'fs';
```
Fuzzbender, line 7: Not really about this PR, but since you are re-arranging these...
Put to this import (and others from NodeJS) a prefix `node:<module>`

Fuzzbender (approved):

PR #827, opened 2025-08-26
INTDEV-950 create parameter schema for graph views 

ghost:
Added defaultGraphView to assests package and then modified create to simply copy that. Added validation of macro content to validate step in graph macro

### DIFF
--- a/tools/assets/src/static/defaultGraphView/parameterSchema.json
+++ b/tools/assets/src/static/defaultGraphView/parameterSchema.json
@@ -1,8 +1,7 @@
 {
-  "title": "Graph view",
   "$id": "myGraphMacroSchema",
+  "additionalProperties": false,
   "description": "Parameters for the graph macro",
-  "type": "object",
   "properties": {
     "model": {
       "description": "The name of the graph model",
@@ -17,6 +16,7 @@
       "type": "string"
     }
   },
-  "additionalProperties": false,
-  "required": ["model", "view"]
+  "required": ["model", "view"],
+  "title": "Graph view",
+  "type": "object"
 }
--- a/tools/data-handler/test/utils/resource-utils.test.ts
+++ b/tools/data-handler/test/utils/resource-utils.test.ts
@@ -247,7 +247,7 @@ describe('resource utils with Project instance', () => {
           prefix: 'base',
           type: 'type',
           identifier: 'identifier',
-        }
+        },
       ],
     ]);
     for (const name of validNames) {


tools/data-handler/src/macros/base-macro.ts
```diff
@@ -52,7 +52,10 @@ abstract class BaseMacro {
     input: unknown,
   ): Promise<string>;
 
-  protected abstract handleValidate(input: unknown): void;
+  protected abstract handleValidate(
+    context: MacroGenerationContext,
+    input: unknown,
+  ): void;
```
Fuzzbender, line 58: We have not been using `void`, just leave out.

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -132,8 +100,59 @@ class ReportMacro extends BaseMacro {
...
+    }
+
+    if (schema) {
+      validateMacroContent(this.metadata, input, schema);
+    }
+
+    return options;
+  }
+
+  private resolveResourcePath(
```
Fuzzbender, line 135: Make a similar function to src/utils/resource/resource-utils.ts with signature: 
```
function resolveResourcePath(
    resourceFolder: string,
    name: string,
    fileName: string,
  ): string { ... }
```

And then make this one call that one. 

The functionality is very close with existing `resourceNameToPath`. Could the functions be merged? This basically adds `filename` to the path instead of using `<identifier>.json`. 

Also, as this does not really resolve a path (in the sense that JS terminology uses resolve with paths), maybe rename just to `resourcePath()`.

tools/data-handler/src/resources/graph-view-resource.ts
```diff
@@ -73,10 +74,13 @@ export class GraphViewResource extends FolderResource {
     }
 
     await super.create(newContent);
-    const handleBarFile = join(this.internalFolder, 'view.lp.hbs');
-    await writeFileSafe(handleBarFile, '', {
-      flag: 'wx',
-    });
+    await cp(
```
Fuzzbender, line 77: Maybe use `copyDir` like the other resources are now using and let's change that function (in another PR) to use `cp` (now that `cp` is no longer experimental). That way the functionality of various resources looks similar.

tools/data-handler/src/resources/graph-view-resource.ts
Fuzzbender, line 1: Graph view resource should probably validate the `parameters.json` validation is invoked.

tools/data-handler/src/macros/base-macro.ts
samutoljamo-pareto: Can't leave it out, because it's an abstract method. Without explicit `void` you'll get: `'handleValidate', which lacks return-type annotation, implicitly has an 'any' return type.`

tools/data-handler/src/resources/graph-view-resource.ts
samutoljamo-pareto: Isn't that handled by the .schema file?
```
[
  {
    "version": 1,
    "id": "jsonSchema",
    "file": "parameterSchema.json"
  }
]

```

Fuzzbender: Is that validated when `cyberismo validate` is called? Or only when the card with a graph is validated on UI?
At least testing with report's parameter.json file with invalid values, `cyberismo validate` does not seem to react to any validation errors in that file (e.g. modify the `required` to contain a property name that is not in the json file).

tools/data-handler/src/macros/graph/index.ts
```diff
@@ -132,8 +100,59 @@ class ReportMacro extends BaseMacro {
...
-    return validateMacroContent<GraphOptions>(this.metadata, input);
+  private parseOptions(
+    input: unknown,
+    context: MacroGenerationContext,
+  ): GraphOptions {
+    const options = validateMacroContent<GraphOptions>(this.metadata, input);
+
+    let schema: Schema | null = null;
+    try {
+      schema = JSON.parse(
```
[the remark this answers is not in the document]
Fuzzbender: Yes. We could append the folder resource files to the content (which is ~metadata). Maybe something like
```
content: { 
         ...  (whatever there currently is)
         files: [ { path: <fullPathToFile, content: <fileContent> } ]
 }
 ```
 
Then we could have additional APIs for folder based resources to:
- list the files they have (instead of using `show()` for this)
- update a file
- add new file (not sure if we need this - well templates could benefit for adding a card)
- remove file

I don't think we need "bulk" operations that add/update/delete multiple files.

This way the file operations would be encapsulated into resources (which might be beneficial in the long run).
I'll do a ticket about this.

tools/data-handler/src/resources/graph-view-resource.ts
samutoljamo-pareto: Yes. We could consider using `avj` instead of `jsonschema`: https://ajv.js.org/strict-mode.html 

Seems like it is also much more used(149M weekly vs 4M)

tools/data-handler/src/macros/graph/index.ts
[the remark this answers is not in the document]
Fuzzbender: Here: https://cyberismo.atlassian.net/browse/INTDEV-1007

PR #828, opened 2025-08-26
ITNDEV-974 Add highlighting for selected resource

samutoljamo-pareto:
Adds highlighting for the selected resource in config editor. Also opens treemenu so that the selected resource is visible during refreshes

### DIFF
--- a/tools/app/src/lib/utils.ts
+++ b/tools/app/src/lib/utils.ts
@@ -32,6 +32,7 @@ import {
 } from '@cyberismo/data-handler/types/queries';
 import { CardResponse } from './api/types';
 import { AppConfig } from './definitions';
+import type { ResourceNode } from './api/types';
 
 // Gets type of a child of an array
 type ItemType<T> = T extends (infer U)[] ? U : never;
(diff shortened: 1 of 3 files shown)


tools/app/src/lib/utils.ts
```diff
@@ -238,6 +239,25 @@ export function findPathTo(
...
+ * Finds a resource node anywhere in the resource tree by its full name
+ * @param nodes: resource tree
+ * @param name: name of the resource node to find
+ * @returns resource node if found, otherwise null
+ */
+export function findResourceNodeByName(
+  nodes: ResourceNode[] | undefined,
+  name: string,
+): ResourceNode | null {
+  if (!nodes) return null;
```
Fuzzbender, line 252: I think this is not needed. If `nodes` does not have a value, it will skip the `for` loop and return `null` from line 258.

Fuzzbender (approved):

tools/app/src/lib/utils.ts
samutoljamo-pareto: trying to loop over `undefined` gives a `TypeError`, because `undefined` is not an iterable

PR #830, opened 2025-08-27
INTDEV-975 Allow adding cards to templates

samutoljamo-pareto:
Added the possibility to adds cards to templates using the configuration editor


tools/app/src/components/modals/resource-forms/NewTemplateCardModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const handleCreate = async (data: CreateTemplateCardData) => {
+    const createdCards = await createTemplateCard(
+      templateResource,
+      data.cardType,
+      parentCardKey,
+    );
+    if (createdCards.length && project?.prefix) {
+      return `${project.prefix}/cards/${createdCards[0]}`;
+    }
+    return templateResource;
```
Fuzzbender, line 48: Can you explain the return values in this function?
If there is a project defined and card creation succeeds, it returns the first created card, otherwise the template resource itself. When does the latter happen?

tools/app/src/locales/en/translation.json
```diff
@@ -89,6 +89,7 @@
       "preview": "Creating cards with this macro is not supported in the preview mode"
     }
   },
+  "templateCard": "Template card",
```
Fuzzbender, line 92: These translations could be in alphabetical order.

tools/backend/src/domain/templates/service.ts
```diff
@@ -27,3 +28,26 @@ export async function createTemplate(
 ) {
   await commands.createCmd.createTemplate(templateName, '');
 }
+
+export async function addTemplateCard(
+  commands: CommandManager,
+  template: string,
+  cardType: string,
+  parentKey?: string,
+  count: number = 1,
```
Fuzzbender, line 37: I think this could be `count?: number`. If the count is not defined in D-H, then it automatically will use 1.

tools/app/src/lib/hooks/utils.ts
```diff
@@ -296,3 +298,72 @@ export function useIsInCards() {
...
+    templateResource,
+    parentCardKey,
+  };
+}
+
+function findTemplateForCard(
+  nodes: ResourceNode[],
+  cardKey: string,
+): string | null {
+  function dfs(node: ResourceNode, ancestors: ResourceNode[]): string | null {
```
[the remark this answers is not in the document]
samutoljamo-pareto: Though dfs is quite known term. `depthFirstSearch` would work although it's not a general implementation, since it is not exposed.

tools/app/src/components/modals/resource-forms/NewTemplateCardModal.tsx
samutoljamo-pareto: Since after each create, we need to redirect the user somewhere, `BaseResourceModal` redirects user to whatever `createFn` returns. Thus, on success we want to redirect the user to the created card. Note that if adding cards fails, it will throw and stop the operation so even if the operation fails, the latter does not happen. However, since we've defined that the type of the response is `string[]`, we need to check that there is at least one element in the array. In practice, this should not happen, but I prefer the sanity check so that UI won't fail even if DH gave an empty list. So it will simply redirect to the template the cards were added to.

PR #831, opened 2025-08-27
INTDEV-976 Allow editing calculation files

samutoljamo-pareto:


henryhaverinen:
Looks good!

I'd like the Update button to appear only after the user has made any changes. In that case there should also be a Cancel button to cancel the changes. If the user clicks cancel, there should be a similar confirmation dialog as in the card editor: "You have unsaved changes. Are you sure you want to leave?". If the user confirms the cancelling, then the file should be reverted to the original version and the buttons should disappear.

tools/backend/src/domain/calculations/service.ts
```diff
@@ -12,10 +12,21 @@
...
   await commands.createCmd.createCalculation(fileName);
 }
+
+export async function updateCalculation(
+  commands: CommandManager,
+  prefix: string,
+  identifier: string,
+  content: string,
+) {
+  const name = resourceName(`${prefix}/calculations/${identifier}`, true);
```
Fuzzbender, line 30: Not related to this PR, but god damn I hate that boolean parameter from code readability point-of-view.
I guess, I need to split the `resourceName` function to strict and non-strict versions. :)

tools/data-handler/src/interfaces/resource-interfaces.ts
```diff
@@ -20,7 +20,9 @@ import type { Schema } from 'jsonschema';
  */
 
 // Calculation does not have own metadata.
-export type CalculationMetadata = ResourceBaseMetadata;
+export interface CalculationMetadata extends ResourceBaseMetadata {
```
Fuzzbender, line 23: Maybe fix the comment, as it now has its own metadata.

tools/backend/src/domain/calculations/service.ts
samutoljamo-pareto: `resourceName(`${prefix}/calculations/${identifier}`, { strict: true});` wouldn't be bad either

PR #832, opened 2025-08-28
INTDEV-986 Show error in image macro if attachment does not exist

ghost:
In inject mode, it wasn't checked whether the attachment exists before returning the result.

### DIFF
--- a/tools/data-handler/src/macros/image/index.ts
+++ b/tools/data-handler/src/macros/image/index.ts
@@ -94,6 +94,13 @@ export default class ImageMacro extends BaseMacro {
       throw new Error(`Card '${cardKey}' not found`);
     }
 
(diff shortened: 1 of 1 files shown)


tools/data-handler/test/macros/index.test.ts
```diff
@@ -753,6 +753,22 @@ Some content here`;
           'image::/api/cards/decision_1/a/the-needle.heic[alt="Test image",title="A test needle image"]',
         );
       });
+      for (const mode of ['static', 'inject'] as const) {
```
Fuzzbender, line 756: Is there a positive test where the file is found?

samutoljamo-pareto: Yes, above this

PR #833, opened 2025-08-28
INTDEV-987 Fix xref in export site

samutoljamo-pareto:
We have a problem: For now, we've only had 2 modes(excluding `validate`), which are `static` and `inject`. Static was meant to be used when only adoc is usable while inject was meant to be used for app, which supports html. However, the new static site falls between these: It could render content just like app can, but it's content is static so we do not want to render the `createCards` button.

So I added I third mode: `exportSite`. I feel like the naming is a bit confusing and it would probably be a better idea to simply move on to using `context` also in macros. Now we have both the mode and context, which also partially overlap.

### DIFF
--- a/tools/backend/src/domain/cards/lib.ts
+++ b/tools/backend/src/domain/cards/lib.ts
@@ -64,7 +64,7 @@ export async function getCardDetails(
   try {
     asciidocContent = await evaluateMacros(cardDetailsResponse.content || '', {
       context: staticMode ? 'exportedSite' : 'localApp',
-      mode: staticMode ? 'static' : 'inject',
+      mode: staticMode ? 'staticSite' : 'inject',
       project: commands.project,
       cardKey: key,
     });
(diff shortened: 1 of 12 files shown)


samutoljamo-pareto:
Context is used for as context for logic programs and while you could use a different mode and context, practically that does not happen ever, because there is no reason to run logic programs e.g. in "app" context when creating a pdf-document

Fuzzbender:
> We have a problem: For now, we've only had 2 modes(excluding `validate`), which are `static` and `inject`. Static was meant to be used when only adoc is usable while inject was meant to be used for app, which supports html. However, the new static site falls between these: It could render content just like app can, but it's content is static so we do not want to render the `createCards` button.
> 
> So I added I third mode: `exportSite`. I feel like the naming is a bit confusing and it would probably be a better idea to simply move on to using `context` also in macros. Now we have both the mode and context, which also partially overlap.

I am fine with using `context`.  The third mode could be called `staticSite` instead of `exportSite`. The "how" of site being generated (via export) does not need to be used outside of the actual export.

samutoljamo-pareto:
I agree, I renamed it

PR #834, opened 2025-08-28
INTDEV-985 Add metadata to pdf content

ghost:
This PR will add metadata and linked cards to PDF export. The query is very similar to card query(basically a subset of it), but there's no way currently to share the logic between these two, so that's why it was duplicated.

### DIFF
--- a/tools/assets/src/exportPdfReport/query.lp.hbs
+++ b/tools/assets/src/exportPdfReport/query.lp.hbs
@@ -108,6 +108,18 @@ field((Card, Field), "value", Value) :-
     childResult(Card, (Card, Field), "fields"),
     field(Card, Field, Value).
 
+resultField(((Card, Field), "value"), "displayValue", EnumDisplayValue, "shortText") :-
+    childResult(Card, (Card, Field), "fields"),
+    dataType(Card, Field, "enum"),
+    field((Card, Field), "value", Value),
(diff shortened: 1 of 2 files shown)


henryhaverinen:
Great stuff! Here are some observations:

- For enum and list fields, we should use the display name of the enum value instead of the technical name
- For Boolean fields, we should use "Yes" or "No" as display values, similarly as in the app
- Multi-line long text fields are not rendered correctly
- The names and values of the fields may get confused if there are null values, see the example about a decision card below

=>  Consider putting the metadata fields into a table with hidden borders

Decision in the app:

image

Decision in the PDF:

image

Fuzzbender:
> Great stuff! Here are some observations:
> 
> * For enum and list fields, we should use the display name of the enum value instead of the technical name
> * For Boolean fields, we should use "Yes" or "No" as display values, similarly as in the app
> * Multi-line long text fields are not rendered correctly
> * The names and values of the fields may get confused if there are null values, see the example about a decision card below
> 
> => Consider putting the metadata fields into a table with hidden borders
> 
> Decision in the app:
> 
> image
> Decision in the PDF:
> 
> image

Looking at the image; timestamp should generate as more readable date.

Fuzzbender (approved):

samutoljamo-pareto:
Table seemed fix formatting issues

PR #842, opened 2025-09-03
Release v0.0.11

samutoljamo-pareto:
Release version 0.0.11. This PR contains version bump changes for release 0.0.11.

### DIFF
--- a/.github/workflows/build-cyberismo-app-image.yml
+++ b/.github/workflows/build-cyberismo-app-image.yml
@@ -52,7 +52,7 @@ jobs:
     runs-on: cyberismo_ubuntu_x64
 
     permissions:
-      contents: read
+      contents: write
 
     steps:
       - name: Check out code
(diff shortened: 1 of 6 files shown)


samutoljamo-pareto:
There was in issue in the previous PR: While npmJS version would release any unreleased package, docker version would compare version of cli to version of cli in the previous commit. Since I rebased the release commit with 2 commits, docker build was not done. Also I noticed that when rebasing(and squashing) via github, the commit hash is changed. Thus the tag becomes invalid. Instead I made it so that the docker build will also tag the commit. Then it will use the tags to determine the previous version.

Minor inconvenience: if version is updated and some other branch is merged before release has been made by the first commit, the second commit will also run the release pipeline and overwrite the tag / docker image.

samutoljamo-pareto:
I think we could fix this by instead just depending on the tag directly, but it's a bit bigger change so I won't do it here. At least that's first thing that comes to my mind. E.g. something triggers a "tagger" CI(could be a CI that runs on every single commit) and tags the commit where the version change has been made --> publish/publish container builds react to the created tag

PR #856, opened 2025-09-05
Fix Dependabot when it gives errors from public monorepo subprojects.…

Fuzzbender:
Also increase open PR count to maximum.

Public subprojects give this kind of error: 
Screenshot 2025-09-05 at 11 40 10

`app` does not since it is `private`.

The recommended way to fix this is to use only Dependabot on root level and it **should** automatically pick up the requirements.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,25 +6,11 @@
 version: 2
 updates:
   - package-ecosystem: "npm" # See documentation for possible values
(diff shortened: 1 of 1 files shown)


samutoljamo-pareto (approved):
Not sure why private/public context matters in this context. Was this the reason why dependabot did not know how to update the lock file which is in the root?

PR #895, opened 2025-09-09
Switch to actual default hub URL

Fuzzbender:
Move the default URL location from feature branch to `main` branch.


samutoljamo-pareto (approved):
Seems good. What do you think of adding prefixes to the URL? While allowing `file:` urls will not be useful to the end user, we could refer to the file in unit tests

Fuzzbender:
> Seems good. What do you think of adding prefixes to the URL? While allowing `file:` urls will not be useful to the end user, we could refer to the file in unit tests

Let's do that in a separate PR.

PR #896, opened 2025-09-09
Make react dependency group

Fuzzbender:
Since some of the react dependencies are connected to each other (cannot update one unless the other is updated as well), move at least some of the React related dependencies to be updated with one PR. This will also lower the amount of PRs that the Dependabot creates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,14 +15,9 @@ updates:
       react-dependencies: # Update all React dependencies in one PR
         patterns:
           - "react"
-          - "react-arborist"
           - "react-dom"
-          - "react-hook-form"
-          - "react-redux"
-          - "react-router"
           - "@types/react"
           - "@types/react-dom"
-          - "@uiw/react-codemirror"
       chai-dependencies: # Update all Chai dependencies in one PR
         patterns:
           - "@types/chai"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,6 +12,17 @@ updates:
     rebase-strategy: "auto"
     open-pull-requests-limit: 20 # Increase from default of 5 to 20 (maximum allowed)
     groups:
+      react-dependencies: # Update all React dependencies in one PR
+        patterns:
+          - "react"
+          - "react-arborist"
```
samutoljamo-pareto, line 18: This is a 3rd party package

Fuzzbender:
> Not sure if this group is a good idea. There are lots of 3rd party packages in the list, do we really want to add these? Also if we do, then there's some packages missing, which do not have `react` in their name. Or maybe we could do `react-core` group, which only has react / react-dom. And if we need to group the 3rd party packages, we could keep the packages together that likely "very easy" to fix.

I intentionally left a lot of packages out. 

In the last PR-flood there were a couple of PRs that were targeting a specific one dependency but it failed since the other dependant dependency was not updated (DependBot cannot update two dependencies in one PR). So, at least those (react, react-dom and their types) should be in one group to make DB make nice PRs that actually can pass CI system.

samutoljamo-pareto:
Yes, those should absolutely go together. While breaking changes are quite rare, I wouldn't still add all the non-react related packages together. Now that the lock file is updated correctly, it's not really a problem that we need to update `react-hook-form` and the react version separately and when there breaking changes, I think it's nicer to have those separately.

PR #928, opened 2025-09-17
INTDEV-979 Metadata editors

ghost:
This PR includes metadata editor for resources. I'll do customFields, states, transitions etc in a different PR. I added a method called `applyResourceOperation` to DH. This is basically `updateValue`, but it takes in the operation object. White the syntax of `updateValue` does make sense to CLI, on the app it's just easier to create the `operation` object directly. `applyResourceOperation` does not have its own tests, but the tests of `updateValue` also test it.

UI/UX:
image
As you can see, when you modify fields, you'll get little buttons that allow you to save/cancel to edit. Various ways to save the field:
* Press enter(for text areas ctrl + enter)
* Lose focus(except for textareas). This might require more exceptions
* Press the save icon

### DIFF
--- a/tools/backend/test/resources.test.ts
+++ b/tools/backend/test/resources.test.ts
@@ -283,7 +283,7 @@ test('/api/resources/decision/cardTypes/decision/operation returns 500 for inval
         operation: {
           name: 'remove',
           target: '',
-          to: 'New Display Name'
+          to: 'New Display Name',
         },
       }),
     },
--- a/tools/app/src/components/config-editors/fields/index.ts
+++ b/tools/app/src/components/config-editors/fields/index.ts
@@ -1,6 +1,6 @@
-  export {default as BooleanInput} from './BooleanInput';
-  export {default as IdentifierInput} from './IdentifierInput';
-  export {default as MultiSelectInput} from './MultiSelectInput';
-  export {default as SelectInput} from './SelectInput';
-  export {default as TextareaInput} from './TextareaInput';
-  export {default as TextInput} from './TextInput';
+export { default as BooleanInput } from './BooleanInput';
+export { default as IdentifierInput } from './IdentifierInput';
+export { default as MultiSelectInput } from './MultiSelectInput';
+export { default as SelectInput } from './SelectInput';
+export { default as TextareaInput } from './TextareaInput';
+export { default as TextInput } from './TextInput';


samutoljamo-pareto:
As you can see in the backend tests, the endpoint responds 500 for invalid operations, which is not ideal, but I did implement validation on the backend for this, because then we'd have to keep those in sync. However, if we add zod to DH, we can import the schema to backend and do the validation before the operation

tools/app/src/lib/api/types.ts
```diff
@@ -174,10 +174,36 @@ interface FileNode extends BaseResourceNode {
 }
 
 // Union type for all possible resource nodes
-export type ResourceNode =
+export type AnyNode =
```
Fuzzbender, line 177: alphabetical order

tools/app/src/lib/api/types.ts
```diff
@@ -174,10 +174,36 @@ interface FileNode extends BaseResourceNode {
...
+  templates: TemplateNode;
+  reports: ReportNode;
+  graphModels: GraphModelNode;
+  graphViews: GraphViewNode;
+  calculations: CalculationNode;
+};
+
+export type GenericNode<T extends NodeKey> = NodeTypeMap[T];
+
+export type ResourceNode =
```
Fuzzbender, line 206: alphabetical order

tools/app/src/pages/configuration/resource.tsx
```diff
@@ -23,31 +23,31 @@ import {
...
 
-const resourceMap: Partial<
-  Record<NodeType, (node: ResourceNode, key: string) => React.ReactNode>
-> = {
+type ResourceRendererMap = {
+  [K in NodeKey]: (node: GenericNode<K>, key: string) => React.ReactNode;
+};
+
+const resourceMap: Partial<ResourceRendererMap> = {
   file: (node, key) => <TextEditor node={node} key={key} />,
```
Fuzzbender, line 31: alphabetical order

Fuzzbender (approved):
Looks good to me, some minor comments.

tools/backend/test/resources.test.ts
```diff
@@ -38,3 +39,78 @@ test('/api/resources/decision/cardTypes/decision/validate returns validation res
...
+
+  expect(response).not.toBe(null);
+  expect(response.status).toBe(200);
+
+  const result = await response.json();
+  expect(result).toHaveProperty('message');
+  expect(result.message).toBe('Updated');
+});
+
+test('/api/resources/decision/cardTypes/decision/operation performs add operation successfully', async () => {
```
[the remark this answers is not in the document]
samutoljamo-pareto: Good idea. I also changed tests so that it also checks whether these changes actually happened. However, they don't cover all cases, DH should cover that

PR #929, opened 2025-09-17
INTDEV-1007 Use resource content through existing resource APIs

Fuzzbender:
Add new property `content` for folder based resources (except `template` which has cards, and have its own APIs for handling them). 

Content for each folder resource can be seen through using `show()` and changed through `update()`. 
Accessing certain known content type uses well-known names (e.g. `schema`, `templateContent`). 

Interfaces for the folder resources have been modified. Report has been standardised not to have double `displayName` and `description` properties visible when using `show()`.  There are APIs in place (for internal use) that map well known names to actual filenames and vice versa. 

For example, to update the Graph Model calculations (in file 'model.lp'), one would call:
```
const res = new GraphModelResource(
    project,
    resourceName('decision/graphModels/newGraphModel'),
);
...
await res.update(
    {
        key: 'content',
        subKey: 'model', // note that this is now strongly typed; so calling with 'whatever' will cause compilation error
    },
    {
        name: 'change',
        target: '',
        to: '<new content>',
    },
);
```
<del>
 It is also possible to update ALL of the content (files) by using just `content` and then providing an object where each property match to well known property.
</del>

Folder resources have internal cache for files. Thus, the files are not read from disk again unless resource is being changed.

By default, CLI will show "basic" data of the resources. If using `--details` it will show `content` object as well.

**Once this has been merged:** 
- remove `Show::showFile()` method, should use resource specific `show()`
- remove `Show::showFileNames()` method, should use resource specific `show()`
- remove `Edit::editCalculation()` method, should calculation resource `update()`
- remove `Create::createCalculation()` method, should use normal create resource methods
- change `Edit:resourceContent()` to use `update()` with `content/<type>`
- for removed method callers, change to use various resource methods.

### DIFF
--- a/tools/cli/src/index.ts
+++ b/tools/cli/src/index.ts
@@ -989,6 +989,16 @@ program
         [type, typeDetail],
         Object.assign({}, options, program.opts()),
       );
+      // By default, do not show resources' content files
+      if (!options.details) {
+        if (
+          typeof result.payload === 'object' &&
+          result.payload !== null &&
+          'content' in result.payload
+        ) {
+          delete (result.payload as { content?: unknown }).content;
+        }
+      }
       handleResponse(result);
     }
   });
--- a/tools/data-handler/src/interfaces/folder-content-interfaces.ts
+++ b/tools/data-handler/src/interfaces/folder-content-interfaces.ts
@@ -31,13 +31,6 @@ export const REVERSE_FILE_MAPPINGS = {
   viewTemplate: 'view.lp.hbs',
 } as const;
 
-// Content interface for Report resources
-export interface ReportContent {
-  contentTemplate: string;
-  queryTemplate: string;
-  schema?: Schema;
-}
-
 // Content interface for Graph Model resources
 export interface GraphModelContent {
   model?: string;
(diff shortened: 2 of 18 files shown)


tools/data-handler/src/resources/folder-resource.ts
```diff
@@ -129,6 +217,35 @@ export class FolderResource extends FileResource {
     return files.filter((file) => VALID_FOLDER_RESOURCE_FILES.includes(file));
   }
 
+  /**
+   * Gets content of all files to properties.
+   * @returns object with property names as keys and file contents as values.
+   */
+  public async contentData(): Promise<Record<string, string | Schema>> {
```
samutoljamo-pareto, line 224: What do you think of caching these values? It does have some memory footprint, but we could save a few file system calls. However, they don't need to be read during initialization, but when contentData is called.

samutoljamo-pareto: Personally I feel like we should cache everything until it becomes problem. In reality, these files are usually quite small

tools/data-handler/src/interfaces/folder-content-interfaces.ts
```diff
@@ -0,0 +1,97 @@
...
+  return filename in GRAPH_VIEW_FILE_MAPPINGS;
+}
+
+// Get property name for a filename
+export function propertyName(filename: string): string | undefined {
+  return ALL_FILE_MAPPINGS[filename as keyof typeof ALL_FILE_MAPPINGS];
+}
+
+// Get filename with property name
+export function filename(propertyName: string): string | undefined {
```
samutoljamo-pareto, line 93: I think as could be avoided, if instead this and the function before would demand more accurate types. This will and should then demand the types eventually from the user's of this API. However, in the current format it would be a hard to the types, but that could be changed.

```diff
- await res.update('content/model', {
-       name: 'change',
-       to: '<new content>',
-     })
+ await res.update({
+       key: 'content',
+       subKey: 'model',
+       name: 'change',
+       to: '<new content>',
+     })
```

tools/cli/src/index.ts
```diff
@@ -679,7 +679,7 @@ importCmd
             options.projectPath,
           );
           const moduleListPath = resolve(projectPath, '.temp/moduleList.json');
-          let moduleListContent = '{ modules: [] }';
+          let moduleListContent = "{ 'modules': [] }";
```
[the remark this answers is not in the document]
Fuzzbender: Yes, I just needed the fix here as well; since the PRs are parallel.

tools/cli/src/index.ts
```diff
@@ -979,6 +979,16 @@ program
         [type, typeDetail],
         Object.assign({}, options, program.opts()),
       );
+      // By default, do not show resources' content files
+      if (!options.details) {
```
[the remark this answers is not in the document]
Fuzzbender: Could, but it would require a lot more wrangling with types. And it is the CLI that throws away the details, so it should be here.

PR #931, opened 2025-09-18
INTDEV-1015 CLI: Show version command

Fuzzbender:
Shows version information by using: `cyberismo show version`. 
For now, only singular version (data-handler's) is shown in the received object.

In the future, we can expand this with other sub-projects (though they should share same version name) and versions of the data (if we make data versional apart from binary version) and used modules.

### DIFF
--- a/tools/cli/test/cli.test.ts
+++ b/tools/cli/test/cli.test.ts
@@ -25,7 +25,11 @@ describe('Cli BAT test', function () {
         console.log(error);
       }
       expect(error).to.be.null;
-      expect(stdout).to.include('0.0.12');
+      // Expect version information to be <int>.<int>.<int>,
+      // where each can be single or multiple digits (0-9)
+      const re = /^(\d+\.)?(\d+\.)?(\d+)$/gm;
+      const valid = re.test(stdout.trim());
+      expect(valid).to.equal(true);
       done();
     });
   });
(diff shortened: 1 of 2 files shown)


samutoljamo-pareto (approved):
CLI test wouldn't hurt, since this uses relative paths and thus could break the docker container. Also `cyberismo show version` does not work outside of a cyberismo project. I feel like it would be more natural to leave `cyberismo show version` to project related versions and use `cyberismo -v` or `cyberismo --version` for the version of the binary.

tools/cli/test/cli.test.ts
```diff
@@ -19,6 +19,16 @@ describe('Cli BAT test', function () {
     rmSync(cliPath, { recursive: true, force: true });
     return true;
   });
+  it('check version', function (done) {
+    exec('cyberismo --version', (error, stdout, _stderr) => {
+      if (error != null) {
+        console.log(error);
+      }
+      expect(error).to.be.null;
+      expect(stdout).to.include('0.0.12');
```
samutoljamo-pareto, line 28: This will break the release script, since it depends on the version directly. Maybe you could check that is contains 3 numbers separated by dots?

tools/cli/test/cli.test.ts
```diff
@@ -19,6 +19,16 @@ describe('Cli BAT test', function () {
     rmSync(cliPath, { recursive: true, force: true });
     return true;
   });
+  it('check version', function (done) {
+    exec('cyberismo --version', (error, stdout, _stderr) => {
```
samutoljamo-pareto, line 23: Not related to this PR, but it seems like for some reason CLI tests do not use async tests

samutoljamo-pareto (approved):
Looks good, left a small comment related to tests

PR #934, opened 2025-09-23
INTDEV-1002 Change calculations to be real resources

Fuzzbender:
Calculations have been "sort of" resource for some time.
Implement them as normal folder-based resource with full support for resource APIs (rename, create, ...).
Added support CLI operations (`show`, `rename`, ...). 

The resource exists in `.cards/local/calculations`. Each resource is:
```
resourceName.json -- the metadata of the resource
<resourceName folder> -- internal folder
  calculation.lp -- the actual calculation file
```
For now, there is only support for one calculation file for each resource. This is only for convenience, it then resembles the other folder resources the most. Adding support for multiple different calculation files would not be too difficult to implement.

Accessing the internal files was done "properly" in PR https://github.com/CyberismoCom/cyberismo/pull/929.
There are still some issues with App wanting to use Edit Command with filenames instead of content types. 
Also, `graphViews` and `graphModels` still allow access to their files with certain APIs. They need to be cleaned up at some point.

**DATA BREAK**
This causes a data break with existing calculations.
`module-test` has been migrated. After merging this PR, all of the content repositories must be converted using the attached migrator. I can run the migrator for our internal modules. The migrator should be attached to Discussion topic as well (for those on the community who are not following Pull Requests).

[migrate_calculate_resources.zip](https://github.com/user-attachments/files/22505936/migrate_calculate_resources.zip)


tools/data-handler/src/resources/calculation-resource.ts
```diff
@@ -0,0 +1,172 @@
...
+    const existingName = this.content.name;
+    await super.rename(newName);
+    return this.onNameChange(existingName);
+  }
+
+  /**
+   * Shows metadata of the resource.
+   * @returns calculation metadata.
+   */
+  public async show(): Promise<Calculation> {
```
samutoljamo-pareto, line 110: Do we need to inject contentData? Or is it just for compatibility?

tools/data-handler/src/resources/file-resource.ts
```diff
@@ -242,6 +243,11 @@ export class FileResource extends ResourceObject {
           references.push(calculation.name);
         }
       } catch (error) {
+        // Skip files that don't exist (they may have been renamed or deleted)
+        if (hasCode(error) && error.code === 'ENOENT') {
+          console.warn(`Skipping non-existent file: ${filename}`);
```
samutoljamo-pareto, line 248: Could also add a `logger.warn`. Do you think this error is relevant to the user? If not, `console.warn` could be removed

tools/data-handler/src/resources/resource-object.ts
```diff
@@ -211,19 +213,26 @@ export class ResourceObject extends AbstractResource {
...
         try {
           const content = await readFile(filename, 'utf-8');
           const updatedContent = content.replaceAll(from, to);
           await writeFile(filename, updatedContent);
         } catch (error) {
-          throw new Error(
-            `Failed to process file ${filename}: ${(error as Error).message}`,
-          );
+          if (hasCode(error) && error.code === 'ENOENT') {
+            // Skip files that don't exist (they may have been renamed or deleted)
```
samutoljamo-pareto, line 228: `logger.warn`

samutoljamo-pareto (commented):
Notifications and Policy checks seem to be broken in `module-test`. CI also fails because e2e tests test those. Those are defined in the calculations of `module-test`

tools/data-handler/src/commands/rename.ts
```diff
@@ -341,6 +353,14 @@ export class Rename {
     }
     console.info('Updated templates');
 
+    const calculations = await this.project.calculations(
+      ResourcesFrom.localOnly,
+    );
+    for (const calculation of calculations) {
+      await this.updateCalculation(calculation.name);
```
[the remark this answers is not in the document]
Fuzzbender: To clarify what happens is that in "fetch all resources of certain type" only resource paths and filenames are returned, not resource objects. Thus, we need to instantiate the found resources by using `new` and call `rename` for them.

samutoljamo-pareto (approved):
Looks good. I just made the release so we should:
1. Merge this
2. Update modules
3. Do another release

PR #936, opened 2025-09-25
Feature/samutoljamo/intdev 1020

samutoljamo-pareto:


.github/workflows/prebuild.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           # Native builds
           - os: ubuntu-latest
             arch: x64
-          - os: macos-13
+          - os: macos-15-intel
```
Fuzzbender, line 74: Do we need the old Intel Mac support anymore? Those Macs are ~three years old or more.

tools/node-clingo/src/clingo_solver.cc
```diff
@@ -0,0 +1,245 @@
...
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+#include "clingo_solver.h"
+
+namespace node_clingo
+{
+    ClingoSolver::ClingoSolver() {}
+
+    ClingoSolver::~ClingoSolver() {}
```
Fuzzbender, line 19: Since both destructor and default constructor are empty, you can leave them out; compiler will create them for you automatically.

tools/node-clingo/src/clingo_solver.cc
```diff
@@ -0,0 +1,245 @@
...
+        {
+            delete[] atoms;
+            return false;
+        }
+
+        std::stringstream answerStream;
+        bool success = true;
+
+        // Convert each symbol to string
+        for (size_t i = 0; i < atoms_size && success; ++i)
```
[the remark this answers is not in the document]
Fuzzbender: Also why is `success` in the loop? It is set above, and afaik not changed in the loop.

tools/node-clingo/src/clingo_solver.cc
```diff
@@ -0,0 +1,245 @@
...
+            // Get the string representation
+            std::string symbolString = node_clingo::get_symbol_string(atoms[i]);
+
+            // If the string is empty, we skip it
+            if (symbolString.empty())
+            {
+                continue;
+            }
+
+            if (i > 0)
```
Fuzzbender, line 105: use parentheses: 

```
if (...) 
{
...
}
```

tools/node-clingo/src/clingo_solver.cc
```diff
@@ -0,0 +1,245 @@
...
+
+    SolveResult ClingoSolver::solve(const Query& query)
+    {
+        errorMessages.clear();
+        todayCalled = false;
+        auto t1 = std::chrono::high_resolution_clock::now();
+
+        // Initialize Clingo control
+        clingo_control_t* ctl = nullptr;
+        if (!clingo_control_new(
```
Fuzzbender, line 160: What are the magic numbers in the control creation? (0 and 20)
Could they have constant that would describe them?

tools/node-clingo/src/clingo_solver.cc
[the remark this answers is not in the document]
samutoljamo-pareto: The solve code was mostly copied from clingo's examples so the `success` might be a leftover from that. And no I can't use range-based loop with c-style arrays. Clingo forces us to use them. Here they are so lightly used that I don't think we benefit from converting it to an `std::vector`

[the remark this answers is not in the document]
samutoljamo-pareto: Rewrote this to use `std::unique_ptr` so it's safer and got rid of `success`

tools/node-clingo/src/clingo_solver.cc
samutoljamo-pareto: They could.
<dl class="params" style="font-weight: 400; font-size: 14px; font-family: Roboto, sans-serif; line-height: 22px; padding: 0px; margin-left: 0px; color: rgb(0, 0, 0); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><dt style="font-weight: bold;">Parameters</dt><dd>
[in] | arguments | C string array of command line arguments</br>
[in] | arguments_size | size of the arguments array </br>
[in] | logger | callback functions for warnings and info messages </br>
[in] | logger_data | user data for the logger callback </br>
[in] | message_limit | maximum number of times the logger callback is called </br>
</dd></dl>

The first is 0 as long as we do not use string-based arguments. I don't think it needs a constant, because we should likely use a `std::string` and use the `.size()` member function as its size.

The second one is the maximum amount of clingo log messages. This could definitely have a name. I could also increase to, let's say, 50.

tools/node-clingo/src/solve_result_cache.cc
```diff
@@ -0,0 +1,100 @@
...
+            currentBytes -= it->second.sizeBytes;
+            lru.erase(it->second.lruIt);
+            entries.erase(it);
+        }
+
+        std::size_t bytes = estimateSizeBytes(result);
+
+        if (bytes > CACHE_CAPACITY_BYTES)
+        {
+            return;
```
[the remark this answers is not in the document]
samutoljamo-pareto: With `pnpm build-debug` you can create a build that logs everything to stdout. In normal builds, nothing is done.

PR #961, opened 2025-10-08
INTDEV-1009 Allow adding multiple labels

samutoljamo-pareto:
At the same, I made a separate component for the label editor and also added tests for it.


tools/app/src/components/LabelEditor.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    const existingLabels = Array.isArray(value) ? [...value] : [];
+    const segments = inputValue
+      .split(LABEL_SPLITTER)
+      .map((segment) => segment.trim())
+      .filter((segment) => segment.length > 0);
+
+    // combine labels without duplicates
+    const newLabels: string[] = [];
```
Fuzzbender: Wouldn't it be simpler to initialise the `newLabels` to `existingLabels`? 
Then you wouldn't need to check two arrays on row 52 and spread two arrays on row 58.

Fuzzbender (approved):

tools/app/src/components/LabelEditor.tsx
samutoljamo-pareto: yes

PR #1068, opened 2025-11-28
Improve TreeMenu UX and drag-and-drop reliability

samilempinen:
This is a largely vibe-coded improvement to the TreeMenu usability.

### DIFF
--- a/tools/app/src/components/BaseTreeComponent.tsx
+++ b/tools/app/src/components/BaseTreeComponent.tsx
@@ -74,8 +74,7 @@ export function BaseTreeComponent<T>({
     [onMove],
   );
 
-  // Stable reference required by react-arborist: a fresh renderer function each
-  // render causes row components to remount, which aborts in-progress drags.
+  // react-arborist remounts rows when this ref changes, aborting in-progress drags.
   const renderNode = useCallback(
     (props: NodeRendererProps<T>) => (
       <NodeRenderer {...props} onNodeClick={onNodeClick} />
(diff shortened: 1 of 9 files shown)


henryhaverinen (approved):
I tested this and it clearly improves the UX

tools/app/src/components/BaseTreeComponent.tsx
```diff
@@ -55,17 +55,28 @@ export function BaseTreeComponent<T>({
...
-  const handleMove = (moveData: {
-    dragIds: string[];
-    parentId: string | null;
-    index: number;
-  }) => {
-    if (onMove && moveData.dragIds.length === 1) {
-      onMove(moveData.dragIds, moveData.parentId, moveData.index);
-    }
-  };
+  const handleMove = useCallback(
```
samutoljamo-pareto: In my experience, AI tends to be over aggressive with these optimizations. There is always a cost involved with using `useCallback` and `useMemo`. This reminded that we could enable React compiler, which should handle memoization, though it is not perfect yet, since it was released recently. No need to remove these though, since we can remove them once react compiler is enabled

tools/app/src/pages/cards/layout.tsx
```diff
@@ -55,6 +65,40 @@ export default function AppLayout() {
     },
   );
 
+  // Stable refs so the renderer chain down to BaseTreeComponent's renderNode
+  // useCallback stays valid mid-drag (react-arborist remounts rows on
```
samutoljamo: This comment is duplicated many times I think it could be less verbose. At least I'd remove references to components etc, because it will drift some day from the implementation

samutoljamo (approved):
Looks good to me!

PR #1069, opened 2025-11-28
INTDEV-923 Responsive UI

samilempinen:
Here is basic responsive support for the app. We cannot do much to wide content such as tables in the content itself, but it's a start.


tminkkinen:
This has been at least partly implemented in different pr:s -> closing

PR #1287, opened 2026-04-02
INTDEV-1243 Add session timeout functionality

CyberJML:
Current implementation did not function properly when we get 401 response from API. This implementation does show user session expired notification with a button to log in again. Also the state of the application is halted, so they can possibly copy paste the existing work before reauthenticating

### DIFF
--- a/tools/app/src/components/SessionExpiredBanner.tsx
+++ b/tools/app/src/components/SessionExpiredBanner.tsx
@@ -1,0 +1,62 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation. This program is distributed in the hope that it
+  will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+  of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Affero General Public License for more details.
+  You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { useTranslation } from 'react-i18next';
+import { useAppSelector } from '@/lib/hooks';
+import { Button, Sheet, Stack, Typography } from '@mui/joy';
+
+export default function SessionExpiredBanner() {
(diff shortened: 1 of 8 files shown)


tools/app/src/lib/swr.ts
```diff
@@ -40,8 +41,33 @@ export async function createApiCallError(
...
+  btn.addEventListener('click', () => window.location.reload());
+  banner.appendChild(text);
+  banner.appendChild(btn);
+  document.body.appendChild(banner);
+}
+
 async function handleResponse<T>(response: Response): Promise<T> {
+  // Once session expired banner is shown, block all API processing silently
+  if (isReloading) {
+    return new Promise<T>(() => {});
```
samutoljamo: If expired was just a redux state, we wouldn't need this

samutoljamo (commented):
Suggestion: Use redux to create a state, which represents that auth is invalid. Then have a global component, which will render tsome modal with "auth expired" and have a button that says "refresh" or something like that

tools/app/src/globals.css
```diff
@@ -1116,3 +1116,32 @@
...
+  justify-content: center;
+  gap: 12px;
+  padding: 12px 24px;
+  background: #c41c1c;
+  color: #fff;
+  font-size: 14px;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.2);
+}
+
+.session-expired-banner button {
```
samutoljamo: Just use the props available in the button component

tools/app/src/components/SessionExpiredBanner.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <Sheet
+      color="danger"
+      variant="solid"
+      sx={{
+        position: 'fixed',
+        top: 0,
+        left: 0,
+        right: 0,
```
samutoljamo, line 34: I'd prefer the modal component, but if it does not work for some reason, this is fine.

samutoljamo (approved):
Seems good!

tools/app/src/components/SessionExpiredBanner.tsx
CyberJML: It's a bad way allowing users to still interact with the edit view and possible changes, to store them before relog. Rare event, and wanted to have some "fail safe" for that situation

PR #1290, opened 2026-04-08
Feature/joonalepola/intdev 1100

CyberJML:
"Simple" solution to bring users valuable information of other users currently working with the same card that they are working. This will add a simple indicator whether somebody else is viewing/editing the current document, like this:

image

image

The viewers and editors are combined separately and by hovering over then, we can see the list of users.

This uses Hono's Server Sent Events: Open up a connection to the server, create event to the frontend, server sends new information when needed, GUI gets updated

### DIFF
--- a/tools/app/src/components/PresenceIndicator.tsx
+++ b/tools/app/src/components/PresenceIndicator.tsx
@@ -1,0 +1,99 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation. This program is distributed in the hope that it
+  will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+  of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Affero General Public License for more details.
+  You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { Chip, Stack, Tooltip, Typography } from '@mui/joy';
+import EditIcon from '@mui/icons-material/Edit';
+import VisibilityIcon from '@mui/icons-material/Visibility';
+import { useTranslation } from 'react-i18next';
+import type { PresenceEntry } from '@/lib/api/presence';
+import { getInitials } from '@/lib/utils';
+
+const MAX_VISIBLE = 2;
+
+const groups = [
+  {
+    mode: 'editing',
+    color: 'warning',
+    icon: <EditIcon sx={{ fontSize: 14 }} />,
+  },
+  {
+    mode: 'viewing',
+    color: 'neutral',
+    icon: <VisibilityIcon sx={{ fontSize: 14 }} />,
+  },
+] as const;
+
+interface PresenceIndicatorProps {
+  presence: PresenceEntry[];
+  currentUserId?: string;
+}
+
+export default function PresenceIndicator({
+  presence,
+  currentUserId,
+}: PresenceIndicatorProps) {
+  const { t } = useTranslation();
+
+  const others = currentUserId
+    ? presence.filter((e) => e.userId !== currentUserId)
+    : presence;
+
+  if (others.length === 0) return null;
+
+  return (
+    <Stack direction="row" gap={1} alignItems="center" sx={{ mx: 1 }}>
+      {groups.map(({ mode, color, icon }) => {
+        const users = others.filter((e) => e.mode === mode);
+        if (users.length === 0) return null;
+        const visible = users.slice(0, MAX_VISIBLE);
+        const hiddenCount = users.length - MAX_VISIBLE;
+        const tooltip = t(`presence.${mode}`, {
+          names: users.map((e) => e.userName).join(', '),
+        });
+
+        return (
+          <Tooltip
+            key={mode}
+            title={tooltip}
+            placement="bottom"
+            sx={{ whiteSpace: 'pre-line' }}
+          >
+            <Chip size="sm" variant="soft" color={color} endDecorator={icon}>
+              <Stack direction="row" gap={0.5}>
+                {visible.map((user) => (
+                  <Typography
+                    key={user.userId}
+                    level="body-xs"
+                    component="span"
+                    sx={{ fontWeight: 600 }}
+                  >
+                    {getInitials(user.userName)}
+                  </Typography>
+                ))}
+                {hiddenCount > 0 && (
+                  <Typography
+                    level="body-xs"
+                    component="span"
+                    sx={{ fontWeight: 600 }}
+                  >
+                    +{hiddenCount}
+                  </Typography>
+                )}
+              </Stack>
+            </Chip>
+          </Tooltip>
+        );
+      })}
+    </Stack>
+  );
+}
--- a/tools/app/src/components/toolbar/CardToolbar.tsx
+++ b/tools/app/src/components/toolbar/CardToolbar.tsx
@@ -22,12 +22,19 @@ import StatusSelector from '../StateSelector';
 import { findWorkflowForCardType } from '../../lib/utils';
 import { useAppRouter, useAppSelector } from '../../lib/hooks';
 import { useTranslation } from 'react-i18next';
-import { useCard, useProject, useTree } from '../../lib/api';
+import {
+  useCard,
+  usePresence,
+  useProject,
+  useTree,
+  useUser,
+} from '../../lib/api';
 import { useAppDispatch } from '../../lib/hooks';
 import { addNotification } from '../../lib/slices/notifications';
 import { getConfig } from '@/lib/utils';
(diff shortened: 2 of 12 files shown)


tools/backend/src/domain/cards/presence.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+interface Connection {
+  user: UserInfo;
+  mode: 'viewing' | 'editing';
+  send: (data: string) => void;
+}
+
+/**
+ * In-memory presence tracker for card editing/viewing.
```
adam-kuhn, line 30: Samu was talking that there may be a need to having multiple servers in the future. If this is in-memory, then I don't think this implementation would work in that case right?
Not an issue for now, but just raising here, so you are aware of what may come

CyberJML: Yeah, true. We will most likely have redis (already used in SaaS version) that we can utilize with this feature when it's needed

samutoljamo: Yes, but I'd think this is fine in this case. Likely this would be replaced with redis. I don't think it makes to abstact this yet in this PR, but we could think about whether we'd want to just commit to redis and use some in-memory implementation for CLI rather than create abstractions in our codebase.

tools/app/src/components/toolbar/CardToolbar.tsx
```diff
@@ -55,6 +62,9 @@ export function CardToolbar({
   const { tree } = useTree();
   const isEdited = useAppSelector((state) => state.page.isEdited);
   const { card, updateWorkFlowState, isUpdating } = useCard(cardKey);
+  const { user } = useUser();
```
samutoljamo, line 65: No need to do in this PR, but we could start using suspense. Basically, it will show a top-level fallback component until these kind of requests are done rather than handling loading in each component/page. It's kind of stupid that we need to handle the case where user is undefined in the subcomponents.

tools/app/src/components/PresenceIndicator.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <Tooltip
+            key={mode}
+            title={tooltip}
+            placement="bottom"
+            sx={{ whiteSpace: 'pre-line' }}
+          >
+            <Chip size="sm" variant="soft" color={color} endDecorator={icon}>
+              <Stack direction="row" gap={0.5}>
+                {visible.map((user) => (
+                  <span key={user.userId} style={initialsStyle}>
```
samutoljamo: Prefer `Typography` so that we don't break the theme

tools/app/src/components/PresenceIndicator.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    {getInitials(user.userName)}
+                  </span>
+                ))}
+                {remaining > 0 && (
+                  <span style={initialsStyle}>+{remaining}</span>
```
samutoljamo: Same here

samutoljamo: Would it be better if `remaining` would be a translation like:
`and x others`

tools/app/src/lib/api/presence.ts
```diff
@@ -0,0 +1,67 @@
...
+  mode: 'viewing' | 'editing' = 'viewing',
+): PresenceEntry[] {
+  const [editors, setEditors] = useState<PresenceEntry[]>([]);
+
+  useEffect(() => {
+    if (!cardKey || getConfig().staticMode) {
+      return;
+    }
+
+    const url = `/api/cards/${encodeURIComponent(cardKey)}/presence?mode=${mode}`;
```
samutoljamo: Could this be in `apiPaths.ts` under `swr.ts`

tools/app/src/lib/api/presence.ts
```diff
@@ -0,0 +1,67 @@
...
+    const eventSource = new EventSource(url);
+
+    eventSource.addEventListener('presence', (event) => {
+      try {
+        const data = JSON.parse(event.data) as { editors: PresenceEntry[] };
```
samutoljamo: Define the schema with zod and parse it here. Then this will also catch non-malformed packets, but which have the wrong type.

tools/app/src/lib/api/presence.ts
```diff
@@ -0,0 +1,67 @@
...
+        setEditors(data.editors);
+      } catch {
+        // Ignore malformed events
```
samutoljamo: can we do `console.warn` here for now?

tools/backend/src/domain/cards/index.ts
```diff
@@ -768,5 +674,57 @@ router.get(
...
+    );
+
+    stream.onAbort(() => {
+      presenceStore.remove(key, connId);
+    });
+
+    try {
+      // Keep connection alive with periodic heartbeat
+      while (true) {
+        await stream.writeSSE({ event: 'heartbeat', data: '' });
```
samutoljamo: I'd prefer an empty comment for heartbeat over heartbeat event. You should be able to use`stream.write(': hb\n\n')` here.

tools/backend/src/domain/cards/index.ts
```diff
@@ -768,5 +674,57 @@ router.get(
...
+        await stream.sleep(30000);
+      }
+    } finally {
```
samutoljamo: Looking at this, I am quite sure this would never be called, because `writeSSE` never throws and you have an infinite loop with no exit condition. I think the while loop could simply check whether the stream has been aborted or not. Then we only need one `presenceStore.remove`, either after the while loop or in `stream.onAbort`.

tools/backend/src/domain/cards/presence.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Broadcast current presence to all connected clients for a card.
+   */
+  private broadcast(cardKey: string): void {
+    const cardConns = this.connections.get(cardKey);
+    if (!cardConns) return;
+
+    const presence = this.getPresence(cardKey);
+    const data = JSON.stringify({ editors: presence });
+    const message = `event: presence\ndata: ${data}\n\n`;
```
samutoljamo: I'd use hono's `writeSSE` over manually writing the payload here.

tools/backend/src/domain/cards/presence.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    let hadFailures = false;
+    for (const [connId, conn] of cardConns.entries()) {
+      try {
+        conn.send(message);
+      } catch {
+        cardConns.delete(connId);
```
samutoljamo: Do we need this whole thing? If we use `writeSSE` it will just ignore errors. If client disconnects, it should be removed from the connections by the backend anyway. Broadcast could be just fire and forget

tools/backend/src/domain/cards/presence.ts
```diff
@@ -0,0 +1,131 @@
...
+        hadFailures = true;
+      }
+    }
+
+    if (hadFailures) {
```
samutoljamo: See above comment, then this could be removed

tools/backend/src/domain/cards/presence.ts
```diff
@@ -0,0 +1,131 @@
...
+  mode: 'viewing' | 'editing';
+  send: (data: string) => void;
+}
+
+/**
+ * In-memory presence tracker for card editing/viewing.
+ * Tracks which users are currently looking at or editing each card,
+ * and notifies all connected clients via SSE when presence changes.
+ */
+class PresenceStore {
```
samutoljamo, line 34: This wouldn't necessarily have to be a class, if you had 
```
function createPresenceStore() {
  const connections = ...;
  ....
  function broadcast () {
    ...
  }
  return {
      add: ...,
      remove: ....,
      getPresence: ...,
  }
```
More of a matter of taste so feel free to ignore, but I personally avoid classes if there's no need to use them.

tools/backend/test/presence.test.ts
```diff
@@ -0,0 +1,280 @@
...
+      noop,
+    );
+
+    const presence = presenceStore.getPresence('test-card-3');
+    expect(presence.length).toBe(1);
+    expect(presence[0].userId).toBe('user-1');
+    expect(presence[0].mode).toBe('editing');
+
+    // Cleanup
+    presenceStore.remove('test-card-3', conn1);
```
samutoljamo: I'd add some kind of `removeAll` method that could be used for tests. Then there's less risk of data leaking between the tests

tools/backend/test/presence.test.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    const reader = response.body!.getReader();
+    const decoder = new TextDecoder();
+    let text = '';
+
+    for (let i = 0; i < 10; i++) {
+      const { value, done } = await reader.read();
+      if (done) break;
+      text += decoder.decode(value, { stream: true });
+      if (text.includes('event: presence')) break;
```
samutoljamo: Will this work always or does pass only if text happens to include the whole event? This will break even if the whole event hasn't gone through? Maybe u could wait rather for `\n\n` as that marks the end of the message.

samutoljamo (approved):
Overall looks good, left a few comments

PR #1291, opened 2026-04-08
Fix error messages & Remove link modification from static site

CyberJML:
- Fix /api prefix matching to only match /api and /api/..., not unrelated paths like /apidocs or /apikey
- Return 404 instead of 400 for non-existent API routes
- Return 404 status for SPA fallback pages instead of 200 to prevent search engine indexing of non-existent pages
- Static site preview: return proper JSON 404 for missing API/JSON resources instead of serving
- Static site preview: return 404 status for unknown routes
- Hide link edit/delete buttons in exported static site

### DIFF
--- a/tools/backend/src/app.ts
+++ b/tools/backend/src/app.ts
@@ -98,7 +98,7 @@ export function createApp(
   // serve index.html for all other routes
   app.notFound(async (c) => {
     if (c.req.path.startsWith('/api')) {
-      return c.text('Not Found', 400);
+      return c.text('Not Found', 404);
     }
     const file = await readFile(
       path.join(import.meta.dirname, 'public', 'index.html'),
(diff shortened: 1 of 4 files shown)


tools/backend/src/app.ts
```diff
@@ -97,8 +97,8 @@ export function createApp(
 
   // serve index.html for all other routes
   app.notFound(async (c) => {
-    if (c.req.path.startsWith('/api')) {
-      return c.text('Not Found', 400);
+    if (c.req.path === '/api' || c.req.path.startsWith('/api/')) {
```
samutoljamo: `if(c.req.path.startsWith('/api')` ?

samutoljamo (approved):
Looks good to me

PR #1292, opened 2026-04-08
INTDEV-1227 Fix link creation, update & remove bugs

tminkkinen:
Fixes:
- Creation of inbound link
- Verifying new link is possible to create before removing old link during update
- Removing similar links when removing link without description


adam-kuhn (approved):
🚀 
Nothing else from my side - good stuff

PR #1294, opened 2026-04-09
Feature/adam kuhn/intdev 1232

adam-kuhn:
Enable project and card level pdf exports

This PR does not address adding a loading indicator as this would need to be discussed with design/Cyberismo to understand exactly how this should be handled. Based on that decision would determine how to update the current notification system, and that should be done in separate PR and ticket

- update the notification slice to have persistent toasts (no auto-closing). While the export request is pending show a persistent toast, then when the request resolved, close that specific toast and create a new success toast (or edit the persistent toast to a success state)
- Update the card tree to also include the cardKey,  property. Then use the useCard isUpdating method, to display a loading spinner beside the card title to indicate a pending request that affects that specific card
[slack thread](https://paretosoftwareoy.slack.com/archives/C074694EFAB/p1775718522519439)


tools/data-handler/src/macros/index.ts
```diff
@@ -409,11 +410,16 @@ export function createCodeBlock(content: string) {
...
  * @param image base64 encoded image
+ * @param mode The mode in which the image is being rendered
  * @param controls Add controls
  * @returns valid asciidoc with the image
  */
-export function createImage(image: string, controls: boolean = true) {
-  if (process.env.EXPORT_FORMAT) {
+export function createImage(
+  image: string,
+  mode: Mode,
```
samutoljamo, line 421: Bit of a nitpick, but what do you think would `passthrough: boolean` be better over `mode: Mode`? Then we wouldn't need to tie this to the macro context, but this would then require calling `context.mode === 'static'` in the methods that use it

adam-kuhn: I'm not sure I have enough context to fully answer this, but I feel like it would be best if we could rely on some sort on configuration (e.g `mode`) to behave the same way each time this method is called. That way the behavior is determined within `createImage` instead of the caller having the responsibility to set the behavior

CyberJML:
One thing I noticed while trying out this. There's no indicator that the PDF is being generated. I took a bit larger export and wasn't sure if it was working, until it did open the export to me

adam-kuhn:
> One thing I noticed while trying out this. There's no indicator that the PDF is being generated. I took a bit larger export and wasn't sure if it was working, until it did open the export to me

Yes, that has been left out of this ticket because exact behavior of how to handle this has not been defined. Depending on how to handle this would have a relatively large impact on some of the existing functionalities, so it will need to be done in a different PR

henryhaverinen:
Exporting works fine for both projects and cards. I think we should figure out how to improve the following parts:

- Could we simply show a regular notification to indicate that the PDF export is in progress? Then there would be another notification when it is ready. That could be better for now than not showing anything
- There is no way to set the version field on the title page, so it will be shown as "Version , :". I think we should just not include a version at all on the page, if it is not given
- When exporting the entire project, we should hide the disabled checkbox "Export child cards"

tminkkinen:
Updates:
- Added optional version field to the export modal, which appears on the PDF title page when provided
- Fixed PDF title page to not show stray punctuation (Version , and :) when version, date or revision remark are not set
- Added a progress notification while export is in progress, which is replaced by a success or error notification when done

tools/app/src/components/context-menus/CardContextMenu.tsx
```diff
@@ -92,12 +94,17 @@ export function CardContextMenu({ cardKey }: CardContextMenuProps) {
...
           </MenuItem>
           <Divider />
-          <MenuItem data-cy="deleteCardButton" onClick={handleDeleteClick}>
-            <Typography color="danger">{t('deleteCard')}</Typography>
-          </MenuItem>
           <MenuItem onClick={openModal('logicProgram')}>
             <Typography>{t('viewLogicProgram')}</Typography>
           </MenuItem>
+          <Divider />
+          <MenuItem onClick={openModal('exportCard')}>
```
samutoljamo: This should be hidden in a static site. Otherwise, we'd have to generate all possible combinations of pdf's the user might request, because there is no backend.

tools/app/src/components/modals/ExportCardModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+  forceChildExport = false,
+  heading,
+}: ModalProps) => {
+  const { t } = useTranslation();
+  const [exportChildCards, setExportChildCards] =
+    React.useState(forceChildExport);
+  const { card, exportCard } = useCard(cardKey);
+  const [title, setTitle] = React.useState(defaultTitle);
+  const [name, setName] = React.useState(defaultFileName);
+  const [version, setVersion] = React.useState('');
```
samutoljamo: a bit of a nitpick, but I'd import useState / useEffect instead of calling React.x

tools/app/src/components/modals/ExportCardModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+            fullWidth
+            id="name"
+            value={name}
+            onChange={(e) => setName(e.target.value)}
+          />
+          <Typography level="body-sm">
+            {t('exportCardNameDescription')}
+          </Typography>
+
+          <label htmlFor="version">
```
samutoljamo: Any reason for not using Joyui's form label?

tools/app/src/components/modals/ExportCardModal.tsx
```diff
@@ -0,0 +1,179 @@
...
+            <Typography>{t('version')}</Typography>
+          </label>
+          <Input
+            fullWidth
+            id="version"
+            value={version}
+            onChange={(e) => setVersion(e.target.value)}
```
samutoljamo: This is a small form, do you think it's better to use useState here vs react hook form for consistancy?

tools/app/src/lib/api/card.ts
```diff
@@ -178,6 +178,22 @@ export const useCardMutations = (key: string | null) => {
         null
       );
     },
+    exportCard: async ({
```
samutoljamo: I find it a bit weird that this is under card mutations? It's not mutating a card. Though since it uses `call`, it's part of the logic that determines whether a card is being updated or not. This mainly is used to disable buttons globally, is this behaviour required?

tools/app/src/lib/api/card.ts
```diff
@@ -224,3 +240,66 @@ export async function createCard(
...
+  const progressNotification = dispatch(
+    addNotification({
+      message: `Exporting ${title} to PDF...`,
+      type: 'info',
+      disableAutoClose: true,
+    }),
+  );
+  try {
+    const swrKey = apiPaths.exportCard();
+    const result = await fetch(swrKey, {
```
samutoljamo: Instead of doing fetch manually, I'd use `callApi`. It must be extended to support blobs, but I think it would be good, because now  this does not share the logic for handling an expired auth for example and  error parsing too

tools/app/src/lib/api/card.ts
```diff
@@ -224,3 +240,66 @@ export async function createCard(
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+    if (result.status !== 200) {
+      throw new Error(`Unable to export ${name}`);
+    }
+    const blob = await result.blob();
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement('a');
```
samutoljamo: This could be a helper, since it used in two places and I would believe there will be more users for this

tools/data-handler/src/commands/export.ts
```diff
@@ -208,6 +208,31 @@ export class Export {
     }
   }
 
+  private async evaluateExportContent(
+    options: ExportPdfOptions,
+  ): Promise<string> {
+    await this.project.calculationEngine.generate();
```
samutoljamo: Remove this, there is no need to do `generate()` manually

samutoljamo (approved):
Did a codewise review, overall looks good, but left a few comments

tools/app/src/components/modals/ExportCardModal.tsx
tminkkinen: I don’t think using hook form adds much value here right now since there’s no validation or complex state to manage. UseState keeps it somewhat simpler and seems to be used with similar modals we already have. For consistency it could make sense to use hook form here and in similar modals, but maybe that can be added later if needed

PR #1297, opened 2026-04-14
INTDEV-1089 Version support patch 1

samutoljamo:
First of a set of PRs realted to versioning. The following features are implemented here:

**cyberismo show modules:** shows all module names and installed version

**cyberismo show module <prefix>:** shows module details including installed version

**cyberismo show project:** shows both the project current version and all installed modules versions

**cyberismo update-modules [name] [version]**: Updates all modules or a specific one. If updating specific one, you may specify a version. 

Note that if you update a module, its dependencies are also updated. Because we don't have version ranges yet or anything, the transitive deps are updated to the latest version. A bit confusing behavior, but we need to enable version ranges before we could make this properly, but  I wanted to split PRs so that they're a bit more manageable to review instead of creating largest PR we've ever seen.

### DIFF
--- a/tools/data-handler/src/command-handler.ts
+++ b/tools/data-handler/src/command-handler.ts
@@ -23,6 +23,7 @@ import type {
   Credentials,
   HubSetting,
   ModuleContent,
+  ModuleInfo,
   ModuleSettingFromHub,
   ProjectMetadata,
   RemovableResourceTypes,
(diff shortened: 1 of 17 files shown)


tools/data-handler/src/containers/project.ts
```diff
@@ -1052,6 +1054,19 @@ export class Project extends CardContainer {
...
+  /**
+   * Returns installed modules with their versions.
+   */
+  public moduleInfos(): ModuleInfo[] {
+    return this.resources
+      .moduleNames()
+      .sort()
+      .map((name) => {
+        const setting = this.configuration.modules.find((m) => m.name === name);
+        return { name, version: setting?.version };
```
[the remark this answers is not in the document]
tminkkinen: Do you think that the amount of modules in project will raise so much this would have any observable difference? Probably not, but on the other hand copilot suggestion wouldn't hurt either

tools/data-handler/src/command-handler.ts
```diff
@@ -451,10 +452,19 @@ export class Commands {
...
+            module,
+            credentials,
+            targetVersion,
+          );
         } else {
+          if (targetVersion) {
+            throw new Error(
+              'A target version can only be specified together with a module name',
+            );
+          }
```
tminkkinen, line 467: Would it be clearer to follow a similar approach as pnpm here? Combining the package name and version into a single argument instead of passing them separately?

samutoljamo: That does sound like a good idea.

samutoljamo:
I close this in favor of the other PR

PR #1303, opened 2026-04-15
INTDEV-942 Include the project prefix in the URL

CyberJML:
Mostly placeholders waiting for the backend implementation, currently this is basically a gimmick that shows the correct project in the url, while allowing all the URLs to work properly. Validations and selections are automated to project given by /api/project, but those should be easy to modify to fit the backend in the future. 

URL will be: http://localhost:3000/projects/<prefix>/cards/<cardKey>

Current idea how this could work:
1. User sees a separate? view where they can select the project from a list (no previous projects selected and no last project information)
2. Selection of project/project got from the navigation URL is set as a currentProject for the gui (validations against url project)
3. React Router will use this project name as a basename (/projects/prefix), also API calls should have this project information with API calls
4. Switching project will change the currentProject value (probably requires the full page navigation to new project url to create new router?)
5. ???
6. Profit

### DIFF
--- a/tools/app/src/routes.ts
+++ b/tools/app/src/routes.ts
@@ -22,7 +22,9 @@ import ResourceOverviewPage from './pages/configuration/resource-overview';
 import Layout from './pages/layout';
 import ConfigLayout from './pages/configuration/layout';
 import NotFoundPage from './pages/not-found';
-import { isSafeRedirectPath } from './lib/utils.js';
+import { store } from './lib/store.js';
+import { selectProjectPrefix, setProjectPrefix } from './lib/slices/project.js';
+import { fetchAvailableProjects } from './lib/projectUtils.js';
 
 // Export mode guard - unfortunately need to refetch config.json to check export mode since hooks
 function createEditLoader(cardKey: string) {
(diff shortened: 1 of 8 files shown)


tools/app/src/lib/definitions.ts
```diff
@@ -40,6 +40,7 @@ export { type CardAttachment } from '@cyberismo/data-handler/interfaces/project-
 export type AppConfig = {
   staticMode: boolean;
   logoutUrl?: string;
+  projectPrefix?: string;
```
samutoljamo: I wouldn't put this to appconfig, these variables are usually for setting deployment specific values. These are like env variables of frontend, not meant to be changed during the runs. We use a json over env variables, because env variables are determined at build time for frontend and would require doing separate builds instead of just 1

tools/app/src/lib/projectRedirect.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+  return { prefix: undefined, basename: undefined, remainingPath: undefined };
+}
+
+/**
+ * Fetch the list of available project prefixes.
+ * Currently uses GET /api/project (single project) and returns a one-element array.
+ * TODO: Replace with GET /api/projects when multi-project backend is available.
+ */
+async function fetchAvailableProjects(): Promise<string[]> {
```
samutoljamo: If not now, then one there is multi-project support available, see how other requests are made and use the same style. This skips caching and doesn't have the same error handling

tools/app/src/routes.ts
```diff
@@ -50,86 +52,130 @@ function createEditLoader(cardKey: string) {
...
+          Component: NotFoundPage,
+        },
+      ],
+    },
+    {
+      // Catch legacy URLs without /projects/ prefix (e.g. /cards/KEY, /configuration/...)
+      path: '*',
+      loader: async ({ request }) => {
+        const { prefix } = await resolveProject();
+        const url = new URL(request.url);
```
samutoljamo, line 173: Does this become an infinite loop if you use a url like: `/thisdoesnotexist`

tools/app/src/routes.ts
```diff
@@ -50,86 +52,130 @@ function createEditLoader(cardKey: string) {
...
-    path: '/',
-    loader: () => {
-      if (window.location.search) {
-        const newPath = window.location.search.slice(2).split('&')[0];
-        if (newPath && isSafeRedirectPath(newPath)) {
-          return redirect(newPath);
-        }
-      }
-      return redirect('/cards');
+export function createAppRouter() {
```
samutoljamo, line 76: Why not just export this as like `export const router`. I wouldn't recreate the router ever

samutoljamo (approved):

tools/app/src/routes.ts
CyberJML: I noticed that the static mode would break and I'm fairly certain that it was because of this being created at the import time, when it needs to happen after await initConfig. After making this a factory, I can control when the router is created (after initialization of configuration) the static mode also works properly

tools/app/src/routes.ts
CyberJML: Wasn't able to identify any infinite loops. Unprefixed URLs get redirected to /projects/{prefix}/..., which matches a different route and terminates at NotFoundPage if the path is invalid. If no project can be resolved at all, the catch-all redirects to /, whose loader returns null (no further redirect), so it stops there too (currently a blank page, will be replaced with a project selection view). This implementation might change a bit when we do the project selection page along with the backend logic.

PR #1304, opened 2026-04-16
INTDEV-1125 Metadata & linked cards section updates

tminkkinen:


henryhaverinen:
This looks visually very good to me! 

The following things still need some work:

- It is no longer possible to edit links. We need to keep this functionality, because some link types (such as dataflows) include link descriptions and it would not be nice to only support deleting and creating links
- It is not possible to delete links
- It doesn't seem to be possible to cancel editing a metadata field
- I think we agreed to include the per-field editing experience in this story, so each field would be edited separately with an accept/cancel button similarly to how the configuration editor works. So we should not need the edit button on the metadata box level
- What is the checkmark icon at the end of each metadata field:

image

henryhaverinen:
Also, now that I'm testing this, the big blue edit button at the top right corner makes less sense after this change. Should we remove the global edit button completely, and include this:

image

I would also make the card contents to go to the edit mode when the user clicks the content (except when the user clicks an interactive element such as a button), because it is now going to be harder to find the edit button, if we move it at the top of the content part

tminkkinen:
> This looks visually very good to me!
> 
> The following things still need some work:
> 
> * It is no longer possible to edit links. We need to keep this functionality, because some link types (such as dataflows) include link descriptions and it would not be nice to only support deleting and creating links
> * It is not possible to delete links
> * It doesn't seem to be possible to cancel editing a metadata field
> * I think we agreed to include the per-field editing experience in this story, so each field would be edited separately with an accept/cancel button similarly to how the configuration editor works. So we should not need the edit button on the metadata box level
> * What is the checkmark icon at the end of each metadata field:
> 
> image

Clearly this still needs a bit work on UX, since some of the features seems hard to find / use. It is still possible to create, edit and remove links. Adding first link is possible from top bar "add link" button, and after that through the inline link editing view. Also in the edit view it shows remove button and when user edits any field of the link that changes to "save" button (checkmark).

Currently there is no separate button for canceling metadata field editing, it could be added if necessary. Canceling is possible by exiting the editing mode. The checkmark icon button is meant for saving the changes of the field and it will be disabled unless there is some changes that can be saved. Design was somewhat copied from the link section, maybe we could go through this on Monday?

> Also, now that I'm testing this, the big blue edit button at the top right corner makes less sense after this change. Should we remove the global edit button completely, and include this:
> 
> image
> I would also make the card contents to go to the edit mode when the user clicks the content (except when the user clicks an interactive element such as a button), because it is now going to be harder to find the edit button, if we move it at the top of the content part

I believe removing the edit button from the top bar completely is the goal, but I figured out not to touch the card content editing yet. I think it could be done after this on separate story / ticket.

samutoljamo:
Would be nice that it would be possible to also save by just exiting the field. Also, there's quite many buttons so I wonder if making them float / only appear when a single field is edited would make it feel nicer than displaying that many buttons.

PR #1305, opened 2026-04-17
Pre-build Clingo before CodeQL init so submodule isn't analyzed

samutoljamo:
CodeQL's manual build-mode only traces compilations that happen after `codeql-action/init` (which exports the tracer env vars for subsequent steps). Moving `cmake -P scripts/build-clingo.cmake` before init keeps the Clingo submodule's C++ sources out of the database; only our own node-clingo bindings get analyzed via the post-init `node-gyp rebuild`.


CyberJML (approved):
Seems to identify vulnerabilities nicely, just remember to remove them before merge :)

PR #1306, opened 2026-04-17
INTDEV-943 Add multiproject functionality

CyberJML:
Adds multi-project support. The backend can now serve multiple Cyberismo projects simultaneously under project-scoped API routes (/api/projects/:prefix/...). A new ProjectRegistry maps project prefixes to CommandManager instances, and a scanForProjects utility discovers projects from a directory. The frontend apiPaths were refactored to include the project prefix, and a placeholder project selector was added. Includes new tests for the registry, middleware, projects endpoint, and project scanner. The server and CLI fail fast with clear errors on invalid or empty project paths.

Future work for multiproject support

1. Empty project and project creation support
   Currently starting with an empty directory (no projects) results in a blank page.
   Need a project creation flow and an empty state view.

2. Dedicated project management view
   The project selector in SearchableTreeMenu is a placeholder.
   Replace with a proper project management view for browsing, creating, and managing projects.

3. Multi-project export support
   Export currently rejects multiple projects. Add support for selecting which
   project to export, or exporting all projects.

4. Make MCP project-scoped
   MCP endpoint currently uses registry.first() — hardcoded to the first project.
   Needs multi-project MCP support. Also, the /mcp route is mounted unconditionally;
   if a "no projects" server mode is added, either guard the mount or serve an
   "MCP unavailable" response.

### DIFF
--- a/tools/app/src/lib/swr.ts
+++ b/tools/app/src/lib/swr.ts
@@ -15,6 +15,82 @@ import { getConfig } from './utils';
 import { store } from './store';
 import { setSessionExpired } from './slices/session';
 
+/**
+ * Resolves the project prefix from the URL if not explicitly provided.
+ */
+function resolveProjectPrefix(projectPrefix?: string): string {
+  if (projectPrefix) return projectPrefix;
+  const match = window.location.pathname.match(/\/projects\/([^/]+)/);
+  if (!match) throw new Error('No project prefix found in URL');
+  return decodeURIComponent(match[1]);
+}
+
+/** Global API paths that are not project-scoped. */
+export const globalApiPaths = {
+  projects: () => '/api/projects',
+  user: () => '/api/auth/me',
+};
+
+/**
+ * Returns project-scoped API paths for the given project.
+ * If `projectPrefix` is omitted, it is resolved from `window.location.pathname`.
+ */
+export function projectApiPaths(projectPrefix?: string) {
+  const base = `/api/projects/${encodeURIComponent(resolveProjectPrefix(projectPrefix))}`;
+  return {
+    cards: () => `${base}/cards`,
+    card: (key: string) => `${base}/cards/${key}`,
+    rawCard: (key: string) => `${base}/cards/${key}?raw=true`,
+    calculations: () => `${base}/calculations`,
+    calculation: (name: string) => `${base}/calculations/${name}`,
+    cardTypes: () => `${base}/cardTypes`,
+    cardTypeFieldVisibility: (cardTypeName: string) =>
+      `${base}/cardTypes/${encodeURIComponent(cardTypeName)}/field-visibility`,
+    connectors: () => `${base}/connectors`,
+    fieldTypes: () => `${base}/fieldTypes`,
+    graphModels: () => `${base}/graphModels`,
+    graphViews: () => `${base}/graphViews`,
+    logicPrograms: (resourceName: string) =>
+      `${base}/logicPrograms/${resourceName}`,
+    cardType: (cardType: string) => `${base}/cardTypes?name=${cardType}`,
+    templates: () => `${base}/templates`,
+    templateCard: () => `${base}/templates/card`,
+    templateTree: () => `${base}/templates/tree`,
+    attachment: (cardKey: string, attachment: string) =>
+      `${base}/cards/${cardKey}/a/${encodeURIComponent(attachment)}`,
+    cardAttachments: (cardKey: string) =>
+      `${base}/cards/${cardKey}/attachments`,
+    cardAttachment: (cardKey: string, filename: string) =>
+      `${base}/cards/${cardKey}/attachments/${encodeURIComponent(filename)}`,
+    cardAttachmentOpen: (cardKey: string, filename: string) =>
+      `${base}/cards/${cardKey}/attachments/${encodeURIComponent(filename)}/open`,
+    cardLinks: (cardKey: string) => `${base}/cards/${cardKey}/links`,
+    cardParse: (cardKey: string) => `${base}/cards/${cardKey}/parse`,
+    linkTypes: () => `${base}/linkTypes`,
+    reports: () => `${base}/reports`,
+    workflows: () => `${base}/workflows`,
+    resources: (type: string) => `${base}/resources/${type}`,
+    resourceTree: () => `${base}/resources/tree`,
+    resource: (resourceName: string) => `${base}/resources/${resourceName}`,
+    resourceOperation: (resourceName: string) =>
+      `${base}/resources/${resourceName}/operation`,
+    tree: () => `${base}/tree`,
+    labels: () => `${base}/labels`,
+    validateResource: (resourceName: string) =>
+      `${base}/resources/${resourceName}/validate`,
+    project: () => `${base}/project`,
+    projectModulesUpdate: () => `${base}/project/modules/update`,
+    projectModuleUpdate: (module: string) =>
+      `${base}/project/modules/${encodeURIComponent(module)}/update`,
+    projectModuleDelete: (module: string) =>
+      `${base}/project/modules/${encodeURIComponent(module)}`,
+    projectModulesAdd: () => `${base}/project/modules`,
+    projectModulesImportable: () => `${base}/project/modules/importable`,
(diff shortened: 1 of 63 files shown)


tools/cli/src/index.ts
```diff
@@ -1463,45 +1465,68 @@ const appCmd = new CommandWithPath('app')
...
-  );
-  if (!result.message) {
-    program.error('Expected validation result, but got none');
-    return;
+  const basePath = options.projectPath || process.cwd();
+  const projects = await scanForProjects(basePath);
+
+  if (projects.length === 0) {
+    console.log('No projects found. Starting with empty project collection.');
   }
```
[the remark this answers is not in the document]
CyberJML: This is the work for the future feature in this feature branch

tools/backend/src/main.ts
```diff
@@ -55,12 +56,31 @@ function createAuthProvider(): AuthProvider {
   process.exit(1);
 }
 
-const projectPath = process.env.npm_config_project_path || '';
-const commands = await CommandManager.getInstance(projectPath);
+const projectPath = process.env.npm_config_project_path || process.cwd();
+const projects = await scanForProjects(projectPath);
```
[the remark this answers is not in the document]
CyberJML: Will be done in the future feature also, but added a small preventive measure

tools/backend/src/app.ts
```diff
@@ -59,33 +63,46 @@ export function createApp(
...
 
   // MCP endpoint for AI assistant integration
+  // TODO: Make MCP project-scoped when multi-project MCP is implemented
+  const mcpCommands = registry.first();
+  if (mcpCommands) {
+    const mcpMiddleware = attachCommandManager(mcpCommands);
+    app.use('/mcp', mcpMiddleware);
+    app.use('/mcp/*', mcpMiddleware);
+  }
   app.route('/mcp', mcpRouter);
```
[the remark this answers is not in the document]
CyberJML: Future feature

tools/backend/src/app.ts
```diff
@@ -59,33 +63,46 @@ export function createApp(
...
+  );
 
   // MCP endpoint for AI assistant integration
+  // TODO: Make MCP project-scoped when multi-project MCP is implemented
+  const mcpCommands = registry.first();
+  if (mcpCommands) {
+    const mcpMiddleware = attachCommandManager(mcpCommands);
+    app.use('/mcp', mcpMiddleware);
+    app.use('/mcp/*', mcpMiddleware);
+  }
```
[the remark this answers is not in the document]
CyberJML: Future

tools/data-handler/src/project-scanner.ts
```diff
@@ -0,0 +1,110 @@
...
+  prefix: string;
+  name: string;
+}
+
+/**
+ * Check whether a directory is a Cyberismo project (has `.cards/` and `cardRoot/`).
+ */
+function isProjectDirectory(dirPath: string): boolean {
+  return (
+    pathExists(join(dirPath, '.cards')) && pathExists(join(dirPath, 'cardRoot'))
```
samutoljamo, line 30: I think cardRoot is optional and not all projects have them

tools/data-handler/src/project-scanner.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Read project prefix and name from a project directory's cardsConfig.json.
+ */
+function readProjectMeta(projectPath: string): {
+  prefix: string;
+  name: string;
+} {
+  const configPath = join(projectPath, '.cards', 'local', 'cardsConfig.json');
+  const config = readJsonFileSync(configPath) as {
```
samutoljamo, line 42: Just use `ProjectConfiguration` for this so that we don't have many of reading the project configuration.

tools/data-handler/src/project-scanner.ts
```diff
@@ -0,0 +1,110 @@
...
+  } catch {
+    throw new Error(
+      `Cannot scan for projects: '${rootPath}' does not exist or is not a directory`,
+    );
+  }
+  const projects: ProjectEntry[] = [];
+  const seenPrefixes = new Map<string, string>();
+
+  for (const entry of entries) {
+    if (!entry.isDirectory()) continue;
```
samutoljamo: This allows symlinks, but I think that's fine, no?

tools/backend/src/app.ts
```diff
@@ -59,33 +63,46 @@ export function createApp(
...
+  projectScoped.route('/templates', templatesRouter);
+  projectScoped.route('/tree', treeRouter);
+  projectScoped.route('/workflows', workflowsRouter);
+  projectScoped.route('/resources', resourcesRouter);
+  projectScoped.route('/logicPrograms', logicProgramsRouter);
+  projectScoped.route('/labels', labelsRouter);
+  projectScoped.route('/project', projectRouter);
+
+  // In export mode (opts set), mount at the concrete prefix so SSG sees
+  // static routes instead of dynamic :prefix patterns it would skip.
```
samutoljamo, line 91: `ssgParams` is for this purpose, then you can use the dynamic path and ssgParams can return the prefix of the first project

tools/backend/src/app.ts
```diff
@@ -59,33 +63,46 @@ export function createApp(
...
+  const exportPrefix = opts ? registry.list()[0]?.prefix : undefined;
```
samutoljamo, line 92: `registry.first()`

tools/app/src/lib/swr.ts
```diff
@@ -14,6 +14,20 @@ import type { SWRConfiguration } from 'swr';
...
 import { store } from './store';
 import { setSessionExpired } from './slices/session';
+import { selectProjectPrefix } from './slices/project';
+
+/**
+ * Get the current project prefix from the Redux store.
+ * For use in non-hook contexts (standalone functions, event handlers).
+ * Throws if no project is selected.
+ */
+export function getProjectPrefix(): string {
```
samutoljamo: This is a bit fragile as it depends on store being pre-populated. Also in some context there can be subtle bugs: Sometimes we do a request and then mutate certain apiPaths to invalidate them. Since times goes between the start of the request and the mutation, we might end up mutating the wrong project.

 We don't really need to access project directly in non-hook contexts. I would refactor it so that we have a `useProjectPrefix` hook that can be used and apipaths require project prefix as a param.

tools/backend/src/main.ts
```diff
@@ -55,12 +56,47 @@ function createAuthProvider(): AuthProvider {
...
 } else {
+  if (projects.length === 0) {
+    console.error(
+      `No projects found in "${projectPath}". Cannot start the server without at least one project.`,
+    );
+    process.exit(1);
+  }
+  const registry = new ProjectRegistry();
+  for (const project of projects) {
+    const commands = new CommandManager(project.path);
```
samutoljamo, line 96: I'd consider try catch so a bad project doesn't kill the server

samutoljamo (approved):
Overall looks good, left a few comments. I'd add tests for the new functionality on the frontend/backend side. Also, I would consider moving project-scanner to backend. Or then implementing the multi-project on DH side properly.

tools/data-handler/src/project-scanner.ts
CyberJML: Tried to remove cardRoot and then to validate the project, and got "No 'cardRoot' in the current folder" + it is automatically created to the empty project. I would keep it here

tools/data-handler/src/project-scanner.ts
CyberJML: Yes I think it supports symlinks also, not sure what would be the harmful way to use them, but I guess it's just better to prevent them for now

tools/backend/src/app.ts
CyberJML: I think this is a placeholder anyways until the export is properly fixed on the next feature set. Not sure how it will turn out to be, so i'll add a todo comment there so it will be taken care of

tools/backend/src/app.ts
CyberJML: registry.first() is a commandManager instance, which would result in registry.first()?.project.configuration.cardKeyPrefix to get the same prefix

PR #1308, opened 2026-04-20
INTDEV-1090/1089: module versioning, check-updates, show modules

samutoljamo:
## Summary

  Implements **INTDEV-1090** (`cyberismo check-updates`) and the module-related parts of **INTDEV-1089** (`show modules`, `show module <name>`, enhanced `show project`), plus a ground-up rewrite of the internal      
  module system with semver-based version pinning.

## Note about allium
I left the allium spec, it's not 100% up to date, but close. It definitely did help implementing this feature and deserves a thought. I will remove it though before merging, because it's very useless if it's not up to date and we don't commit to it elsewhere too. Note that this PR was indeed mostly done by opus, but first it did go through extensive opus review and then finally I verified everything.
                                                                                                                                                                                                                        
  ### New CLI surface                                    
  - `cyberismo check-updates [moduleName]` — queries each installed module's remote source, prints a per-module table of `installed → latest` (with constraint/blocked/unreachable annotations), then interactively
  prompts to apply auto-updatable updates. Updates blocked by a version constraint in `cardsConfig.json` are reported but not offered.                                                                                  
  - `cyberismo show modules` — lists installed modules with their versions.
  - `cyberismo show module <prefix>` — full module content details.                                                                                                                                                     
  - `cyberismo show project` — now includes installed modules (+ versions) alongside existing project metadata.                                                                                                         
  - `cyberismo import module <source>@<version>` — `@<version>` suffix pins a semver version or range (e.g. `my-module@^1.0.0`). SSH-style `git@host:org/repo` URLs are parsed correctly.                               
                                                                                                                                                                                                                        
  ### Module system rewrite                                                                                                                                                                                             
  `tools/data-handler/src/module-manager.ts` (~719 lines) is deleted and replaced by a layered `tools/data-handler/src/modules/`:                                                                                       
  - `types.ts`, `version.ts`, `location.ts`, `credentials.ts` — core types and helpers (semver, URL/location parsing, git credentials).                                                                                 
  - `source.ts` + `source-file.ts` / `source-git.ts` / `source-composite.ts` — per-protocol source layers (local file, git HTTPS/SSH, composite routing).                                                               
  - `inventory.ts` — declared vs. installed module discovery.                                                                                                                                                           
  - `resolver.ts` — transitive resolution + diamond version-conflict detection.                                                                                                                                         
  - `installer.ts` — staged install, `cardsConfig.json` integration.                                                                                                                                                    
  - `orphans.ts` — cascade cleanup when a top-level module is removed.                                                                                                                                                  
  - `module-system.allium` — architecture spec.                                                                                                                                                                         
                                                                                                                                                                                                                        
  `Remove.removeModule` and `Import` are rewired onto the new layers; `ModuleManager` is gone. I think this was required, because ModuleManager was becoming this god class for modules and otherwise these 6k lines would've been there.                                                                                                                   
                                                                                                                                                                                                                        
  ### Test coverage                                                                                                                                                                                                     
  - New per-layer unit tests under `tools/data-handler/test/modules/` (9 files: `version`, `credentials`, `source-file`, `source-git`, `source-composite`, `inventory`, `resolver`, `installer`, `orphans`).
  - New integration tests: `command-check-updates.test.ts`, `import-live-git.test.ts`, expanded `command-import.test.ts` and `command-remove.test.ts`.                                                                  
  - Shared fixture helpers extracted to `test/helpers/module-fixtures.ts`.                                                                                                                                              
                                                                                                                                                                                                                        
  ## Breaking changes                                                                                                                                                                                                   
                                                                                                                                                                                                                        
  `cardsConfig.json` module schema (`tools/assets/src/schema/cardsConfigSchema.json`):                                                                                                                                  
  - `modules[].branch` **removed**. Replace with `modules[].version` (semver version or range, e.g. `"1.0.0"`, `"^1.0.0"`, `">=1.2.0 <2.0.0"`). Omit to accept any version.
  - `modules[].name` now validated as `^[a-z]+$`, length 3–10.                                                                                                                                                          
                                                                                                                                                                                                                        
  `import module` CLI:                                                                                                                                                                                                  
  - The positional `[branch]` argument is removed. Pass `source@version` instead (`my-module@main` or `my-module@^1.0.0`).                                                                                              
                                                                                                                                                                                                                        
  ## Scope note
                                                                                                                                                                                                                        
  INTDEV-1089 also lists `cyberismo show migration current` and `cyberismo show migration <migration_to>`. Those are **not** part of this PR — only the module-related `show` commands landed here. Migration-log       
  commands will be a follow-up.


tools/data-handler/src/commands/remove.ts
```diff
@@ -315,9 +313,33 @@ export class Remove {
...
   }
+
+  /**
+   * Remove a top-level module declaration and cascade orphan cleanup to a
+   * fixed point per the spec's `RemoveModule` rule. Transitive-only
+   * modules (no top-level declaration) cannot be removed directly — their
+   * lifetime is controlled by the parent installation.
+   */
+  private async removeModule(targetName: string) {
+    const inventory = createInventory();
```
samutoljamo: A bit ugly yes. Since the inventory is not cached in anyway, it is quite cheap. Would be nice to start considering whether it would be worth to create a way of injecting these kind of dependencies. So far, we've delayed the problem by having the project, which acts as kind of a godclass. This makes it easy to pass deps, but it also makes code less readable

tools/assets/src/schema/cardsConfigSchema.json
```diff
@@ -53,19 +53,22 @@
         "properties": {
           "name": {
             "description": "Module name (project prefix). Must be unique.",
-            "type": "string"
+            "type": "string",
+            "pattern": "^[a-z]+$",
```
samutoljamo: This is same as project prefix

tools/cli/src/index.ts
```diff
@@ -917,25 +936,26 @@ importCmd
...
       useCredentials: boolean,
       options: CommandOptions<'import'>,
     ) => {
-      let resolvedSource = source;
-      let resolvedBranch = branch;
+      const { source: parsedSource, version } = source
+        ? parseSourceAtVersion(source)
+        : { source: undefined, version: undefined };
+
+      let resolvedSource = parsedSource;
```
samutoljamo, line 955: Reminder to check whethere this is needed

tools/data-handler/src/commands/check-updates.ts
```diff
@@ -0,0 +1,221 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2025
```
samutoljamo: year

tools/data-handler/src/commands/remove.ts
```diff
@@ -315,9 +313,33 @@ export class Remove {
...
+    const declaration = inventory
+      .declared(this.project)
+      .find((d) => d.name === targetName);
+    if (!declaration) {
+      throw new Error(`Module '${targetName}' is not part of the project`);
+    }
+
+    // Delete the top-level declaration (removes it from cardsConfig.json
+    // and invalidates the project's module cache).
+    await this.project.removeModule(targetName);
```
samutoljamo: Since import module was removed from project, I will remove this method too

tools/data-handler/src/modules/types.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Transitive resolution diagnostics
+// ---------------------------------------------------------------------------
+
+/**
+ * Emitted by the resolver when a later transitive declaration's range rejects
+ * a version already chosen for the same name. First resolution wins; the
+ * conflict surfaces as a structured event instead of aborting.
```
samutoljamo, line 136: I think this behaviour is fine for now

tools/data-handler/src/modules/installer.ts
```diff
@@ -0,0 +1,331 @@
...
+      const entry = stage.resolved;
+      if (entry.declaration.parent !== undefined) {
+        continue;
+      }
+      await project.configuration.upsertModule(this.toPersistedSetting(entry));
+    }
+
+    // Clean up staging for cleanly-applied modules. Failed stages are
+    // intentionally left on disk so the offending tree can be inspected
+    // after the run — see module-system.allium's "Should temp staging
```
samutoljamo: Reminder to cleanup comment

samutoljamo:
Feel free to review whenever, I will still fix the comments I've given, but otherwise it looks good to me

tools/data-handler/src/commands/check-updates.ts
```diff
@@ -0,0 +1,221 @@
...
+   * @param moduleName Optional module name to check. If omitted, checks all.
+   * @param credentials Optional credentials for private modules.
+   * @returns Array of update status for each checked module.
+   */
+  @read
+  public async checkUpdates(
+    moduleName?: string,
+    credentials?: Credentials,
+  ): Promise<ModuleUpdateStatus[]> {
+    const inventory = createInventory();
```
CyberJML: Why not just new Inventory()? Or fix the comments/descriptions for this so I can easily see the purpose of this

samutoljamo: No reason here, forgot to remove these as the abstractions where simplified so this and `createResolver` ended up as wrappers

tools/data-handler/test/command-check-updates.test.ts
CyberJML, line 1: Tests that could be added
Multiple modules checked at once — every test has exactly one module
Filter by name (happy path) — passing a valid module name to only check that one
Reachable remote with zero version tags — remote works but has no tags at all
No version range declared — module with no version constraint
Declared but not installed — module in config but not on disk
Private module success — credentials forwarded correctly for a working private repo

tools/data-handler/src/modules/resolver.ts
```diff
@@ -0,0 +1,257 @@
...
+ * Read a fetched module's `cardsConfig.json`. Exported so callers can
+ * read `cardKeyPrefix` before handing a pre-fetched module to the resolver.
+ *
+ * @throws if the file does not exist or is empty.
+ */
+export async function readModuleConfig(
+  path: string,
+): Promise<ProjectConfiguration> {
+  const configPath = new ProjectPaths(path).configurationFile;
+  const config = (await readJsonFile(configPath)) as
```
CyberJML: const config = new ProjectConfiguration(configPath, false); ?

tools/data-handler/src/modules/resolver.ts
```diff
@@ -0,0 +1,257 @@
...
+        `Conflicting source for module '${decl.name}': ` +
+          `installed from '${existing.declaration.source.location}' ` +
+          `(private=${existingPrivate}), but also declared with ` +
+          `'${decl.source.location}' (private=${declPrivate})`,
+      );
+    }
+  }
+
+  private versionSatisfies(version: Version, range: VersionRange): boolean {
+    return satisfies(version, range);
```
CyberJML: :D

tools/data-handler/src/modules/resolver.ts
samutoljamo: Would require refactoring that class

tools/data-handler/src/modules/resolver.ts
samutoljamo: It was only for the types but now it's kinda useless, since it has a single caller

tools/data-handler/src/modules/remote-url.ts
CyberJML, line 1: Should this file be renamed to remote-url.ts or similar?

tools/data-handler/src/modules/installer.ts
```diff
@@ -0,0 +1,331 @@
...
+  ModuleSetting,
+} from '../interfaces/project-interfaces.js';
+import type { Project } from '../containers/project.js';
+
+/**
+ * Options consumed by {@link Installer.install}.
+ */
+export interface InstallOptions {
+  /** Credentials for private HTTPS remotes. */
+  credentials?: Credentials;
```
CyberJML: Never used?

tools/data-handler/src/modules/installer.ts
```diff
@@ -0,0 +1,331 @@
...
+    }
+    const stagedRoot = pathResolve(join(tempDir, stage.name));
+    return stage.resourcesFolder.startsWith(stagedRoot);
+  }
+
+  private async cleanupStage(
+    stage: StagedModule,
+    tempDir: string,
+  ): Promise<void> {
+    const stagedRoot = join(tempDir, stage.name);
```
CyberJML: Should this also use pathResolve like line 301?

tools/data-handler/src/modules/installer.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  /**
+   * Copy a staged module into `.cards/modules/<name>/`, replacing any
+   * existing installation.
+   */
+  private async applyOne(project: Project, stage: StagedModule): Promise<void> {
+    const destinationPath = join(project.paths.modulesFolder, stage.name);
+    await deleteDir(destinationPath);
+    await copyDir(stage.resourcesFolder, destinationPath);
```
CyberJML: Small possibility that stage.resourcesFolder does not actually exists? It's created with new ProjectPaths, but not verified that it actually exists?

CyberJML: Could this process also be improved by copying first the content to temp directory at destination path /something.installing, then delete, then rename?

tools/data-handler/src/modules/inventory.ts
```diff
@@ -0,0 +1,218 @@
...
+    project: Project,
+    name: string,
+    declaredSetting: ModuleSetting | undefined,
+  ): Promise<ModuleInstallation | undefined> {
+    const path = join(project.paths.modulesFolder, name);
+    const configPath = project.paths.moduleConfigurationFile(name);
+
+    let config: InstallationConfig | undefined;
+    try {
+      config = await readJsonFile(configPath);
```
CyberJML: Should we verify the configuration before using it?

tools/data-handler/src/modules/orphans.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  // Default cap: initial count + 2 is enough for any finite graph.
+  const initialInstalled = await inventory.installed(project);
+  const maxIterations = options.maxIterations ?? initialInstalled.length + 2;
+
+  let removed = 0;
+  let currentInstalled = initialInstalled;
+
+  for (let iteration = 1; iteration <= maxIterations; iteration++) {
+    const declared = inventory.declared(project);
```
CyberJML: can be moved outside of the for-loop?

tools/data-handler/src/modules/resolver.ts
```diff
@@ -0,0 +1,257 @@
...
+                `'${decl.name || decl.source.location}'`,
+            );
+          }
+          version = picked;
+          ref = versionToTag(version);
+        }
+        // Empty list → source doesn't support versioning; leave
+        // version/ref undefined (silently ignore the range).
+      }
+      // No override, no range → leave version/ref undefined (default branch for git).
```
CyberJML: Did I understand this correctly, if there's no versions on remote, but the user has defined v2.0.0 to be installed, this will silently just install the default branch from git and leave the version undefined? Should there be some notification to the user that this happened and the version might not be the correct one?

tools/data-handler/src/modules/resolver.ts
```diff
@@ -0,0 +1,257 @@
...
+        { location: decl.source.location, remoteUrl, ref },
+        options.tempDir,
+        decl.name,
+      );
+      const childConfig = await readModuleConfig(path);
+
+      const resolvedEntry: ResolvedModule = {
+        declaration: { ...decl },
+        ref,
+        remoteUrl,
```
CyberJML, line 147: Care that this remoteUrl does not endup in logs

tools/data-handler/src/utils/git-manager.ts
```diff
@@ -130,4 +130,28 @@ export class GitManager {
...
   }
+
+  /**
+   * List available version tags from a remote repository.
+   * Does not require a local repo — queries the remote directly.
+   * @param remoteUrl Git remote URL to query
+   * @returns Semver version strings sorted descending (e.g. ["2.1.0", "1.0.0"])
+   */
+  static async listRemoteVersionTags(remoteUrl: string): Promise<string[]> {
+    const git = simpleGit();
```
CyberJML: const git: SimpleGit = simpleGit({
      timeout: { block: gitTimeout() },
    });
?

tools/data-handler/src/modules/version.ts
```diff
@@ -0,0 +1,90 @@
...
+/** Prefix used on git tags to mark a semver release, e.g. `v1.2.3`. */
+export const TAG_PREFIX = 'v';
+
+/** Convert a semver version to a git tag name (prefixes with `v`). */
+export function versionToTag(version: Version | string): string {
+  return `${TAG_PREFIX}${version}`;
+}
+
+/** Strip the tag prefix; returns input unchanged if the prefix is absent. */
+export function tagToVersion(tag: string): string {
```
CyberJML: stripTagPrefix() would be better name?

tools/data-handler/src/modules/version.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import { toVersion, type Version, type VersionRange } from './types.js';
+
+/** Pure semver helpers used by the module system. */
+
+/** Prefix used on git tags to mark a semver release, e.g. `v1.2.3`. */
+export const TAG_PREFIX = 'v';
+
+/** Convert a semver version to a git tag name (prefixes with `v`). */
+export function versionToTag(version: Version | string): string {
```
CyberJML, line 25: prefixTagVersion() would be better name?

CyberJML (approved):
Fix the pipeline errors and check the existing comments

tools/data-handler/src/modules/installer.ts
samutoljamo: yup

tools/data-handler/src/modules/remote-url.ts
samutoljamo: Yes

tools/data-handler/src/modules/installer.ts
samutoljamo: No need, pathResolve is needed just for the comparison

tools/data-handler/src/modules/installer.ts
samutoljamo: Generally speaking, we should not check for existance before an operation as it's a race condition(TOCTOU). Currently the git modules are in project temp dir while local file imports are in wherever it was before. Maybe we could indeed put the files to the project temp dir always and rename from there.

tools/data-handler/src/modules/inventory.ts
samutoljamo: Yes, I think we could for now have a single method for reading the config and I refactor so that ProjectSettings also uses it. ProjectSettings' role will be the mutator.

tools/data-handler/src/modules/resolver.ts
samutoljamo: Not logged by us, but we log git output. Git should redact the username and password from the url and I tried it seems that is the case.

tools/data-handler/src/modules/resolver.ts
samutoljamo: Yeah it's bad for git, required for file. I'll extend source layer to have a "supportsVersioning". If yes ---> this is an error and if no --> leave version undefined

tools/data-handler/src/modules/version.ts
samutoljamo: I like versionToTag here, since the version type is branded.

tools/data-handler/src/modules/version.ts
samutoljamo: Yup

PR #1311, opened 2026-04-22
INTDEV-1125 Metadata & linked cards UI updates & inline editing

tminkkinen:
Includes Metadata & Linked items UX changes discussed on Monday.

Does not include save on blur as it turned out to be bit more complicated due to how MUI renders some of the inputs (using portal). Figured out it could be added later if needed.


tools/app/src/components/LinkedCardsSection/EditMode.tsx
```diff
@@ -0,0 +1,521 @@
...
+  Stack,
+} from '@mui/joy';
+import Add from '@mui/icons-material/Add';
+import Check from '@mui/icons-material/Check';
+import Close from '@mui/icons-material/Close';
+import Delete from '@mui/icons-material/Delete';
+import Edit from '@mui/icons-material/Edit';
+import Search from '@mui/icons-material/Search';
+import { Controller, useForm, useWatch } from 'react-hook-form';
+import type { ExpandedLinkType } from '../../lib/definitions';
```
samutoljamo: Prefer @/lib for all imports

tools/backend/test/api.test.ts
```diff
@@ -424,3 +424,100 @@ test('PUT /api/cards/:key/links returns 400 when required fields are missing', a
...
+    body: JSON.stringify({
+      metadata: { title: 'Updated Title' },
+    }),
+  });
+  expect(response.status).toBe(200);
+  const result = (await response.json()) as CardApiResponse;
+  expect(result.title).toBe('Updated Title');
+
+  // Restore original title
+  await app.request('/api/cards/decision_5', {
```
samutoljamo, line 445: A bit risky restore habit as the cleanup is part of the test. If the test fails, this might cause the other tests to fail. But since existing tests also use this pattern, it's not really a concern of this PR.

tools/app/src/components/MetadataSection/FieldRow.tsx
```diff
@@ -0,0 +1,228 @@
...
+              color="neutral"
+              onClick={handleCancel}
+            >
+              <CloseIcon />
+            </IconButton>
+          </Stack>
+        ) : (
+          <Box
+            onClick={isClickable ? onStartEdit : undefined}
+            data-cy={isClickable ? 'editableFieldRow' : undefined}
```
samutoljamo: I think it's fine that `data-cy` is set even if `isClickable` is false

tools/app/src/components/MetadataSection/FieldRow.tsx
```diff
@@ -0,0 +1,228 @@
...
+  dataType: DataType | 'label';
+  description?: string;
+  enumValues?: EnumDefinition[];
+  canEdit?: boolean;
+  isEditing?: boolean;
+  disabled?: boolean;
+  forceReadOnly?: boolean;
+  onStartEdit?: () => void;
+  onSave?: (value: MetadataValue) => void;
+  onAutoSave?: (value: MetadataValue) => void;
```
samutoljamo, line 36: Why we need both onAutoSave and onSave?

tools/app/src/components/MetadataSection/FieldRow.tsx
```diff
@@ -0,0 +1,228 @@
...
+    formState: { isDirty },
+  } = useForm({ defaultValues: { value: initialValue } });
+
+  const serializedInitial = JSON.stringify(initialValue);
+  useEffect(() => {
+    reset({ value: initialValue });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [serializedInitial, reset, isEditing]);
+
+  const handleChange = useCallback(
```
samutoljamo: I would not use `useCallback` here

tools/app/src/components/MetadataSection/FieldRow.tsx
```diff
@@ -0,0 +1,228 @@
...
+export interface FieldRowProps {
+  expanded?: boolean;
+  value: MetadataValue | null | undefined;
+  label: string;
+  dataType: DataType | 'label';
+  description?: string;
+  enumValues?: EnumDefinition[];
+  canEdit?: boolean;
+  isEditing?: boolean;
+  disabled?: boolean;
```
samutoljamo, line 33: We have quite many onEdit, disabled, forceReadOnly which seem to resolve to the same mode. I wonder if we could just have a single boolean for controlling this behaviour. Maybe just `disabled`?

tools/app/src/lib/slices/pageState.ts
```diff
@@ -22,12 +22,14 @@ export interface PageState {
...
   title: string | null; // title of the section being viewed as an asciidoc id
   cardKey: string | null; // cardKey of the section being edited. Mainly used to make sure we don't scroll on other cards
+  linkedCardsExpanded: boolean;
 }
 
 export const initialState: PageState = {
   isEdited: false,
   title: null,
   cardKey: null,
+  linkedCardsExpanded: false,
```
samutoljamo, line 32: This is blacklisted so it won't persist over refreshes, but maybe this is fine

tools/app/src/components/MetadataSection/MetadataSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+          canEdit={canEdit}
+          isEditing={editingFieldKey === 'labels'}
+          description={
+            editingFieldKey === 'labels'
+              ? t('labelEditor.splitterHint', { splitter: LABEL_SPLITTER })
+              : undefined
+          }
+          onStartEdit={() => setEditingFieldKey('labels')}
+          onAutoSave={(val) =>
+            onUpdate!({ metadata: { labels: val } })
```
samutoljamo: I think because of the full array update labels(and other similar fields) have a few race conditions, but I think that risk can be taken here or we could fix this later. Though I feel that this should be done on the backend than trying to mitigate this on the forntend

tools/app/src/components/MetadataSection/MetadataSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
+    setEditingFieldKey(null);
+  }, [card.key]);
+
+  const canEdit = !getConfig().staticMode && !!onUpdate;
+
+  const handleSaveField = async (metadataKey: string, value: MetadataValue) => {
+    await onUpdate!({ metadata: { [metadataKey]: value } });
```
samutoljamo: Could use `onUpdate?.` or tighten the type for `onUpdate`

tools/app/src/components/MetadataSection/MetadataSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+    setEditingFieldKey(null);
```
samutoljamo: Currently onMetadataUpdate(that is onUpdate here) swallows the error so this will be always run regardless of whether the update fails or is successful. Is that intentional?

samutoljamo (approved):
Left some comments. Some other open question:
* isEdited is not dispatched for the inline edits. I think that's fine as it makes it content-only(isEdited makes app show a warning when leaving the page or closing the browser)
* What do you think about component testing? Early on I personally didn't really do them too much as the project was living quite much and focused on testing any utilities / api + we have some e2e tests on top. I feel like now things are more stable and AI is much better, so maybe we could form the habit of doing component testing too?

samutoljamo:
One more thing, I think this will break `goToField`. Basically the policy errors can reference a field that causes it. This will show a "goToField" link. I guess it should now focus the field, no?

tools/backend/test/api.test.ts
tminkkinen: True, it's bit risky... Maybe tests could be updated in some point and cleanup improved generally

tools/app/src/lib/slices/pageState.ts
tminkkinen: Based on mondays discussion I believe it's fine if this state doesn't persist over refreshes

tminkkinen:
> Left some comments. Some other open question:
> 
> * isEdited is not dispatched for the inline edits. I think that's fine as it makes it content-only(isEdited makes app show a warning when leaving the page or closing the browser)
> * What do you think about component testing? Early on I personally didn't really do them too much as the project was living quite much and focused on testing any utilities / api + we have some e2e tests on top. I feel like now things are more stable and AI is much better, so maybe we could form the habit of doing component testing too?

I think component testing could benefit with some key components which includes logic / handles interactions. Now that project is getting more stable it could make sense to start adding those also.

PR #1312, opened 2026-04-23
INTDEV-1097 frontend role-based permissions

samutoljamo:
Add auth primitives (UserRole enum, parseRole, roleSatisfies, useHasRole, Gate) and gate card/config UI by role.

Static mode resolves to a synthetic reader user so a single role check drives gating everywhere. Readers and editors can view /configuration; mutations stay admin-only.

Adds a dev-only mock-role cookie middleware so roles can be exercised locally without a restart.

### DIFF
--- a/tools/app/src/lib/auth/roles.ts
+++ b/tools/app/src/lib/auth/roles.ts
@@ -11,26 +11,26 @@
   License along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { useSWRHook } from './common';
-import { apiPaths } from '../swr';
-import { getConfig } from '../utils';
-import type { User } from './types';
+export enum UserRole {
+  Reader = 0,
+  Editor = 1,
+  Admin = 2,
+}
 
-import type { SWRConfiguration } from 'swr';
-
-const STATIC_READER_USER: User = {
-  id: 'static-reader',
-  email: '',
-  name: '',
-  role: 'reader',
+const ROLE_MAP: Record<string, UserRole> = {
+  reader: UserRole.Reader,
+  editor: UserRole.Editor,
+  admin: UserRole.Admin,
 };
 
-export const useUser = (options?: SWRConfiguration) => {
-  const staticMode = getConfig().staticMode;
-  return useSWRHook<'user', User | null>(
-    staticMode ? null : apiPaths.user(),
-    'user',
-    staticMode ? STATIC_READER_USER : null,
-    options,
-  );
-};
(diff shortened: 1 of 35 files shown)


tools/app/src/lib/auth/usePermissions.ts
```diff
@@ -0,0 +1,24 @@
...
+import { UserRole, parseRole, roleSatisfies } from './roles';
+
+export function useHasRole(minRole: UserRole): boolean {
+  const { user } = useUser();
+  if (!user) return false;
+  return roleSatisfies(parseRole(user.role), minRole);
+}
+
+export const useCanEdit = () => useHasRole(UserRole.Editor);
+export const useCanAdmin = () => useHasRole(UserRole.Admin);
```
tminkkinen: Names of `useHasRole` and `useCanAdmin` feels bit off. `useHasRole` reads like exact role check, but in reality it's more like "role satisfies". Can admin on the other hand feels to be tied to the role, not the action (like can edit). 

Not sure if this is real problem and how these should be named, but in the `useCanAdmin` case something along the lines `useCanManage` sounds a bit more natural to me atleast

tools/app/src/lib/auth/roles.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!raw) return null;
+  return ROLE_MAP[raw.toLowerCase()] ?? null;
+}
+
+export function roleSatisfies(
+  userRole: UserRole | null,
+  minRole: UserRole,
+): boolean {
+  return userRole != null && userRole >= minRole;
+}
```
tminkkinen, line 36: Is it likely that there will be more roles in future? If not, current numeric enum approach is probably fine.

CyberJML (approved):

tools/app/src/lib/auth/roles.ts
samutoljamo: For now, yes

tools/app/src/lib/auth/usePermissions.ts
samutoljamo: I'll rename `useCanAdmin` to `useIsAdmin`. Then I could use `useHasMinRole` for the role check. I think for the simple authorization we have, this will work fine, and when we'd have more complex authorization, we'll need to refactor anyway

samutoljamo: Actually, I could remove `useX` hooks, they are very simple wrappers

PR #1316, opened 2026-04-27
INTDEV-1179 Render workflow state-machine diagram in workflow editor

samutoljamo:
## Changes
  - New built-in Clingo asset (`workflowGraph/{model.lp,view.lp.hbs}`) that renders the workflow's states and transitions as a Graphviz state machine.
  - `Calculate.runWorkflowGraph(workflowName)` produces a sanitized, base64-encoded SVG.
  - New endpoint `GET /api/resources/:prefix/workflows/:identifier/graph` returning `{ svg }`.
  - New `WorkflowGraph` field component wired into `ResourceEditor` for workflow resources, fetched via SWR.

### DIFF
--- a/tools/app/src/components/config-editors/fields/WorkflowGraph.tsx
+++ b/tools/app/src/components/config-editors/fields/WorkflowGraph.tsx
@@ -1,0 +1,92 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation.
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { useState } from 'react';
+import {
+  Box,
+  CircularProgress,
+  IconButton,
+  Tooltip,
+  Typography,
+} from '@mui/joy';
+import Fullscreen from '@mui/icons-material/Fullscreen';
+import { useTranslation } from 'react-i18next';
+import { useWorkflowGraph } from '@/lib/api';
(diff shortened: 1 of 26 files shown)


samutoljamo:
I feel like we should stop rendering these svgs on the backend and just send the dot graph, but that's for another story

tools/data-handler/src/commands/calculate.ts
```diff
@@ -11,11 +11,21 @@
...
+import { workflowGraph } from '@cyberismo/assets';
+
 import type { Context } from '../interfaces/project-interfaces.js';
 import type { Project } from '../containers/project.js';
 import type { QueryName, QueryResult } from '../types/queries.js';
+import { registerClingoHelpers } from '../utils/handlebars-helpers.js';
 import { read } from '../utils/rw-lock.js';
 
+const workflowGraphHandlebars = Handlebars.create();
+registerClingoHelpers(workflowGraphHandlebars);
```
samutoljamo, line 24: Rather than handling these one by one, I think it would be better to "formalize" the concept of the content langauge so that we wouldn't need to use handlebars / clingo directly.

tools/app/src/components/config-editors/fields/WorkflowGraph.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <Box>
+      <Typography level="h4" marginTop={6} marginBottom={4}>
+        {t('workflowGraph.title')}
+      </Typography>
+      {body}
+    </Box>
+  );
+}
+
+export default WorkflowGraph;
```
tminkkinen, line 103: Is there reason to have both named & default exports in components? This seems to be quite common pattern in the codebase

tminkkinen (commented):
lgtm but maybe it's better if someone else checks the clingo part

tools/app/src/components/config-editors/fields/WorkflowGraph.tsx
samutoljamo: Good catch, I feel like for components we could use default exports and keep exports for things like prop definitions etc. that might need to be exported alongisde with the component. Or do you have a preference?

henryhaverinen (approved):
Looks great!

- instead of the icon, there should be an option in the workflow menu "View workflow". It should be the last item and separated by a separator 
- please make the diagram clickable in the configuration editor, to open the full screen viewer

CyberJML (approved):

PR #1317, opened 2026-04-27
INTDEV-1125 Card content inline editing

tminkkinen:


CyberJML:
image
Since card keys are most likely quite regular lengths could we align those splitter lines? There's a case of external links that could be longer, not sure what to do with those?

CyberJML:
Another thing is the info boxes allow selection of text, not a problem, but by clicking them it will start editing + you cannot scroll down the infoboxes and read them through since the infobox will go over the next one:
image

Maybe the solution is to show infobox only if you are on the icon allowing the easy scroll through the infoboxes

CyberJML:
Also some of the metadata is not editable, but I did not notice clear indication of which are editable and whatnot. Should there be some visual cue, or possibly groupping based on whether they are "calculated" metadata or editable?

CyberJML:
The clickable area stops before the end of metadata box 
image
not a major problem, but more of a visual problem as seen in the image

CyberJML:
Link edit mode does not give indication why I cannot modify these links 
image

I believe those are calculated links, but user does not get any indication of that and are wondering why cannot I remove/edit this one

CyberJML:
image
Plus sign is coming over the edge on half screen window

CyberJML:
> image Plus sign is coming over the edge on half screen window

image

:D

CyberJML:
Not sure if content edit should start if the content are is clicked, even though we have prevention against it when selecting text, but accidental click to the content are will cause unnecessary jumping to the page.

Maybe the solution would be to start the edit mode only from the edit button, but it would be fixed on top if scrolling down the content view?

CyberJML:
And the toolbar + other buttons might need some modification to be usable in half window mode 
image

CyberJML:
image
Linked items box might be a bit too large (height wise), and maybe the padding between the boxes are also a bit too much.

This might be on other boxes too, too much padding between the boxes + too much padding/margin within the box, making them too big (at least header and linked items)

CyberJML:
would it make sense to have the whole linked items box be clickable when collapsed? 
image

Not sure about this, but tried to click on the linked items and it did nothing, maybe the header line could be clickable and it would open the box?

CyberJML (approved):
Feels nice

PR #1343, opened 2026-05-04
Enable multiproject MCP

CyberJML:


tools/backend/src/domain/mcp/index.ts
```diff
@@ -71,89 +69,92 @@ const cleanupInterval = setInterval(() => {
...
 
-/**
- * SSE endpoint for server-to-client messages
- */
-router.get('/sse', async (c) => {
-  const sessionId = c.req.header('mcp-session-id');
+  /**
+   * SSE endpoint for server-to-client messages
+   */
+  router.get('/sse', async (c) => {
```
samutoljamo: I think we can remove this route, since SSE is deprecated

tools/mcp/src/resources/index.ts
```diff
@@ -14,11 +14,38 @@
...
 import type { ResourceTypeConfig } from '../lib/mcp-helpers.js';
+import {
+  resolveCommands,
+  type ProjectProvider,
+} from '../lib/resolve-project.js';
 
 import { getCardTree } from '../lib/render.js';
 
+/**
+ * Wrap a resource template callback so that thrown errors become McpErrors
```
samutoljamo: Sounds very weird that they crash the transport. At a quick glance docs don't seem to suggest this is the case

tools/backend/src/domain/mcp/index.ts
```diff
@@ -71,89 +69,92 @@ const cleanupInterval = setInterval(() => {
...
-    if (sessionId) {
-      await destroySession(sessionId);
+export function createMcpRouter(provider: ProjectProvider): Hono {
+  const router = new Hono();
+
+  /**
+   * MCP HTTP endpoint handler.
+   * Supports GET (SSE streaming), POST (messages), and DELETE (session cleanup).
+   */
+  router.all('/', async (c) => {
```
samutoljamo: It seems to me that MCP does not have any role checks

samutoljamo: Also I think we could scope the sessions per user. Right now it's possible to use/delete other sessions with a valid auth.

tools/backend/src/domain/mcp/index.ts
```diff
@@ -71,89 +69,92 @@ const cleanupInterval = setInterval(() => {
...
+        { error: 'Method not allowed. Use POST to initialize a session.' },
+        405,
+      );
     }
-  };
 
-  const server = createMcpServer(commands);
-  await server.connect(transport);
+    // Reject new sessions when at capacity
+    if (sessions.size >= MAX_SESSIONS) {
```
samutoljamo, line 121: This has a race condition, but I think it's fine if we go a bit over the limit

tools/backend/src/domain/mcp/index.ts
```diff
@@ -71,89 +69,92 @@ const cleanupInterval = setInterval(() => {
...
-      void destroySession(closedSessionId, true);
-    },
-  });
+    // Handle existing session
+    if (sessionId && sessions.has(sessionId)) {
+      const session = sessions.get(sessionId)!;
+      session.lastActivity = Date.now();
+      const response = await session.transport.handleRequest(c.req.raw);
+      return response;
+    }
```
samutoljamo, line 105: We could reject a non-existing session I think. Ie. just for the scenario where backend(because of a restart since sessions are just memory) loses the session, I am not sure how mcp server reacts

samutoljamo (approved):
The stateful MCP seems already a bit complex and we don't use it in anyway, I'd consider removing it too. Otherwise it would nice if the limits were not global as a single user can lock out all other users from the MCP

tools/data-handler/src/interfaces/project-provider.ts
```diff
@@ -0,0 +1,25 @@
...
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import type { CommandManager } from '../command-manager.js';
+
+/**
+ * Minimal interface for project lookup by prefix.
+ * Implemented by ProjectRegistry in the backend and used by the MCP server.
+ */
+export interface ProjectProvider {
```
samutoljamo, line 22: I'd also add this to backend so that we keep multi-project as a backend thing

tools/backend/src/domain/mcp/index.ts
CyberJML: I had this situation when testing this feature, and it was able to recover from it via it's own logic after a while. But this could be more clear to it

PR #1345, opened 2026-05-05
INTDEV-1268 Improve multiproject view on gui

CyberJML:
Current flow is:
If user navigates to root path / and they have not visited the domain before they'll see 
image

If server has been set up with default project, this will not be shown, but the default project will be automatically opened.

Current design is following the Confluence spaces design, where we can see the projects and previous projects at top left corner like this

image

The recent projects list shows only 5 previous projects in visiting order (might not be good, since it's "jumping" when switching projects)

At the start the list is small (stored on the browser, it's user specific), but grows if another project is visited
image

More projects button is designed to contain the create new project, but for now it's not there, since we would first need to enable easy module import from GUI. Current More projects view is following create card view
image

But it shows a bit more information, Title, prefix, category, and description, like seen in testiprojekti box. Also the project with different border is the current project.

Static site also follows the same idea if the export site is done in a directory with multiple projects and no project directly at the root (I don't think we want to allow structure of project contains projects). We can use cyberismo export site something --default-project docs to force the root path nagivation to always end up in docs project. 
Static page view: 
image


tools/app/src/components/modals/ProjectSelectionModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+                <CircularProgress size="md" />
+              </Stack>
+            ) : filteredProjects.length === 0 ? (
+              <Typography level="body-xs" color="neutral">
+                {t('projectDialog.noProjectsFound')}
+              </Typography>
+            ) : (
+              <Grid
+                container
+                spacing={2}
```
tminkkinen: Should be gap? This doesn't seem to do anything in this case

tools/app/src/components/modals/ProjectSelectionModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+  onClose,
+  dismissable = true,
+}: ProjectSelectionModalProps) {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+  const { projectPrefix: currentPrefix } = useParams();
+  const { data: projects, isLoading } = useAvailableProjects();
+
+  const [filter, setFilter] = useState('');
+  const [selectedProject, setSelectedProject] = useState<string | null>(null);
```
tminkkinen: Should current project be selected by default? Highlighting makes it clear which one of the projects is current, but on the other hand it feels somewhat weird to me that it is not selected initially but can be selected. What do you think?

tools/app/src/components/SearchableTreeMenu.tsx
```diff
@@ -99,31 +121,102 @@ export const SearchableTreeMenu = ({
...
+          textTransform="uppercase"
+          fontWeight="lg"
+          px={2}
+          pt={2}
+          pb={0.5}
+        >
+          {t('projectDialog.recentProjects')}
+        </Typography>
+
+        <List size="sm" sx={{ px: 1, py: 0 }}>
```
tminkkinen, line 138: This starts to take quite some space from the card three menu with 5 items. Maybe we could consider if this could / should be collapsible in future.

tools/app/src/routes.ts
```diff
@@ -83,15 +89,18 @@ export function createAppRouter() {
...
     {
       path: '/projects/:projectPrefix',
       Component: Layout,
       loader: async ({ params }) => {
-        const { prefix, fromUrl } = await resolveProject(params.projectPrefix);
-        if (fromUrl) return null;
+        const { prefix } = await resolveProject(params.projectPrefix);
+        if (prefix === params.projectPrefix) return null;
         if (prefix) return redirect(`/projects/${prefix}/cards`);
         return redirect('/');
```
tminkkinen: This has been probably decided earlier, but I was wondering if redirect is best approach here with multiple projects or should it be just 404 if project does not exist?

tminkkinen (approved):
Looks good to me 👍 Some thoughts in the comments, but I don't think those require any changes (except the broken spacing).

tools/app/src/components/modals/ProjectSelectionModal.tsx
CyberJML: yes

tools/app/src/components/modals/ProjectSelectionModal.tsx
CyberJML: Makes sense

tools/app/src/components/SearchableTreeMenu.tsx
CyberJML: You might be right, but cannot make the decision right now. Maybe future UX or user feedback can guide us with this one. There are couple options, collapsible, reduce the amount of projects, or even other location for this view. Samu had a good idea where to put these, but I'm still holding the location here and want to hear team's opinion on these first

tools/app/src/routes.ts
CyberJML: Makes sense, Guess this was partially leftover idea from the original frontend only project view, where we didn't have any backend connections, but now it does not make sense anymore

CyberJML: Now current flow is: 
1. User adds some random project name
2. New ProjectNotFoundPage is shown to the user
3. Page has some basic jargon + Browse projects button
4. Page also removes last project information from user's redux
5. Button navigates user back to / where they will be presented the project selection list (ugly, but waiting for better eyes to make it better)

PR #1348, opened 2026-05-07
INTDEV-1236 Name parent module in errors for transitive-only modules …

tminkkinen:
…remove/update/check

### DIFF
--- a/tools/data-handler/src/commands/check-updates.ts
+++ b/tools/data-handler/src/commands/check-updates.ts
@@ -60,14 +60,23 @@ export class CheckUpdates {
       ? allDeclared.filter((d) => d.name === moduleName)
       : allDeclared;
 
-    if (moduleName && declared.length === 0) {
-      throw new Error(`Module '${moduleName}' is not part of the project`);
-    }
-
     const installed = await installedModules(this.project);
     const installedByName = new Map<string, ModuleInstallation>(
       installed.map((i) => [i.name, i]),
     );
+
+    if (moduleName && declared.length === 0) {
+      const parents = installed
+        .filter((m) => m.declaredDependencies.includes(moduleName))
+        .map((m) => m.name);
+      if (parents.length > 0) {
+        const parentList = parents.map((n) => `'${n}'`).join(', ');
+        throw new Error(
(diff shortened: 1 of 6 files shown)


tools/data-handler/src/commands/check-updates.ts
```diff
@@ -60,15 +60,24 @@ export class CheckUpdates {
...
-    }
-
     const installed = await installedModules(this.project);
     const installedByName = new Map<string, ModuleInstallation>(
       installed.map((i) => [i.name, i]),
     );
 
+    if (moduleName && declared.length === 0) {
+      const parents = installed
+        .filter((m) => m.declaredDependencies.includes(moduleName))
```
samutoljamo, line 70: This logic has been done 3 times, could it be a helper? Maybe `isTransient` or something like that

samutoljamo: Actually it cannot be `isTransient`, because this is not a check whether a module is transient, it checks whether it depends on any declared modules.

samutoljamo (commented):
We need to think about how we want to handle transient deps and half transient half project deps.

For fully direct deps, we want that they are visible and user can do anything to them.
For half project half transient deps, we want they are also visible. If they are visible, it's a bit confusing if u cannot do any operations to them. And sometimes you need to:
```
project p depnds on module A
project p depends on module B
module A depends on module B
```
^ Now with the current logic, it becomes impossible to remove module B as a project dependency. It also impossible to import modules used by other modules. Ie. if you import EU-cra and then try to import base module.

I think one solution for this is that non-direct deps are always hidden(but there may be some show transitive dependencies, but they wouldn't be shown on the UI). Now "remove" would not mean remove it from file system, it means remove it from the module configuration. And it is deleted from the project if there are no references to it(cleanOrphans already supports this)

samutoljamo (approved):
Ignore the earlier comment I understood the logic wrong there, but what is true is that we'd need to later think about whether we want to actually show transitive deps at all. Now the behaviour is a bit confusing if you are deleting a half transient half project dep:
On the UI nothing seems to happen as the module stays in the project and if you try to delete it again then u get this error. I think we'd need to consider hiding the transitive deps so it's not even possible to do any actions to them.

PR #1351, opened 2026-05-08
INTDEV-1212 Enable MCP authentication

CyberJML:


samilempinen (approved):

samutoljamo (approved):
Works, but could consider using @hono/mcp rather than implementing this and other stuff yourself.

PR #1355, opened 2026-05-12
INTDEV-1125 Inline editing for card metadata, links, and content

tminkkinen:
Replaces the separate `/cards/:key/edit` route with inline editing on the card
  view — title, metadata, labels, linked cards, and AsciiDoc content are now
  editable in place. `ContentArea.tsx` is split into focused components under
  `components/card/`, and `CardToolbar` drops its edit/cancel/update buttons in
  favor of the inline flow.


henryhaverinen:
@tminkkinen This is looking and feeling very nice! I'll give it a bit more testing, but there is this problem that I noticed: the field editor may sometimes overlap with the "Show more" button (and maybe show less as well?). See for example the first cards of the cyberismo-docs project:

image

henryhaverinen:
@tminkkinen  A minor editorial comment: the font size for "Linked items" seems a bit large to me. Should it be, for example, the same font size as we have for Table of contents?

image

samutoljamo (approved):
Assumption was that the code has been reviewed before, so fix the tests and LGTM

henryhaverinen:
@tminkkinen  thank you, the show more button works correctly now, and the links heading looks good, too. 

I was still wondering, if we should rather hide the Show more thing when editing a field instead of making the box taller. One of the reasons for the UX exercise is to make the top of the card lighter and smaller, so it would be good if the metadata box didn't take up unnecessary space

image

tminkkinen:
> @tminkkinen thank you, the show more button works correctly now, and the links heading looks good, too.
> 
> I was still wondering, if we should rather hide the Show more thing when editing a field instead of making the box taller. One of the reasons for the UX exercise is to make the top of the card lighter and smaller, so it would be good if the metadata box didn't take up unnecessary space
> 
> image

@henryhaverinen I pushed this forward before new conflicts will arise, but this could be one solution to that. Also one option could be to reserve space for show more / show less button when user enters editing, but that would cause some layout shift which generally isn't considered good.

PR #1356, opened 2026-05-12
INTDEV-1250 Allow template card moves

tminkkinen:


henryhaverinen (changes requested):
I tested this with module-secure-development-essentials. The Development project template is a good test case, because it is a complicated template.  I found the following issues:

- When trying to move Compliance as the first card under Design, I get an error "rank1 must be smaller than rank2". After refreshing, Compliance has moved as the last card under Design
- When trying to move some other card at the top level of the template, as the first card, I get the same error "rank1 must be smaller than rank2"
- In general it doesn't seem to be possible to rank anything as the first card
- When trying to move a card to the top level of the template as the last card, this fails with the error "Cards must be from the same parent"

henryhaverinen (approved):
I tested with module-secure-development-essentials, and everything seemed to work great for me

PR #1357, opened 2026-05-18
Simplify enum and list definitions

samutoljamo:
This gets rid of the complexity in regarding enum and list definitions in the query language. ResultField continues to be slightly awkward with enum and lists as they depend on other fields, such as the enum display name, index or value being generated. However, that is a different problem, the purpose of this is to improve the performance.


henryhaverinen (approved):
Please fix the comment as spotted by Copilot. Otherwise, this looks good to me. I tested with module-test and list fields and enum fields still work.

PR #1362, opened 2026-05-19
INTDEV-1284 Remove presence indicator from local environment

CyberJML:


tools/app/src/lib/api/presence.ts
```diff
@@ -65,19 +68,14 @@ export function usePresence(
...
 
     return () => {
-      window.removeEventListener('pagehide', handlePageHide);
       eventSource.close();
       setEditors([]);
     };
   }, [url]);
 
-  if (!cardKey || getConfig().staticMode) return [];
+  if (!url) return [];
```
samutoljamo: We can just remove this, `editors` is empty if theres no url

samutoljamo (approved):
Before merging: So that dockerimage allows controlling this via an env variable, add this variable to `entrypoint.sh`. There will likely be more configuration in the future too so might as well make a loop rather than specifying each configurable variable manually

PR #1367, opened 2026-05-22
INTDEV-923 Make static sites usable on mobile

tminkkinen:


samutoljamo:
With empty content:
image

samutoljamo:
image

When you have a graph as the first thing, the controls seem to overlap a bit. Additionally, you can't change between dark / light mode.

It may be a bit unclear to the user that the fields can be edited by clicking them. Especially when they don't have any values. Though this is mostly a design consideration, so I think there's no reason to get stuck with this here so if u dont do it here, would be good to document it.


image

The controls themselves are not mobile friendly, so either fix them here or in a seperate PR

samutoljamo (approved):
Code looks good to me. Content a bit on the edge whether is responsive or not, but heavy elements like tables are not trivial to handle responsively so I think that's fine for now. Though one think that caught my eye is that the sidebar is below content so TOC, policychecks etc require scrolling the whole content. I wonder if it would be better to actually hide TOC, since it would take a lot of space if it was before content, but at the end of the  content I'm not sure if it's useful. Anyway, I think these can be fixed in another PR if required, but they should be documented

PR #1378, opened 2026-05-25
INTDEV-1290 Prevent card content edit button overlap on narrow screens

tminkkinen:


samutoljamo (approved):

samutoljamo:
I started INTDEV-1298, but noticed this one fixes it too. However, on large screens text seems to overlap still
image

PR #1379, opened 2026-05-26
Update docs installation guide

samutoljamo:


setup-modules.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
```
henryhaverinen: Is this file included in the PR on purpose or by accident?

docs/cardRoot/docs_13/c/docs_17/index.adoc
```diff
@@ -1,24 +1,62 @@
...
 === Asciidoctor PDF (optional, for PDF export)
 
-Install Asciidoctor PDF, if you need to export content as PDF.
-
-NOTE: If you are using Docker or Podman to run Cyberismo, you do not need to install Asciidoctor PDF.
+Install Asciidoctor PDF if you need to export content as PDF. This is only needed when running Cyberismo directly on your host (NPM install or build from source) — the Docker and Podman images already include it.
 
 ==== Installing Asciidoctor PDF on macOS
 
   $ brew install ruby && gem install --no-document asciidoctor-pdf rouge
```
henryhaverinen, line 47: Not a part of this PR, but I had to add the Homebrew ruby in path because otherwise the old macOS system ruby would be used instead. Homebrew prints instructions for this, so we we could just make `brew install ruby` a separate command here, and then instruct that the user should put the Homebrew ruby in the path, which should be possible by following the instructions that Homebrew prints

henryhaverinen (approved):

setup-modules.sh
samutoljamo: No, I was wondering where it disappeared :laughing:

PR #1381, opened 2026-05-27
INTDEV-1297 enable autocommit for SaaS

CyberJML:
Bring in the possibility to enable autocommit via environment variables

### DIFF
--- a/tools/backend/src/main.ts
+++ b/tools/backend/src/main.ts
@@ -97,7 +97,10 @@ if (args.export) {
     );
     process.exit(1);
   }
-  const registry = await ProjectRegistry.fromScannedProjects(projects);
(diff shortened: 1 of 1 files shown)


samutoljamo (approved):

tools/backend/src/main.ts
```diff
@@ -97,7 +97,10 @@ if (args.export) {
     );
     process.exit(1);
   }
-  const registry = await ProjectRegistry.fromScannedProjects(projects);
+  const autocommit = process.env.AUTOCOMMIT === 'true';
```
samutoljamo: Actually, I wonder whether we should name is `CYBERISMO_AUTOCOMMIT`? Less likely to exist in OSS

CyberJML: very true

PR #1383, opened 2026-05-27
INTDEV-1291 Improve svg viewer controls on mobile

tminkkinen:


samutoljamo (approved):
Already better, tho I was able to make it unresponsive after zooming in and out on my phone. By unresponsive I mean that nor pinching or trying to move around works. Reopening the modal(without page refresh) works

PR #1387, opened 2026-05-28
Pr1 foundation linktype

samutoljamo:
## EDIT - I've kept everything below unchanged
Ended up doing major simplification, let's do the plan --> apply split separately. They are different concerns, and I feel it makes no sense to try to fit them into one PR so that we can land this refactor.

## Why this change

Today the resource classes mix two concepts that started out nearly identical but have drifted apart:

1. **Mutations** — the intent of a change (rename this link type, delete that card type)
2. **File-system writes** — actually persisting a resource

They were close enough to live together until we started adding **side effects**: renaming a card type has to rename it in the metadata of every card that references it; deleting a resource has to deal with everything that points at it. These side effects come from not having stable references, and they multiply across rename/delete scenarios.

This PR starts a structural separation:

- **All mutations to resources go through a new "mutation" layer.** A mutation represents *what is being changed*, with file-system access abstracted away.
- **Resource classes become thin wrappers over the file system** — direct, side-effect-free. (They aren't fully there yet; that's the direction, completed over the follow-up series.)

## What this buys us

- **Complex mutations get simpler to write.** They're no longer tangled with persistence or with each other — a handler just expresses the change.
- **Cascading changes become planned, not incidental.** Touching other resources used to be a hidden side effect; now it's part of an explicit plan, so we have visibility into exactly what a mutation will change.
- **A `plan → apply` workflow, aimed at the app.** The backend plans a mutation and returns the predicted change set; when the user accepts, the frontend sends the mutation back with a **fingerprint** of the expected changes. Apply re-verifies the change set still matches before committing. This moves the "what will change" computation off the frontend and onto the backend, where the truth lives.
- **A path to unifying mutations and module migrations.** A module migration is just a pre-planned mutation applied the same way, except the primary change isn't re-made — only the cascade. That's why a later PR splits `apply` into two functions (the actual change vs. the cascade effects).

## Scope

**In this PR**
- The mutation engine foundation: `plan`/`apply`, dispatcher + `Handler` strategies, fingerprint, and the reference-rewrite cascade.
- `linkTypes` delete / rename / edit routed through the engine.
- The configuration logger reshaped so the engine is the sole logger on routed paths.

**Deferred to follow-up PRs**
- Splitting `apply` into actual-change + cascade (enables the migration/mutation unification).
- Removing the now-legacy side-effect methods from the resource classes.
- Routing the remaining resource types (cardType, workflow, fieldType, report, template, calculation, graphModel/View) through the engine.
- Wiring the backend `plan → accept → apply` round-trip in the app.

## Things reviewers should know
- **Deleting an in-use link type now cascades instead of being refused.** Previously `delete` threw when a resource was still referenced; the new delete handler strips the references first and then deletes. This is an intentional move toward planned cascades


tools/data-handler/src/mutations/handlers/link-type-rename.ts
```diff
@@ -0,0 +1,98 @@
...
+      await ctx.project.updateCardMetadata(card, metadata);
+    }
+
+    // 2. Rewrite cascading references BEFORE renaming the resource on disk.
+    //    Order matters: cascade scanners look for the old name, and the
+    //    resource file (with that name) must still exist when they run.
+    // TODO: compute accurate counts now that cascade is explicit
+    await rewriteCalculationRefs(ctx.project, oldName, newName);
+    await rewriteHandlebarRefs(ctx.project, oldName, newName);
+    await rewriteCardContentRefs(ctx.project, oldName, newName);
```
tminkkinen, line 56: Is there any risk that we replace something unintentionally using these?

samutoljamo: Yes, but that would be a bug. I don't think there's a safe way to implement this. Git provides means to revert

tools/data-handler/src/mutations/plan.ts
```diff
@@ -0,0 +1,120 @@
...
+    await this.project.lock.write(async () => {
+      await handler.apply(ctx);
+      if (handler.isBreaking) {
+        await this.recordLogEntry(input, recordContext);
+      }
+    });
+    return { success: true };
+  }
+
+  private async affectedFilePathsFor(
```
CyberJML: Could be inline?

tools/data-handler/src/mutations/handlers/default-no-cascade.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class DefaultNoCascadeHandler implements Handler {
+  readonly isBreaking = false;
+
+  matches(ctx: MutationContext): boolean {
+    // Catch-all: dispatcher consults specific handlers first.
+    return ctx.input.kind === 'edit';
+  }
+
+  async apply(ctx: MutationContext): Promise<void> {
```
CyberJML, line 25: Missing input validations?

tools/data-handler/src/mutations/handlers/link-type-delete.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { Card } from '../../interfaces/project-interfaces.js';
+
+export class LinkTypeDeleteHandler implements Handler {
+  readonly isBreaking = true;
+
+  matches(ctx: MutationContext): boolean {
+    return ctx.input.kind === 'delete' && ctx.input.target.type === 'linkTypes';
+  }
+
+  async apply(ctx: MutationContext): Promise<void> {
```
CyberJML, line 26: Missing input validation?

tools/data-handler/src/mutations/handlers/link-type-rename.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { Card } from '../../interfaces/project-interfaces.js';
+
+export class LinkTypeRenameHandler implements Handler {
+  readonly isBreaking = true;
+
+  matches(ctx: MutationContext): boolean {
+    return ctx.input.kind === 'rename' && ctx.input.target.type === 'linkTypes';
+  }
+
+  async apply(ctx: MutationContext): Promise<void> {
```
CyberJML, line 34: Can we trust that ctx always contains input.target or that the follow up steps prevents it?

samutoljamo: I think this could be enforced on the API level(commands), no reason to do that in all handlers.

CyberJML (approved):

PR #1390, opened 2026-05-29
fix(backend): pre-render workflow graphs for static site export (INTDEV-1302)

samutoljamo:
## INTDEV-1302

**Bug:** The static site version of the workflow editor shows an error message in place of the workflow graph.

### Root cause

The workflow graph endpoint `GET /:prefix/workflows/:identifier/graph` (in `tools/backend/src/domain/resources/index.ts`) was **not enumerated for Hono static site generation (SSG)** — it had no `ssgParams` middleware.

During static export (`exportSite` → `toSSG`), only routes that declare `ssgParams` (like `cards/:key`) get their concrete URLs discovered and pre-rendered to `.json` files. Because the workflow graph route had none, no `graph.json` was ever written.

At runtime in static mode the frontend appends `.json` to every API URL (`tools/app/src/lib/swr.ts`). So when the workflow editor (`WorkflowGraph.tsx`) requested `.../graph.json`, it got a 404, and the component fell back to its error state (`workflowGraph.error`).

### Fix

- Add an `ssgParams` middleware to the workflow graph route that enumerates every workflow resource (including module workflows).
- Add a `listWorkflowGraphParams` service helper (`tools/backend/src/domain/resources/service.ts`) that returns `{ prefix, identifier }` for each workflow via `showCmd.showResources('workflows')`.

This mirrors the existing pattern used by the `cards/:key` endpoint.

### Verification

- `tools/backend/test/workflows.test.ts` — added a regression test asserting the enumeration covers every renderable graph route (all backend workflow tests pass).
- End-to-end: ran `exportSite` against the `decision-records` test project and confirmed the static export now produces:
  - `api/projects/decision/resources/decision/workflows/simple/graph.json`
  - `api/projects/decision/resources/decision/workflows/decision/graph.json`
  
  (these files were absent before the fix), with no export errors.
- `pnpm build`, prettier, and eslint all pass on the changed files.

### DIFF
--- a/tools/backend/src/domain/resources/index.ts
+++ b/tools/backend/src/domain/resources/index.ts
@@ -12,8 +12,9 @@
 */
 
 import { Hono } from 'hono';
+import { ssgParams } from 'hono/ssg';
 import * as resourceService from './service.js';
-import type { ResourceValidationResponse } from '../../types.js';
+import type { AppContext, ResourceValidationResponse } from '../../types.js';
 import { UserRole } from '../../types.js';
 import { resourceParamsSchema } from '../../common/validationSchemas.js';
 import { zValidator } from '../../middleware/zvalidator.js';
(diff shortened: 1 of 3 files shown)


samutoljamo:
Opening new PR from non-fork

PR #1392, opened 2026-05-29
fix(app): hide export project option in exported static site

samutoljamo:
The "Export project" entry in the card tree "…" menu is meaningless in a statically exported site, since exporting relies on backend endpoints that are unavailable. Hide it when running in static mode, matching the existing staticMode gating used elsewhere in the app.


samutoljamo:
Wrong repo

PR #1397, opened 2026-06-01
INTDEV-1225 Add first version git push functionality

CyberJML:
First version of the git push feature to the application. The idea is to have a separate SaaS container that is actually doing the pushes, it will also share the public key to the project directory root, like /projects/.git-push/id_rsa.pub, while the projects are /projects/project1 etc. This feature enables seeing/modifying the remote of origin + copying the current tenant-wide public key to allow the pushing to work. 

In the future we most likely will have a separate admin SPA that will handle this kind of settings, but for now we have to do it on the application itself. Pulling is not supported yet


samutoljamo (approved):

tools/app/cypress/e2e/xref.cy.ts
```diff
@@ -78,7 +78,7 @@ describe('Native AsciiDoc xref', () => {
     // Assert the rendered HTML contains a link with the canonical multi-project href.
     cy.then(() => {
       const expectedHref = `/projects/${captured.prefix}/cards/${captured.bKey}`;
-      cy.get('[class="doc"]')
+      cy.get('.doc')
```
samutoljamo: Playwright was merged to main, so rebase and this change may or may not be needed

tools/app/src/lib/utils.ts
```diff
@@ -81,12 +82,22 @@ let _config: AppConfig = { staticMode: false, presenceEnabled: false };
...
 export async function initConfig(): Promise<void> {
-  try {
-    const res = await fetch('/config.json');
-    _config = await res.json();
-  } catch {
-    _config = { staticMode: false, presenceEnabled: false };
+  // Try without extension first (live server), then with .json (static export)
+  for (const url of [
+    globalApiPaths.config(),
+    `${globalApiPaths.config()}.json`,
```
samutoljamo, line 88: Definitely a bit hacky and slows down app start, but I guess there's no good way with the http based config route

tools/backend/src/domain/project/service.ts
```diff
@@ -60,10 +62,16 @@ export async function getProject(
       modules.map((mod) => toModuleInfo(commands, mod.name)),
     );
 
+    const isGitRepo = await commands.project.git.isRepo();
```
samutoljamo: This is not consistant with how we've used data-handler. Project is not meant to be used directly, but the commands themselves are the API. I agree that it's a bit awkward, but I'd keep it consistant as it makes future refactoring easier if we follow the style we've used. I'd recommend just 1 new public function, which returns either the remote url or null if not defined.

tools/backend/src/domain/project/service.ts
```diff
@@ -83,6 +91,10 @@ export async function updateProject(
     }
   }, 'Update project settings');
 
+  if (gitRemoteUrl !== undefined) {
+    await commands.project.git.setRemoteUrl(gitRemoteUrl);
```
samutoljamo: here too(use public command function instead of using it directly) and I'd also put it under atomic for now to be under the write lock for now.

tools/backend/src/app.ts
```diff
@@ -98,6 +99,17 @@ export function createApp(
   }
 
   app.use(treeMiddleware(opts));
+
+  // Public app configuration — no auth required (fetched before login)
```
samutoljamo: I'd reduce the verbosity
`// Public app configuration`

tools/backend/src/app.ts
```diff
@@ -107,6 +119,20 @@ export function createApp(
   app.route('/api/auth', createAuthRouter());
   app.route('/api/projects', createProjectsRouter(registry));
 
+  // Tenant-level SSH public key
+  app.get('/api/public-key', async (c) => {
+    if (!rootPath) return c.json({ publicKey: null });
+    try {
+      const content = await readFile(
+        join(rootPath, '.git-push', 'id_rsa.pub'),
```
samutoljamo, line 185: Why rsa? I'd use `ed25519`

CyberJML: Azure Devops does not support ED :D

CyberJML: image

tools/data-handler/src/utils/git-manager.ts
```diff
@@ -118,8 +118,50 @@ export class GitManager {
     return !status.isClean();
   }
 
+  /** Check whether the project directory is a git repository. */
+  async isRepo(): Promise<boolean> {
+    try {
+      return await this.git.checkIsRepo();
+    } catch {
```
samutoljamo, line 125: Is there a reason to ignore errors?

tools/data-handler/src/utils/git-manager.ts
```diff
@@ -118,8 +118,50 @@ export class GitManager {
...
 
+  /** Get the URL of a named remote. Returns null if the remote does not exist or the directory is not a git repo. */
+  async getRemoteUrl(
+    remoteName: string = GitManager.DEFAULT_REMOTE,
+  ): Promise<string | null> {
+    try {
+      const remotes = await this.git.getRemotes(true);
+      const remote = remotes.find((r) => r.name === remoteName);
+      return remote?.refs?.push ?? null;
+    } catch {
```
samutoljamo, line 140: why ignore errors?

tools/backend/src/app.ts
samutoljamo: Probably easier to use rsa then :D

tools/data-handler/src/utils/git-manager.ts
CyberJML: added logger

tools/data-handler/src/utils/git-manager.ts
CyberJML: added logger

PR #1398, opened 2026-06-01
INTDEV-1200 Replace cypress with playwright

samutoljamo:
Rewrites cypress tests to playwright. Goal was to be a port and thus tests are the same and mostly use the same selectors apart from a few places, which had no direct playwright equivivalent.

Some notably changes:
* Ubuntu action now runs against firefox, webkit and chromium
* Modified the test setup so that there is no need to run tests in seperate processes and also it's possible to run them in parallel, because they don't depend on port 3000, but use an OS-assigned port(however, within test files are test cases are coupled so they obviously cannot be run in parallel)
* Failures are annotated in a way that github support and an html report is produced if at least 1 test fails. For example, see following report made by the last commit: https://github.com/CyberismoCom/cyberismo/actions/runs/26808340210/artifacts/7353136159


tools/app/playwright.config.ts
```diff
@@ -0,0 +1,35 @@
...
+  firefox: 'Desktop Firefox',
+  webkit: 'Desktop Safari',
+};
+
+export default defineConfig({
+  testDir: './e2e',
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 1 : 0,
+  workers: process.env.CI ? 2 : undefined,
```
samutoljamo, line 23: Running locally with 14 cpu threads, I noticed that `strips form action in passthrough block` was a bit flaky. In the future, would be nice to try make tests in such a way that they don't depend on each other, but since we executed them 1-by-1, there surely is some flakyness when running things in parallel. I don't attempt fixing that here and let's not be too aggressive with parallelization.

CyberJML (approved):

tminkkinen (approved):

PR #1425, opened 2026-06-02
INTDEV-1289 Improve TOC and notifications & checks responsivity

tminkkinen:


samutoljamo:
image

In desktop view, the blue thing represents which section are we in. It's kinda useful for the mobile view(if you see TOC, you're on the first section), so maybe disable it?

samutoljamo:
image

Don't have a clear suggestion, but this looks slightly off as "notifications and checks" is the title and the "notifcations", "passed policy checks" and "failed policy checks" are its subtitles, but they are not smaller nor there is any indentation.

tools/app/src/components/card/CardLayout.tsx
```diff
@@ -151,6 +262,38 @@ export const CardLayout = forwardRef<CardLayoutHandle, CardLayoutProps>(
...
+    const attachmentPanel = (
+      <AttachmentPanel
+        cardKey={card.key}
+        attachments={card.attachments ?? []}
+        onInsert={(attachment) => bodyRef.current?.insertAttachment(attachment)}
+      />
+    );
+
+    const viewSidebarContent = (
+      <>
```
samutoljamo, line 282: Noticed that this pattern of saving to variables is used a lot. I think that's fine, since otherwise the final return gets very complex, but gotta be careful, since these are computed on every render although they might not be shown.

samutoljamo (approved):
Code wise seems good, see the previous screenshots though

PR #1431, opened 2026-06-03
Feature/joonalepola/intdev 1306 enable create new project to UI

CyberJML:
More projects view will have "New Project" button
image

There are two options, clone and create new project
image

Clone is simply a clone, might require some extra effort to make functionable in SaaS, since we do not have any keys defined for the OSS container, but the other project should work properly. It will initialize the git directory and leave remote empty, which can be edited from the configurations page. 

Next step is some project information and selecting modules.
image

I removed the external module git address support for now, basically same reason why clone repository will not work for SaaS, might need to hide that one too


tools/app/src/components/modals/CreateProject/projectFormUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+const PREFIX_PATTERN = /^[a-z]+$/;
+// Must match Validate.isValidProjectName in data-handler: ^[A-Za-z ._-]+$, length 1-63
+const PROJECT_NAME_PATTERN = /^[A-Za-z ._-]+$/;
+
+export function isPrefixValid(prefix: string): boolean {
+  return (
+    prefix.length === 0 ||
+    (prefix.length >= 3 && prefix.length <= 10 && PREFIX_PATTERN.test(prefix))
+  );
+}
```
tminkkinen, line 28: Should the length be included in the pattern? Would make this a bit cleaner since we are already using regex `/^[a-z]{3,10}$/`

tools/backend/src/domain/projects/index.ts
```diff
@@ -35,5 +46,149 @@ export function createProjectsRouter(registry: ProjectRegistry) {
...
+
+    try {
+      const clonedPath = await Create.cloneProject(url, rootPath);
+      const scanned = await scanForProjects(clonedPath);
+
+      const registered = [];
+      for (const project of scanned) {
+        if (registry.has(project.prefix)) {
+          continue; // skip duplicates
+        }
```
tminkkinen: Is it possible these duplicates would cause some unexpected problems later?

tminkkinen (commented):
Looked good to me, but maybe @samutoljamo could check at least the data handler part.

tools/app/cypress/e2e/xref.cy.ts
```diff
@@ -0,0 +1,96 @@
...
+      cy.get('.cm-content').type(
+        `See xref:${captured.bKey}.adoc[Go to B] please.`,
+      );
+    });
+    saveContent();
+
+    // Assert the rendered HTML contains a link with the canonical multi-project href.
+    cy.then(() => {
+      const expectedHref = `/projects/${captured.prefix}/cards/${captured.bKey}`;
+      cy.get('.doc')
```
samutoljamo: Might not be needed after rebasing to playwright, but would be nice if there was an e2e test for the project creation

tools/app/src/components/modals/CreateProjectModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+                disabled={!cloneUrl.trim() || isLoading}
+                loading={isLoading}
+              >
+                {t('projectDialog.clone')}
+              </Button>
+              <Button
+                variant="plain"
+                color="neutral"
+                onClick={() => {
+                  setCloneUrl('');
```
samutoljamo, line 254: Quite duplicated logic with these back / next buttons.

tools/app/src/components/modals/CreateProjectModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+    prefix: '',
+    category: '',
+    description: '',
+  });
+  const [createdPrefix, setCreatedPrefix] = useState<string | null>(null);
+  const [selectedModules, setSelectedModules] = useState<Set<string>>(
+    new Set(),
+  );
+
+  const { data: hubModules } = useSWR<ModuleSettingFromHub[]>(
```
samutoljamo: Even tho so far there's only a single place that uses this, I'd add this to frontend's API, since it's easy to do that and keeps the codebase consistant

tools/app/src/lib/api/projects.ts
```diff
@@ -11,9 +11,42 @@
...
+  prefix: string;
+  category?: string;
+  description?: string;
+}
+
+export interface CloneProjectParams {
+  url: string;
+}
+
+export const createProject = async (
```
samutoljamo, line 50: I'd use the `useXMutation` pattern for any requests for consistancy

tools/app/src/locales/en/translation.json
```diff
@@ -409,9 +409,26 @@
...
     "createError": "Failed to create project",
-    "selectProject": "Select a project to get started"
+    "selectProject": "Select a project to get started",
+    "selectMethod": "How would you like to start?",
+    "cloneFromRepo": "Clone from repository",
+    "cloneDescription": "Clone an existing project from a git repository",
+    "createFromScratch": "Create new project",
+    "createDescription": "Start a new project from scratch",
+    "cloneTitle": "Clone from repository",
+    "cloneUrl": "Repository URL",
```
samutoljamo: repoUrl?

tools/backend/src/domain/projects/index.ts
```diff
@@ -35,5 +46,149 @@ export function createProjectsRouter(registry: ProjectRegistry) {
...
+   *     summary: Clone a project from a git repository
+   *     description: Clones a git repo and registers discovered projects
+   */
+  router.post('/clone', requireRole(UserRole.Editor), async (c) => {
+    if (!rootPath) {
+      return c.json({ error: 'Project cloning is not available' }, 500);
+    }
+
+    const body = await c.req.json();
+    const parsed = cloneProjectSchema.safeParse(body);
```
samutoljamo: Prefer zod middleware over defining these manually

tools/backend/src/domain/projects/index.ts
```diff
@@ -35,5 +46,149 @@ export function createProjectsRouter(registry: ProjectRegistry) {
...
+   *     description: Creates a new project on disk, initializes git, adds default hub
+   */
+  router.post('/', requireRole(UserRole.Editor), async (c) => {
+    if (!rootPath) {
+      return c.json({ error: 'Project creation is not available' }, 500);
+    }
+
+    const body = await c.req.json();
+    const parsed = createProjectSchema.safeParse(body);
+    if (!parsed.success) {
```
samutoljamo: zod middleware here too

tools/backend/src/domain/projects/index.ts
```diff
@@ -35,5 +46,149 @@ export function createProjectsRouter(registry: ProjectRegistry) {
...
+      return c.json(
+        { error: `Project with prefix '${prefix}' already exists` },
+        409,
+      );
+    }
+
+    const projectPath = join(rootPath, prefix);
+
+    try {
+      await Create.createProject(
```
samutoljamo: I'd still keep the habit of separating http to `index.ts` while keeping logic in `service.ts`

tools/backend/src/domain/projects/index.ts
```diff
@@ -35,5 +46,149 @@ export function createProjectsRouter(registry: ProjectRegistry) {
...
+        projectPath,
+        prefix,
+        name,
+        category ?? '',
+        description ?? '',
+      );
+
+      const commands = new CommandManager(projectPath, registry.options);
```
samutoljamo: I think all the lines below are not backend's concern, but DH's. Modify createPorject so that it initiailizes the git repository and adds the default hub.

tools/data-handler/src/commands/create.ts
```diff
@@ -648,6 +654,63 @@ export class Create {
...
+      const projects = await scanForProjects(tempClonePath);
+      if (projects.length === 0) {
+        throw new Error(
+          'Cloned repository does not contain any Cyberismo projects',
+        );
+      }
+
+      // Move to the final destination
+      await mkdir(destPath, { recursive: true });
+      await rename(tempClonePath, finalPath);
```
samutoljamo, line 699: This is a bit risky, because in practice /tmp might live on a different disk. And this is not that far fetched, because for example I personally use such configurations(which are by default) that /tmp is `tmpfs` ie. it lives in memory. Thus, it fails on the rename. 

Two options:
1. Use /tmp, but copy
2. Use some directory in the destination(I think this might be better)

samutoljamo (approved):
Left a few comments, otherwise looks good

tools/backend/src/domain/projects/index.ts
CyberJML: This is a hard one, added a response which projects are duplicates. We need to come up with something here

tools/data-handler/src/commands/create.ts
```diff
@@ -621,26 +621,29 @@ export class Create {
     });
 
     await Promise.all(
-      Create.JSONFileContent.map(async (entry) => {
-        if ('cardKeyPrefix' in entry.content) {
-          if (entry.content.cardKeyPrefix.includes('$PROJECT-PREFIX')) {
-            entry.content.cardKeyPrefix = projectPrefix.toLowerCase();
+      // Deep clone to avoid mutating the static template
+      JSON.parse(JSON.stringify(Create.JSONFileContent)).map(
```
samutoljamo: Use `structuredClone`

PR #1443, opened 2026-06-05
feat(mutations): card type handlers (PR2)

samutoljamo:
Second PR of the migration-imp split (follows #1387). Extracts the five CardType mutation handlers into the mutation engine: rename, delete, add-custom-field, remove-custom-field, workflow-change.

- Rename/delete carry their cascade (card metadata rewrite, link-type ref stripping, card removal) and delegate the remainder to the still-alive legacy path.
- The three edit handlers are deliberately thin routers into `CardTypeResource.update` — the legacy in-class cascade still runs exactly once; each handler absorbs its cascade when the legacy path is deleted in the cleanup PR ("scaffolding stays" invariant).
- Routing: cardType renames, deletes, and exactly two edit shapes (workflow change, customFields add/remove) divert to `ResourceMutations`; everything else stays on the legacy path.

### DIFF
--- a/tools/data-handler/src/mutations/handlers/card-type-delete.ts
+++ b/tools/data-handler/src/mutations/handlers/card-type-delete.ts
@@ -17,8 +17,6 @@ import { resourceNameToString } from '../../utils/resource-utils.js';
 import { ResourcesFrom } from '../../containers/project/resources-from.js';
 import type { Card } from '../../interfaces/project-interfaces.js';
 import type { Operation } from '../../resources/resource-object.js';
-import { Remove } from '../../commands/remove.js';
-import { Fetch } from '../../commands/fetch.js';
 
 export class CardTypeDeleteHandler implements Handler {
   readonly isBreaking = true;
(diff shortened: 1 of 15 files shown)


tools/data-handler/src/mutations/handlers/card-type-remove-custom-field.ts
```diff
@@ -0,0 +1,52 @@
...
+ * alwaysVisibleFields / optionallyVisibleFields. The cascade is performed by
+ * CardTypeResource.update; this handler routes the operation and marks it
+ * breaking.
+ */
+export class CardTypeRemoveCustomFieldHandler implements Handler {
+  readonly isBreaking = true;
+
+  matches(ctx: MutationContext): boolean {
+    if (ctx.input.kind !== 'edit') return false;
+    if (ctx.input.target.type !== 'cardTypes') return false;
```
tminkkinen, line 31: Maybe consider creating some helpers if these checks will be common. Not necessary

tminkkinen (approved):

tools/data-handler/src/mutations/handlers/card-type-delete.ts
```diff
@@ -0,0 +1,95 @@
...
+    const linkTypes = ctx.project.resources.linkTypes(ResourcesFrom.localOnly);
+    for (const lt of linkTypes) {
+      const data = lt.data;
+      if (!data) continue;
+      for (const field of [
+        'sourceCardTypes',
+        'destinationCardTypes',
+      ] as const) {
+        if (data[field].includes(cardTypeName)) {
+          await lt.update({ key: field }, {
```
CyberJML, line 51: Could we just use promise.all after the loop? Could there be a lot of parallel waiting happening with the current approach?

CyberJML (approved):

tools/data-handler/src/mutations/handlers/card-type-delete.ts
samutoljamo: Yes but I think it would be better to think about this as a separate concern: this has lots of other stuff that could be run in parallel. Practically the operations are very fast too, so I think it can be solved a bit more nicely just put Promise.all everythere.

PR #1448, opened 2026-06-05
INTDEV-1312 Make template card metadata fields editable

tminkkinen:
## Summary

Template cards opened in the configuration UI showed their metadata fields **read-only with a lock icon** — they couldn't be edited ([INTDEV-1312](https://cyberismo.atlassian.net/browse/INTDEV-1312)). This replaces the old config card editor with a dedicated, **edit-first template card editor**:

- A single working **draft** is the source of truth for both editing and Preview, so per-field **Cancel** and the Edit/Preview toggle never desync.
- **Title, labels, every custom field, and the AsciiDoc content** are edit-first and **save independently**; Save/Cancel are dirty-gated.
- **Preview** (a single edit-first toggle button) renders the card read-only, reflecting the current unsaved draft. No links UI (not applicable to template cards).
- Editing is gated on the **Admin** role (`useHasMinRole(UserRole.Admin)`), matching the previous config-editor permission.
- Split into small components under `config-editors/template-card/` (`TemplateCardEditor`, `TitleEditor`, `MetadataFields`/`MetadataFieldRow`, `ContentEditor`, `TemplateCardPreview`, `PreviewToggle`, `draft.ts`); reuses `EditableField`, `CardTitle`/`MetadataSection`/`CardBody` (preview), `AttachmentPanel`, and a line-numbered CodeMirror editor.
- Consolidated `coerceMetadataValue` / `metadataFieldRowId` / metadata equality into `lib/utils`, and hardened `getDefaultValue` to accept both query-shaped and raw (string-array) **list** values — fixing a bug where list fields showed dirty / couldn't save / broke Preview.

## Testing

- Unit/component (Vitest + RTL): edit-first inputs, per-field save, Cancel reverts, Preview reflects the unsaved draft, list-value regression, read-only node; plus `getDefaultValue` list normalization.
- e2e (`app.spec.ts`): updated the template-card flow (edit + save, unsaved title in Preview, field Cancel→Preview sync).
- `pnpm lint`, `tsc -b`, `vitest run` (220) and `pnpm build` all pass.

[INTDEV-1312]: https://cyberismo.atlassian.net/browse/INTDEV-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


samutoljamo:
A general comment:
Is the separate path worth it for the template card? I wonder it could simply use the same components as normal cards use?

CyberJML:
One thing I would align is the button locations between edit and preview mode. Currently edit mode seems like this:
image

We have the preview button with large margin from the right border, since we have the attachments there. But with the preview mode the attachment view is gone and the page stretches to the right border like this:
image

Maybe just keep the card width same? Or are we actually missing the table of contents from the preview? Anyways, the button is jumping too much

CyberJML:
Is it intended that the header is in readonly mode that seems like it's editable? 
image

Is there a mode that fits better for this situation?

CyberJML (approved):

PR #1450, opened 2026-06-08
INTDEV-1310 Clear error when shortText value exceeds length limit

samutoljamo:
## Root cause
Setting a `shortText` field value longer than 80 characters fails validation in `Validate.validateCustomFields` (`tools/data-handler/src/commands/validate.ts`). `validType()` rejects the value because it exceeds `SHORT_TEXT_MAX_LENGTH` (80), but the message builder had no `shortText`-specific branch — so a too-long *string* fell through to the generic type-mismatch message:

> `In card 'test_z9h94s81' field 'test/fieldTypes/shortText' is defined as 'shortText', but it is 'string' with value of "…"`

This is misleading: the value **is** a string; it's just over the length limit. The 80-char limit and this message both live entirely in our own code — no JSON schema / AJV / Zod validates `shortText` value length.

## Fix
Add a `shortText` branch alongside the existing `enum`/`person` cases that reports the allowed and actual lengths when a string value exceeds `SHORT_TEXT_MAX_LENGTH`:

> `In card '…' field '…' value exceeds the maximum length for 'shortText': 80 characters allowed, but value has 120 characters`

Validation behaviour is unchanged — only the error message improves.

## Verification
- New regression test in `command-edit.test.ts` reproduces the bug via the real `editCardMetadata` path; confirmed it **fails** against current code with the old misleading message, and **passes** with the fix.
- Full `data-handler` suite green: 1409 passed, 1 skipped.
- `eslint`, `prettier --check`, and `tsc --noEmit` clean on the changed package.

Ticket: INTDEV-1310 — https://cyberismo.atlassian.net/browse/INTDEV-1310

### DIFF
--- a/tools/data-handler/src/commands/validate.ts
+++ b/tools/data-handler/src/commands/validate.ts
@@ -811,6 +811,17 @@ export class Validate {
           );
           continue;
         }
+        if (
+          fieldType.dataType === 'shortText' &&
+          typeOfValue === 'string' &&
+          this.length(card.metadata[field.name] as string) >
+            SHORT_TEXT_MAX_LENGTH
+        ) {
+          validationErrors.push(
+            `In card '${card.key}' field '${field.name}' value exceeds the maximum length for 'shortText': ${SHORT_TEXT_MAX_LENGTH} characters allowed, but value has ${this.length(card.metadata[field.name] as string)} characters\n`,
+          );
+          continue;
+        }
         validationErrors.push(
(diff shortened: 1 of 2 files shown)


samutoljamo:
I didnt not do the refactoring here, but I'm not a big fan of how the validation and generating messages is separated as it is a bit confusing. However, if we wanted to improve I think we should do a unified validation approach based on a library(likely zod) that would allow tying schema's, DH's, backend's and apps validation together.

tminkkinen (approved):

PR #1454, opened 2026-06-10
INTDEV-1326 Enable SaaS Git service

CyberJML:


docs/cardRoot/docs_3/c/docs_mlnc22el/index.adoc
```diff
@@ -74,6 +74,8 @@ $ cyberismo import module <https git URL> [branch name]
 
 If the Git service requires authentication, you must provide access token credentials in the environment variables `CYBERISMO_GIT_USER` and `CYBERISMO_GIT_TOKEN` or in a `.env` file that defines these variables.
 
+If your deployment uses a dedicated git-service container, set `GIT_SERVICE_URL` (for example `http://git-service:8080`). When set, Cyberismo delegates git clone and remote tag operations to git-service instead of making network git calls directly.
```
samutoljamo: Probably shouldn't be public documentation?

README.adoc
```diff
@@ -108,6 +108,8 @@ Copy *env.example* files to respective *.env* files:
 cp tools/backend/env.example tools/backend/.env
 ----
 
+Optional for SaaS-style deployments: set `GIT_SERVICE_URL` (for example `http://git-service:8080`) to delegate external git clone and remote tag operations to git-service.
```
samutoljamo: Same here

docs/cardRoot/docs_3/c/docs_mlnc22el/index.adoc
CyberJML: true

tools/app/src/components/modals/CreateProjectModal.tsx
```diff
@@ -210,6 +203,11 @@ export function CreateProjectModal({ open, onClose }: CreateProjectModalProps) {
         <LinearProgress determinate value={progressValue} sx={{ my: 1 }} />
         <Divider />
         <DialogContent sx={{ overflow: 'auto' }}>
+          {error && (
```
CyberJML: Using own error notification, since with empty project the application hasn't loaded yet and there's nothing to show the errors.

samutoljamo: Looks like during the multi-project addition the common layout is not anymore a common layout, likely because it had assumptions that are only useful for projects / configuration. Notifications is not one those things tho, so could just move it upwards so that it's available here, because it's a  bit confusing why it wouldn't be available in certain components

tools/data-handler/src/commands/create.ts
```diff
@@ -706,8 +720,12 @@ export class Create {
       }
       throw error;
     } finally {
-      // Clean up temp directory (contains leftovers on failure, empty wrapper on success)
-      await rm(tempDir, { recursive: true, force: true }).catch(() => {});
+      // Clean up on failure. On success rename() has already moved tempClonePath
+      // so rm with force:true is a no-op. Both paths are handled uniformly here.
+      const cleanupPath = tempDir ?? tempClonePath;
+      if (cleanupPath) {
+        await rm(cleanupPath, { recursive: true, force: true }).catch(() => {});
```
samutoljamo: Could add a logger warning that the cleanup failed

tools/data-handler/src/commands/fetch.ts
```diff
@@ -59,12 +60,14 @@ export class Fetch {
     location: string,
   ): Promise<number | undefined> {
     try {
-      const url = new URL(`${location}/${MODULE_LIST_FILE}`);
+      const url = new URL(
+        `${location.replace(/\/+$/, '')}/${MODULE_LIST_FILE}`,
```
samutoljamo: What about just `new URL(MODULE_LIST_FILE, location)`

tools/data-handler/src/utils/git-service-client.ts
```diff
@@ -0,0 +1,138 @@
...
+        'application/json',
+    },
+  });
+}
+
+export async function fetchTags(url: string): Promise<string[]> {
+  const query = `/tags?url=${encodeURIComponent(url)}`;
+  const body = await requestJson(query);
+
+  if (Array.isArray(body) && body.every((item) => typeof item === 'string')) {
```
samutoljamo: We anyway planned on adding zod to DH, so I'd just do it now and use it here, so the code would be nicer

tools/data-handler/src/utils/git-service-client.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export async function clone(options: {
+  url: string;
+  ref?: string;
+  shallow?: boolean;
+}): Promise<string> {
+  const body = await requestJson('/clone', {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({
```
samutoljamo: `requestJson` already assumes you are requesting JSON so might as well add the headers / body there without manually stringifying at all callers

tools/data-handler/test/modules/source-git.test.ts
```diff
@@ -52,6 +64,21 @@ describe('modules/source-git', () => {
     expect(spy).toHaveBeenCalledWith('https://u:t@example.com/repo.git');
   });
 
+  it('listRemoteVersions uses GitManager normalization path when git-service is enabled', async () => {
```
samutoljamo: Does this test do anything? It's mocking that the `listRemoteVersionTags` returns everything in the the correct format already, no?

tools/app/src/components/modals/CreateProjectModal.tsx
CyberJML: Changed

tools/data-handler/src/utils/git-service-client.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+    body: hasBody ? JSON.stringify(options.json) : undefined,
+  });
+
+  const body = (await response.json()) as unknown;
+  if (!response.ok) {
+    const parsed = errorResponseSchema.safeParse(body);
+    const detail = parsed.success
+      ? parsed.data.error
+      : `git-service request failed (${response.status})`;
```
samutoljamo: Could rather say `unexpected format`. The body would be safe to include in the error too, no?

samutoljamo (approved):

PR #1457, opened 2026-06-12
INTDEV-1231 Add hub management to UI

samutoljamo:
Allows managing hubs through the config UI.

Changes are mostly tests and a small refactor that for re-using a react component.

Currently intentional behavior:
* Tolerates errors so valid urls that hub urls are added even if they are not recognized to be a hub
* During "update all" you get a error for these invalid hubs, but valid hubs do still succeed

Hubs do not have per hub "update" button. DH did not support this beforehand and I was wondering if its even needed? The "update all" is also on the fence since hubs could be always refetched instead of being exclicitly updated. To not be too spammy, some sort of short term cache could be added in that route was token so that visits to the configuration editor do not constantly spam the configured urls.


tools/data-handler/src/commands/fetch.ts
```diff
@@ -176,6 +190,14 @@ export class Fetch {
           return true;
         }
 
+        // Cache written by an older version lacks per-hub module data.
```
samutoljamo, line 193: Assumption for now is that updating the hub is not common so this is just a one-off check rather than versioning that cache file

tools/backend/test/project.test.ts
```diff
@@ -123,3 +175,155 @@ describe('Project endpoints', () => {
...
+  ])('POST /api/project/hubs returns 400 for %s', async (_name, location) => {
+    const response = await app.request('/api/projects/test/project/hubs', {
+      method: 'POST',
+      headers: { 'content-type': 'application/json' },
+      body: JSON.stringify({ location }),
+    });
+    expect(response.status).toBe(400);
+  });
+
+  test('POST /api/project/hubs returns 500 for duplicate hub', async () => {
```
tminkkinen, line 248: Might not be caused by this, since didn't spot where this happens, but should probably throw 409 or 400 instead of 500 for duplicate hub.

samutoljamo: Yeah that's a bit of a tech debt that should be solved and applies to many things. The 500 comes from the global error handler. My read is that at the very initial stages command handler was the one returning and http status code, but app did not use that path, since it was not convinient to call(each endpoint had to make a CLI-like request). The commands became the public interface that returns the results as JS objects and throws for errors. The problem is that throwing plain errors hides the "category" of the error. Personally I'd think the better way in general is that we'd refactor DH itself so that it will throw custom errors that either are or inherit from an error class that represents the category. I.e.
```
throw NotFoundError('card was not found')
// --> backend's global handler can check if type is NotFound and return 404

throw AlreadyExistsError('Hub already exists')
// --> backend's global handler can check if type is AlreadyExistsError and throw 409
```
I think it deserves its own PR tho and a bit more thought to what the actual base errors should really be.

tminkkinen (approved):

PR #1460, opened 2026-06-15
INTDEV-1246 Enable report macro without a query, add math helpers

tminkkinen:
## Summary

Implements [INTDEV-1246](https://cyberismo.atlassian.net/browse/INTDEV-1246) — make the report macro usable without running a logic program query.

- **Empty query → skip Clingo.** When a report's `query.lp.hbs` is empty (or absent), `generateReportContent` no longer runs the logic program; it renders the content template from the macro parameter context alone. This flows through all three report consumers (the `{{#report}}` macro, PDF export, and graphviz) since they share `generateReportContent`.
- **Math helpers** for inline computation in report templates, registered alongside the comparison helpers:

  | Helper | Notes |
  |---|---|
  | `add A B …`, `multiply A B …` | variadic (≥2 operands) |
  | `subtract A B`, `divide A B` | `divide` does integer division when both operands are integers; returns `"Division by zero error"` on divide-by-zero |
  | `int A` | truncates toward zero |
  | `max A B …`, `min A B …` | variadic; preserves operand value/type |

  Non-numeric operands and fewer-than-two variadic operands throw.

## Tests

- `test/utils/handlebars-helpers.test.ts` — per-helper coverage (integer vs number results, variadic forms, divide-by-zero, invalid input).
- `test/utils/report.test.ts` (new) — stub `CalculationEngine` verifies the logic program is skipped for empty/undefined queries, run otherwise, and that math helpers work inside query templates.
- `test/command-report.test.ts` — end-to-end render of the new `percentageReport` fixture with no query.

[INTDEV-1246]: https://cyberismo.atlassian.net/browse/INTDEV-1246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/data-handler/src/utils/handlebars-helpers.ts
+++ b/tools/data-handler/src/utils/handlebars-helpers.ts
@@ -98,6 +98,82 @@ export function registerClingoHelpers(instance: typeof Handlebars) {
 }
 
 /**
+ * Coerces a Handlebars helper argument into a finite number.
+ * @param value - The value to coerce (a number or numeric string)
+ * @returns The value as a finite number
+ * @throws Error if the value is not a finite number
+ */
+function toNumber(value: unknown): number {
+  const result = typeof value === 'number' ? value : Number(value);
+  if (!Number.isFinite(result)) {
+    throw new Error(`Expected a number but got '${String(value)}'`);
+  }
+  return result;
+}
+
+/**
+ * Extracts and coerces the numeric arguments passed to a variadic helper.
+ * The final argument is always the Handlebars options object, so it is dropped.
+ * @param args - The arguments Handlebars passed to the helper
+ * @returns The leading arguments as finite numbers
+ * @throws Error if fewer than two operands are given or a value is not numeric
+ */
+function numericArgs(args: unknown[]): number[] {
+  const operands = args.slice(0, -1);
+  if (operands.length < 2) {
+    throw new Error('Expected at least two numeric operands');
+  }
+  return operands.map(toNumber);
+}
+
+/**
+ * Registers basic math helpers with a Handlebars instance.
+ *
+ * These let report templates compute values inline without a logic program, in
+ * both the query template and the content template. Arithmetic uses native
+ * JavaScript numbers and helpers return numbers, so results are never quoted.
+ * `divide` always performs normal (non-integer) division and throws on
+ * divide-by-zero, which surfaces as an inline error where the report macro is
+ * rendered. `round` rounds to the given number of decimal digits (0 by default).
+ * @param instance handlebars instance
+ */
+export function registerMathHelpers(instance: typeof Handlebars) {
+  instance.registerHelper('multiply', (...args: unknown[]) =>
+    numericArgs(args).reduce((a, b) => a * b),
+  );
+  instance.registerHelper('add', (...args: unknown[]) =>
+    numericArgs(args).reduce((a, b) => a + b),
+  );
+  instance.registerHelper(
+    'subtract',
+    (a: unknown, b: unknown) => toNumber(a) - toNumber(b),
+  );
+  instance.registerHelper('divide', (a: unknown, b: unknown) => {
+    const divisor = toNumber(b);
+    if (divisor === 0) {
+      throw new Error('Division by zero');
+    }
+    return toNumber(a) / divisor;
+  });
+  instance.registerHelper('round', (...args: unknown[]) => {
+    const operands = args.slice(0, -1);
+    if (operands.length < 1) {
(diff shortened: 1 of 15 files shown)


tools/data-handler/src/utils/handlebars-helpers.ts
```diff
@@ -97,6 +97,75 @@ export function registerClingoHelpers(instance: typeof Handlebars) {
...
+    const dividend = toNumber(a);
+    const divisor = toNumber(b);
+    if (divisor === 0) {
+      return 'Division by zero error';
+    }
+    return Number.isInteger(dividend) && Number.isInteger(divisor)
+      ? Math.trunc(dividend / divisor)
+      : dividend / divisor;
+  });
+  instance.registerHelper('int', (a: unknown) => Math.trunc(toNumber(a)));
```
samutoljamo: `int` was in the spec, but I'd reconsider the naming here. We don't really have types, everything is a string, since we have templates. Yes, it's familiar for programmers with certain backgrounds, but I'd assume the rounding would be the more usual operation. If we referenced things using the mathematical terms, the options here would:
- `truncate`/`trunc`
- `ceil`
- `floor`
- `round`

I'd argue the `round` would likely be the most useful one. And it could be ie. `round(num, digits)` instead of framing things as "ints" and "numbers"

samutoljamo: We could discuss this, because these are a bit annoying to change, since technically it requires a migration

tools/data-handler/src/utils/handlebars-helpers.ts
```diff
@@ -97,6 +97,75 @@ export function registerClingoHelpers(instance: typeof Handlebars) {
...
+    'subtract',
+    (a: unknown, b: unknown) => toNumber(a) - toNumber(b),
+  );
+  instance.registerHelper('divide', (a: unknown, b: unknown) => {
+    const dividend = toNumber(a);
+    const divisor = toNumber(b);
+    if (divisor === 0) {
+      return 'Division by zero error';
+    }
+    return Number.isInteger(dividend) && Number.isInteger(divisor)
```
samutoljamo: This is also something I'd reconsider. I'm not sure why we'd want to differenciate between integers and numbers

henryhaverinen: I may have forgot the original thinking already, but the reason might have been the fact that Clingo does not support floating point numbers but only integers. In the Cyberismo, we support the number type and it should be represented as a string in logic programs. We might need to provide floating point numbers for Vega or Vega-Lite, so these helpers should support both integers and numbers.

henryhaverinen: If the parameters are coming directly from the report macro parameters, then they're going to be JSON integers or numbers, not numbers encoded as strings. The original reasoning may have been for being able to use these also with query results.

samutoljamo: > Clingo does not support floating point numbers but only integers

Yes and I see the thinking that by having *integers*, it's less easy to make mistakes and create an invalid logic program. Though that comes at the cost of having the user to think about types of numbers, because using a floating point number that ends up into a Clingo program produces an error. If we only had numbers, there'd be no need to consider types or do conversions except when the output is given to Clingo as a number.

The second problem is that our type separation is not enforced, since JS does not separate between them. A value is an int, if `Number.isInteger`returns `true`. See:
```
> Number.isInteger(3)
true
> Number.isInteger(3.0)
true
```
Thus, if the numbers you're using happen to be whole numbers, you cannot do floating point math with the current spec unless we'd extend the helpers to enforce the floating point / int mode. This is why I was thinking it would be overall simpler to not separate numbers, but have `round` or `int` method that must be used if you need an integer for Clingo.

henryhaverinen:
Conclusion from our discussion:

- These helpers can be used in both the query template (resulting logic program) and the content template (resulting Cyberismo AsciiDoc)
- The helpers do not need special handling for integers and floats
- Even though Cyberismo represents floats in ASP as strings, the helpers never return their results with quotes
- A `round(num, digits)` helper would be good, and `round(num)` as a shortcut for `round(num, 0)`, and we don't need the `int` helper. For example, `round(1.5, 0)` returns 1 without any decimal digits.
- It would be nice if division by zero would cause the report macro where the error happens to render an error, regardless of whether the division by zero happens in the query side or content side. If this is not feasible easily, then divide should just return the error string as it currently does

samutoljamo (approved):

PR #1464, opened 2026-06-17
INTDEV-1332 Seal module configuration logs per minor version

samutoljamo:
Seals a module's configuration-change log on every **minor and major** version bump so consumers can replay those changes on module update.

Previously the seal only had a "target".  This becomes a problem when migration logs have non-linear paths:
If you release v-2.0, it would be impossible to determine whether the migration log covers changes from 1.7 to 2.0 or 1.8 to 2.0. 

This sort of admits that we cannot do non-linear updates, but since we have semvers, we have to make the update paths linear. In theory, nothing prevents from creating manual migration logs for newly released minor versions. In most cases, just appending the updates of say 1.7-1.8 to the migrationlog between 1.7 and 2.0 would likely work. However that does not work always: ie. renaming a resource in both versions requires deducing that the one in the newer version should stay and thus they could combined:
```
1.7-2.0 -> rename A to B
1.7-1.8 -> rename A to C
# likely what we want:
1.8-2.0 -> rename C to B
```
However, even in this simple example, this might not be enough for the full migration, as likely the rename carries not just a name, but also meaning and the template changes are not reflected to cards. However, I think this should be a future consideration, not something to be done in this PR.

NOTE: This is a breaking change, migration added in #1468

### DIFF
--- a/tools/data-handler/src/commands/version.ts
+++ b/tools/data-handler/src/commands/version.ts
@@ -67,8 +67,9 @@ export class Version {
       );
     }
 
-    // Snapshot the current migration log with the new version
-    if (ConfigurationLogger.hasBreakingChanges(this.project.basePath)) {
+    // Seal on the first version and every minor/major bump. Patches on an
+    // existing version never seal; they must keep a clean log (guarded above).
+    if (!currentVersion || bumpType !== 'patch') {
       await ConfigurationLogger.createVersion(
         this.project.basePath,
         newVersion,
--- a/tools/data-handler/src/commands/version.ts
+++ b/tools/data-handler/src/commands/version.ts
@@ -46,12 +46,13 @@ export class Version {
 
     const currentVersion = this.project.configuration.version;
 
-    // Guard: breaking changes require a major bump.
+    // Guard: breaking changes cannot ship in a patch. Minor and major bumps
+    // seal the log; consumers replay it on module update.
     // Skipped for the first version — there is no predecessor to break against.
-    if (currentVersion && bumpType !== 'major') {
+    if (currentVersion && bumpType === 'patch') {
       if (ConfigurationLogger.hasBreakingChanges(this.project.basePath)) {
         throw new Error(
-          'Cannot publish a patch or minor version: breaking configuration changes detected. Use a major version bump.',
+          'Cannot publish a patch version: breaking configuration changes detected. Use a minor or major version bump.',
         );
       }
     }
--- a/tools/data-handler/src/mutations/replay/seal-files.ts
+++ b/tools/data-handler/src/mutations/replay/seal-files.ts
@@ -1,0 +1,66 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation. This program is distributed in the hope that it
+  will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+  of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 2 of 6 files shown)


samutoljamo:
@CyberJML I was thinking that we should, in this PR, also remove the guard for producing empty migration logs. Otherwise it is not possible to determine, whether a log is missing or there are just no changes. This does mean that it would introduce another thing that will break modules if you use them wrong, rather than being more easily modifiable. Though versioning in general depends on that too, it will break if you don't use `cyberismo create version` and `cyberismo publish`

tools/data-handler/src/commands/version.ts
tminkkinen, line 1: This looks a bit confusing as minor version allows breaking changes. Are those changes actually breaking, or what is the reasoning behind allowing breaking changes with minor version bumps? At the moment, I can't wrap my head around what has been discussed regarding this earlier :D

samutoljamo: Below is my mental model right now. I agree that "breaking" is a bit confusing, because now we can auto-resolve those. However, it's important to understand the "migrations" only provide data-integrity, they do not promise a "migration" for a project.

| Bump | Touches a referenced identity? | Migration log (seal)? | What the consumer has to do |
|------|---------------|----------------------|------------------------------|
| **patch** | no | none — clean log required | nothing; it's a pure file copy |
| **minor** | yes, but we can migrate it for you | sealed, replayed as a cascade | nothing *visible* — the cascade runs automatically on update |
| **major** | yes, and we **can't** migrate it | — | a human has to act |

Here referenced identity is for example a name of a resource or "workflow" field of a cardtype(basically a foreign key). Note that the minor/major is not enforced by anything, it is a decision of a module developer. 

For example, our "auto migrations" are deleting all cards after their cardtype was removed in a module. Major versions updates would be mostly used for larger refactoring where there is simply no way to represent the change in such a way that would work for a consumer of the project.

In practice, I'd assume it goes something like the following:
1. Release eu-cra module v1
2. EU defines new laws, update module with new requirements, v1.1
3. You realize that there is a better way to represent things, you do completely new templates/resources and release v2

Now if in step 3 the representation changed substantionally so you created new card types and delete old ones --> all consumers that update get all their cards deleted

With the current system, there be two possible paths:
* Update from v1 to v2 is human/AI driven
* You do not actually delete old cardtypes, but rename them to say: `processDeprecated` Now projects could adopt v2 and a more controlled migration --> drop deprecated card types in v3

tminkkinen: That makes sense, and I was guessing it would be something like that. It's a bit grey area if changes requiring migration should be major bumps if we want to strictly follow semantic versioning for modules. IIRC that would mean also adding new fields.

tminkkinen (approved):

tools/data-handler/src/commands/version.ts
samutoljamo: I feel like correct move may be changing the vocabulary. Cyberismo is content after all, not code. So I feel that since we now guarantee data integrity, we would not have to call them "breaking". We could allow these type of changes even in patch versions, but to provide linear update path, we'd need easier ways to make multiple migration logs to the same version. Say:
release v1.1.0
release v1.2.0 <-- migration from 1.1.0->1.2.0 is born
release v1.1.1 <-- migration from 1.1.0->1.1.1 is born, deduce 1.1.1->1.2 automatically

Though this gets a bit messy when you have multiple patches as you'd need to cover all cases. If that algorithm would require no user attention and is deterministic, we wouldn't need to create the actual log files. But if we wouldn't achieve that, then we'd need to create migration logs for every possible combination. 
Unless: We could remove the old seals, meaning that migrate oath 1.1.0-->1.1.1-->1.2.0 replaces 1.1.0->1.2.0. That though breaks my current assumption that the seals would be immutable.

PR #1465, opened 2026-06-17
INTDEV-1329 Enable service account access to APIs

CyberJML:


tools/backend/src/domain/resources/index.ts
```diff
@@ -213,7 +213,7 @@ router.get(
 
 router.post(
   '/:prefix/:type/:identifier/operation',
-  requireRole(UserRole.Admin),
+  requireRole(UserRole.Admin, [UserRole.Connector]),
```
samutoljamo, line 216: If we start having more of the fine grained needs, the first thing to think would be whether we:
1. Handle authorization on DH
2. Handle authorization fully on the backend, which would require separating the routes

samutoljamo (approved):
Seems good

PR #1467, opened 2026-06-17
INTDEV-1349 Module-update replay engine + import integration

samutoljamo:
The core of the feature: replay a module's recorded breaking changes into the consumer project on update.

## Changes
- **`replay.ts`** — three phases: `planModuleReplays` (correlate installed vs resolved modules by source location, detect downgrade/non-linear/chain-gap conflicts, read + validate seal entries), `filterStepsToApplied` (skip modules whose apply didn't land), `executeModuleReplays` (apply each entry as a `replay`-origin mutation — cascade-only, no host log write — with a single cache refresh at the end). Three error classes mark plan/execute/validate phases.
- **`convert.ts`** — reconstruct a `MutationInput` from each released-format log entry.
- **`import.ts`** — wires replay into the single module-update transaction (plan → apply modules → execute replays → validate).
- **Handler hardening**: field-type rename cascade is existence-blind and complete; removed workflow state falls back to the new-card state; replay-seam guards (cache invalidation, partial-apply filtering, unknown-op).

### DIFF
--- a/tools/data-handler/src/modules/resolver.ts
+++ b/tools/data-handler/src/modules/resolver.ts
@@ -173,6 +173,22 @@ export async function resolveModules(
     );
     const childConfig = await readModuleConfig(path);
 
+    // The applier installs under the DECLARED name while replay cascades
+    // rewrite references to the prefix the staged config declares; letting
+    // a renamed module through would corrupt the project. Imports are
+    // unaffected: their declaration name is read from the new config and
+    // already matches. Relocating an existing installation is out of
+    // scope for automatic updates, so refuse loudly.
+    if (childConfig.cardKeyPrefix !== decl.name) {
+      throw new Error(
+        `Module '${decl.name}' renamed its prefix to ` +
+          `'${childConfig.cardKeyPrefix}'; automatic update cannot ` +
+          `relocate it. Re-import the module ` +
+          `(cyberismo module import ${decl.source.location}) ` +
+          `to adopt the new prefix.`,
+      );
+    }
+
     const resolvedEntry: ResolvedModule = {
       declaration: { ...decl },
       ref,
--- a/tools/data-handler/src/mutations/handlers/field-type-data-type.ts
+++ b/tools/data-handler/src/mutations/handlers/field-type-data-type.ts
@@ -62,9 +62,7 @@ export class FieldTypeDataTypeHandler implements Handler<EditInput> {
           fromType,
           toType,
         );
-        // Value was already null/undefined, or the conversion could not be
-        // determined (e.g. a recorded entry without the old data type);
-        // leave the card untouched rather than writing undefined.
+        // Conversion indeterminate — leave the card untouched.
         if (converted === null || converted === undefined) continue;
         metadata[fieldName] = converted as MetadataContent;
         await ctx.project.updateCardMetadata(card, metadata);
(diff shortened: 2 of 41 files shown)


tools/data-handler/src/modules/resolver.ts
```diff
@@ -173,6 +173,22 @@ export async function resolveModules(
     );
     const childConfig = await readModuleConfig(path);
 
+    // The applier installs under the DECLARED name while replay cascades
```
samutoljamo, line 176: A bit of a problem, but could be fixed later properly

tools/data-handler/src/mutations/handlers/field-type-data-type.ts
```diff
@@ -74,29 +76,21 @@ export class FieldTypeDataTypeHandler implements Handler<EditInput> {
     }
   }
 
-  // Cards of card types that declare this field type: local project cards plus
-  // local (non-module) template cards.
+  // Local project cards plus local (non-module) template cards that hold a
+  // value under this field. Selection is by the metadata KEY, not by the
+  // (possibly-renamed) card type name: a card type rename elsewhere in the
```
samutoljamo, line 79: Verbose comment, will simplify

tools/data-handler/src/mutations/handlers/field-type-enum-remove.ts
```diff
@@ -20,6 +20,15 @@ import type { Card } from '../../interfaces/project-interfaces.js';
 import type { RemoveOperation } from '../../resources/resource-object.js';
 import type { EnumDefinition } from '../../interfaces/resource-interfaces.js';
 
+// An enum value from either recorded shape: a full EnumDefinition object or
```
samutoljamo, line 23: This is a code smell, but deserves its own PR

tools/data-handler/src/mutations/replay/replay.ts
```diff
@@ -0,0 +1,488 @@
...
+    'utf-8',
+  );
+  const lines = content.split('\n');
+  const entries: ConfigurationLogEntry[] = [];
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i].trim();
+    if (line === '') continue;
+    let parsed: unknown;
+    try {
+      parsed = JSON.parse(line);
```
samutoljamo, line 433: Not happy with this yet as now we have two places, which interact with raw config logs

samutoljamo: Probably should use zod here

tools/data-handler/src/mutations/handlers/workflow-remove-state.ts
```diff
@@ -47,24 +48,43 @@ export class WorkflowRemoveStateHandler implements Handler<EditInput> {
...
+  // state); affected cards are then left for final validation to report.
+  private initialState(
+    ctx: MutationContext,
+    workflowName: string,
+  ): string | undefined {
+    const workflow = ctx.project.resources.byType(
+      workflowName,
+      'workflows',
+    )?.data;
+    return workflow?.transitions.find((t) => t.fromState.includes(''))?.toState;
```
CyberJML: Templates.ts is using also || item.fromState.length === 0? Should this also have that condition?

CyberJML (approved):

tools/data-handler/src/mutations/handlers/workflow-remove-state.ts
samutoljamo: Yup

tools/data-handler/src/mutations/handlers/field-type-enum-remove.ts
samutoljamo: https://cyberismo.atlassian.net/browse/INTDEV-1365

PR #1470, opened 2026-06-17
INTDEV-1294 Render Mermaid diagrams in PDF export without a browser

tminkkinen:
## Summary

PDF export rendered Mermaid via the `mmdc` CLI, which requires a global install and a system Chromium (wired up only in the Docker image). Whenever Chromium/`mmdc` was unavailable, the export silently fell back to a code listing instead of a diagram.

This renders Mermaid **in-process, without a browser**:

- A **worker thread** builds the SVG with `svgdom` (no browser) and rasterizes it to PNG with `@resvg/resvg-js`.
- Rendering runs in a worker so the DOM globals Mermaid installs (`window`/`document`/`CSSStyleSheet`) **cannot leak into the main thread** and change how other libraries (e.g. `vega`) detect their environment.
- **PNG, not SVG**: asciidoctor-pdf's prawn-svg renders Mermaid SVGs incorrectly (overlapping text, stray label boxes). A high-DPI PNG (3×, capped at 4000px) is faithful and matches the web view. Note the previous primary path (`[mermaid]` blocks) already embedded PNG too — this just swaps the Chromium rasterizer for resvg.

Both PDF paths now use this: the `[mermaid]` AsciiDoc block preprocessing and the `mermaid` macro's static (PDF) handler. Diagrams that fail to render fall back to a source listing. The web/HTML rendering path is unchanged.

The Chromium/Puppeteer/`mermaid-cli` toolchain is removed from the `Dockerfile`, and the `mmdc` prerequisite is removed from the CLI README.

## Changes

- `tools/data-handler/src/utils/mermaid-worker.ts` (new) — headless Mermaid → SVG → PNG in a worker
- `tools/data-handler/src/utils/mermaid-renderer.ts` — worker-backed batch renderer; PDF preprocessing embeds PNG with graceful fallback
- `tools/data-handler/src/macros/mermaid/index.ts` — macro PDF path renders PNG
- `tools/data-handler/src/types/svgdom.d.ts` (new) — ambient types for `svgdom`
- `tools/data-handler/test/mermaid-renderer.test.ts` (new) — renderer + preprocessing tests
- `tools/data-handler/package.json` — add `mermaid`, `svgdom`, `@resvg/resvg-js`
- `Dockerfile`, `tools/cli/README.md` — drop Chromium/Puppeteer/mermaid-cli

## Testing

- New unit tests: 9/9 pass.
- Full `data-handler` suite: 1536 pass, 1 skipped — no regressions.
- `lint` and `prettier --check` clean.
- End-to-end: exported a card containing a `[mermaid]` block in an environment with **no Chromium and no `mmdc`** — diagram renders correctly in the PDF (with visible labels).

## Notes

- More lightweight than before: removes a ~150–290 MB bundled Chromium; adds ~5 MB of libraries (`svgdom` + resvg's prebuilt binary). `mermaid` itself is roughly a wash (the old `mermaid-cli` pulled it in too).
- `@resvg/resvg-js` is MPL-2.0 (GPL/AGPL-compatible); `mermaid` and `svgdom` are MIT. resvg ships prebuilt binaries for all target platforms incl. linux musl (for the Alpine image).
- Tradeoff: PNG is raster, not vector. Rendered at 3× so it stays crisp.


tools/cli/README.md
```diff
@@ -15,10 +15,6 @@ The Cyberismo CLI is an open-source command-line tool that enables Security-as-C
   - **Windows**: See https://www.ruby-lang.org/en/documentation/installation/ and run `gem install --no-document asciidoctor-pdf rouge`
   - **Verify**: `asciidoctor-pdf -v`
 
-- **Mermaid CLI (optional, for mermaid diagrams in PDF export)**
-  - **All platforms**: `npm install -g @mermaid-js/mermaid-cli`
```
samutoljamo, line 19: Could also update the user facing docs as well

tools/data-handler/src/utils/mermaid-worker.ts
```diff
@@ -0,0 +1,156 @@
...
+
+// Lazily import and configure Mermaid. The dynamic import must happen after the
+// globals above are installed, because Mermaid reads them at module load.
+function getMermaid(): Promise<MermaidApi> {
+  if (!mermaidPromise) {
+    mermaidPromise = import('mermaid').then((module) => {
+      const mermaid = module.default as unknown as MermaidApi;
+      mermaid.initialize({
+        startOnLoad: false,
+        securityLevel: 'loose',
```
samutoljamo, line 94: Is `loose` needed? Preferably as strict as possible without reducing functionality. Shouldn't need much functionality especially since it's a png

tools/data-handler/src/utils/mermaid-worker.ts
```diff
@@ -0,0 +1,156 @@
...
+    });
+  }
+  return mermaidPromise;
+}
+
+function svgSize(svg: string): { width: number; height: number } {
+  const match = svg.match(
+    /viewBox="[\d.eE+-]+\s+[\d.eE+-]+\s+([\d.eE+]+)\s+([\d.eE+]+)"/,
+  );
+  const width = match ? Number.parseFloat(match[1]) : 800;
```
samutoljamo, line 111: Is there any reason to expect that the viewbox would not be defined? If no, this funciton could error. Thought failing whole pdf export for a single failed graph might not be the best behaviour either. Anyway, I think it would good that it would visible somehow(i.e. via logging that without erroring or just error if it truly should never happen)

tminkkinen:
Current solution has better support for diagrams

PR #1473, opened 2026-06-24
INTDEV-1333: Skill resource type with CLI management

tminkkinen:
## Summary

Adds a new **skill** resource type (INTDEV-1333), stored in `.cards/local/skills` and manageable through the CLI like every other resource.

## Changes

- **`SkillResource` + CLI**: `create` / `show` / `list` / `remove`, plus `rename` / `edit` / `delete` rows in the mutation registry.
- **Files**: `query.lp` (optional) + `skill.md` (Markdown), via a `defaultSkill` static template.
- **Metadata**: `name`, `displayName`, `description`, `category`, and a required `relatedTools: string[]` (defaults to `[]`), emitted as `skillRelatedTool(skill, tool)` clingo facts.
- **`cardTreeDirectorySchema.json`**: allow the `skills/` folder under local and modules so `validate` passes.
- **`generateReportContent`**: skip clingo when the query is empty, so content-only skills (and reports) render without running a logic program.
- **Refactor**: replaced the 10-branch type→constructor `if/else` in `ResourceCache.createResourceObject` with a `satisfies`-checked dispatch `Map` (compile-time exhaustiveness).

### DIFF
--- a/tools/assets/src/schema/cardTreeDirectorySchema.json
+++ b/tools/assets/src/schema/cardTreeDirectorySchema.json
@@ -34,6 +34,9 @@
                     "reports": {
                       "$ref": "#/$defs/reportResourceSchema"
                     },
+                    "skills": {
+                      "$ref": "#/$defs/skillResourceSchema"
+                    },
                     "templates": {
                       "$ref": "#/$defs/templatesResourceSchema"
                     },
@@ -380,6 +383,52 @@
         }
       }
     },
+    "skillResourceSchema": {
+      "type": "object",
+      "properties": {
+        "additionalProperties": false,
+        "files": {
+          "$comment": "skills folder can contain skill configuration files (<name>.json), or .gitkeep",
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[A-Za-z0-9-_]+.json$": {
+              "type": "object"
+            },
+            "^.schema$": {
+              "type": "object"
+            },
+            "^.gitkeep$": {
+              "type": "object"
+            }
+          }
+        },
+        "directories": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            ".+": {
+              "type": "object",
+              "properties": {
+                "files": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "skill.md": {
+                      "type": "object"
+                    },
+                    "query.lp": {
+                      "type": "object"
+                    }
+                  },
+                  "required": ["skill.md"]
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "workflowResourceSchema": {
       "type": "object",
       "allOf": [
@@ -572,6 +621,9 @@
                         },
                         "reports": {
                           "$ref": "#/$defs/reportResourceSchema"
+                        },
+                        "skills": {
+                          "$ref": "#/$defs/skillResourceSchema"
                         },
                         "templates": {
                           "$ref": "#/$defs/templatesResourceSchema"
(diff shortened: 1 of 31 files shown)


tools/assets/src/schema/cardTreeDirectorySchema.json
```diff
@@ -380,6 +383,52 @@
...
+        "additionalProperties": false,
+        "files": {
+          "$comment": "skills folder can contain skill configuration files (<name>.json), or .gitkeep",
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[A-Za-z0-9-_]+.json$": {
+              "type": "object"
+            },
+            "^.schema$": {
```
samutoljamo: This is a problem with the rest of the other unchanged schemas, but the dots here are unescaped(in all patterns). In the case of schema, `fschema` passes too, but it should be only `.schema` that passes validation

tools/assets/src/schema/cardTreeDirectorySchema.json
```diff
@@ -380,6 +383,52 @@
         }
       }
     },
+    "skillResourceSchema": {
+      "type": "object",
+      "properties": {
+        "additionalProperties": false,
```
samutoljamo: Shouldn't `additionalProperties` be not under properties but a sibling to it? Again this same issue in the rest of the schema

tools/assets/src/schema/cardTreeDirectorySchema.json
```diff
@@ -380,6 +383,52 @@
...
+            "^.gitkeep$": {
+              "type": "object"
+            }
+          }
+        },
+        "directories": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            ".+": {
```
samutoljamo: Could tighten this to not allow anything, because the folder names match to the identifier jsons. No need to ensure the .json files go one to one with the folders, but it could use the same regex

tools/assets/src/schema/cardTreeDirectorySchema.json
```diff
@@ -380,6 +383,52 @@
...
+              "type": "object",
+              "properties": {
+                "files": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "skill.md": {
```
samutoljamo, line 412: `skill.md` is not `SKILL.md` like typically, but maybe that's good because our skills(at least currently) are not trying to be similar skills as they would be in agents, but they are rather just templated instructions with a custom loader. So maybe that's fine

tools/assets/src/static/defaultSkill/skill.md
```diff
@@ -0,0 +1,3 @@
+# Skill
+
+Describe what this skill does and when an agent should apply it.
```
samutoljamo: Isn't the descriptions main role to explain when the agent should apply it?

tools/data-handler/src/resources/skill-resource.ts
```diff
@@ -0,0 +1,83 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2024
```
samutoljamo: Wrong year. Could vibe code an ESLINT plugin for this :D iirc we looked at the options years ago but there was no plugin that would fully support what we'd want, unless we changed license format so that the embedded license comments wouldn't include the year

tools/assets/src/schema/cardTreeDirectorySchema.json
tminkkinen: Good catch, not needed here. Moving it to a sibling breaks validation though: directory-schema-validator injects path/name onto each directory node, so a node-level additionalProperties: false rejects them (non-allowed additional property: path).

samutoljamo (approved):

PR #1474, opened 2026-06-25
INTDEV-1342 proper module version resolution engine

samutoljamo:
Replaces the best-effort, first-encounter-wins module resolver with a real backtracking constraint solver over the installed closure. Resolution now produces a coherent, replayable set or refuses with an explained conflict, and both `check-updates` (read-only) and the apply path (`import`/`update`/`update-all`) go through the one engine.


tools/data-handler/src/commands/check-updates.ts
```diff
@@ -53,177 +59,98 @@ export class CheckUpdates {
...
-          } satisfies ModuleUpdateStatus;
-        }
-
-        const outcome = await sourceLayer.queryRemote(decl.source, {
-          remoteUrl,
-          range: decl.versionRange,
-        });
+          };
+
+          let plan;
```
samutoljamo, line 101: Note that each availability is checked separately so the whole update list is not a plan that can be executed, because installed modules can have conflicting targets. I.e. if it reports there new versions of module A and B, there is a chance that you cannot upgrade both allthough it is guaranteed that you can update either A or B. "update all" will update both to latest and if they conflict, it chooses one basically by random

CyberJML (approved):

PR #1475, opened 2026-06-26
INTDEV-1334: Dynamic skill discovery via MCP tools

tminkkinen:
## Summary

Adds dynamic skill discovery via two MCP tools (INTDEV-1334). Skills are enabled at runtime by the project's logic programs — `enableSkill(S)` (global) or `enableSkill(S, Card)` (per-card) — and agents discover the currently-enabled ones and fetch their server-rendered content on demand.

## Changes

- **MCP tools** (`tools/mcp/src/tools/index.ts`):
  - `list_skills(category?, cardKey?)` — lightweight summaries of the enabled skills.
  - `get_skill(name, cardKey?)` — full rendered `instructions` for one skill. Not-enabled / needs-cardKey cases return a friendly message rather than an error.
- **Enablement queries** (`tools/assets/src/calculations/queries/`): `enabledSkills.lp` (encodes the global-XOR-per-card rule + `cardKey` filtering) and `globalSkills.lp`, registered in `lpFiles.queries` + `QueryName`.
- **Discovery in the data-handler** (`commands/show.ts`): `Show.listSkills` / `Show.getSkill` resolve enablement, look up metadata, and render `skill.md` server-side via the existing `generateReportContent`. Kept in the command layer so a future backend/UI reuses it; MCP is a thin wrapper.
- **`scope` field**: each listing entry reports `scope: 'global' | 'card'` so an agent knows which skills need a `cardKey`.
- A skill is enabled either globally or per-card (not both); if both, global wins.

### DIFF
--- a/tools/assets/src/calculations/queries/enabledSkills.lp
+++ b/tools/assets/src/calculations/queries/enabledSkills.lp
@@ -1,0 +1,18 @@
+% Resolve which skills are enabled for the current request.
+%
+% A skill is enabled either globally (enableSkill/1) or for specific cards
+% (enableSkill/2) — never both. If both are defined for the same skill, the
+% global enablement wins and the per-card terms are ignored.
+%
+% Each enabled skill name is surfaced via result/1; the caller looks up the
+% skill metadata separately. When a cardKey is given, per-card skills are
+% limited to that card; otherwise every per-card skill is included.
+
+globallyEnabled(S) :- enableSkill(S).
+
+result(S) :- globallyEnabled(S).
+{{#if cardKey}}
+result(S) :- enableSkill(S, {{cardKey}}), not globallyEnabled(S).
+{{else}}
+result(S) :- enableSkill(S, _), not globallyEnabled(S).
+{{/if}}
--- a/tools/assets/src/calculations/queries/enabledSkills.lp
+++ b/tools/assets/src/calculations/queries/enabledSkills.lp
@@ -4,15 +4,16 @@
 % (enableSkill/2) — never both. If both are defined for the same skill, the
 % global enablement wins and the per-card terms are ignored.
 %
-% Each enabled skill name is surfaced via result/1; the caller looks up the
-% skill metadata separately. When a cardKey is given, per-card skills are
-% limited to that card; otherwise every per-card skill is included.
+% Each enabled skill name is surfaced via result/1, together with its scope
+% ("global" or "card") via resultField/4. When a cardKey is given, per-card
+% skills are limited to that card; otherwise every per-card skill is included.
 
-globallyEnabled(S) :- enableSkill(S).
-
-result(S) :- globallyEnabled(S).
+result(S) :- enableSkill(S).
 {{#if cardKey}}
-result(S) :- enableSkill(S, {{cardKey}}), not globallyEnabled(S).
+result(S) :- enableSkill(S, {{cardKey}}).
 {{else}}
-result(S) :- enableSkill(S, _), not globallyEnabled(S).
+result(S) :- enableSkill(S, _).
 {{/if}}
+
+resultField(S, "scope", "global", "shortText") :- enableSkill(S).
+resultField(S, "scope", "card", "shortText") :- result(S), not enableSkill(S).
--- a/tools/data-handler/src/commands/show.ts
+++ b/tools/data-handler/src/commands/show.ts
@@ -660,36 +660,25 @@ export class Show {
    * @param context Calculation context.
    * @returns lightweight summaries of the enabled skills, sorted by name.
    */
+  @read
   public async listSkills(
     options: { cardKey?: string; category?: string } = {},
     context: Context = 'localApp',
   ): Promise<SkillSummary[]> {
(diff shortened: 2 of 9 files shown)


tools/assets/src/calculations/queries/enabledSkills.lp
```diff
@@ -0,0 +1,18 @@
...
+%
+% A skill is enabled either globally (enableSkill/1) or for specific cards
+% (enableSkill/2) — never both. If both are defined for the same skill, the
+% global enablement wins and the per-card terms are ignored.
+%
+% Each enabled skill name is surfaced via result/1; the caller looks up the
+% skill metadata separately. When a cardKey is given, per-card skills are
+% limited to that card; otherwise every per-card skill is included.
+
+globallyEnabled(S) :- enableSkill(S).
```
samutoljamo: I would not use an alias here and simply use `enableSkill(S)` in rules

tools/assets/src/calculations/queries/enabledSkills.lp
```diff
@@ -0,0 +1,18 @@
...
+
+result(S) :- globallyEnabled(S).
+{{#if cardKey}}
+result(S) :- enableSkill(S, {{cardKey}}), not globallyEnabled(S).
```
samutoljamo: No need to use the `not globallyEnabled(S)` here. The atoms in logic programming are unique and that's why it's not possible that there'd be duplicates.
```
result(S) :- enableSkill(S).
{{#if cardKey}}
result(S) :- enableSkill(S, {{cardKey}}).
{{else}}
result(S) :- enableSkill(S, _).
{{/if}}
```
or even
```
result(S) :- enableSkill(S, {{#if cardKey}}{{cardKey}}{{else}}_{{/if}}).
```
Maybe the first one is a bit more readable.

tools/data-handler/src/commands/show.ts
```diff
@@ -626,6 +630,125 @@ export class Show {
     return workflows;
   }
 
+  // Maps a full skill resource to a lightweight discovery summary.
+  private toSkillSummary(
+    skill: Skill,
+    scope: SkillSummary['scope'],
```
samutoljamo, line 636: Yeah so my previous comment is partially wrong, because this uses the scope and the example I gave does not account for that, since all skills are just a list. But the query language supports that. You could give a scope for each result:
```
field(S, "scope", "global", "shortText") :- result(S), enableSkill(S).
field(S, "scope", "card", "shortText") :- result(S), enableSkill(S, _).
```

tools/data-handler/src/commands/show.ts
```diff
@@ -626,6 +630,125 @@ export class Show {
...
+   * @param options.cardKey Optional card to scope card-specific skills to.
+   * @param options.category Optional category filter.
+   * @param context Calculation context.
+   * @returns lightweight summaries of the enabled skills, sorted by name.
+   */
+  public async listSkills(
+    options: { cardKey?: string; category?: string } = {},
+    context: Context = 'localApp',
+  ): Promise<SkillSummary[]> {
+    return this.project.lock.read(async () => {
```
samutoljamo: Use the @read decorator

tools/data-handler/src/commands/show.ts
```diff
@@ -626,6 +630,125 @@ export class Show {
...
+   * @param options.cardKey Card context for per-card template rendering.
+   * @param context Calculation context.
+   * @returns the rendered skill ('ok'), or a marker ('not-enabled' / 'needs-card').
+   */
+  public async getSkill(
+    name: string,
+    options: { cardKey?: string } = {},
+    context: Context = 'localApp',
+  ): Promise<SkillLookupResult> {
+    return this.project.lock.read(async () => {
```
samutoljamo: same here

tools/data-handler/src/commands/show.ts
```diff
@@ -626,6 +630,125 @@ export class Show {
...
+      const scope = globallyEnabled.includes(name) ? 'global' : 'card';
+      // A card-specific skill can't be rendered meaningfully without a cardKey.
+      if (scope === 'card' && !cardKey) {
+        return { status: 'needs-card' };
+      }
+      const resource = this.project.resources.byType(name, 'skills');
+      if (!resource) {
+        return { status: 'not-enabled' };
+      }
+      const skill = resource.show() as Skill;
```
samutoljamo: shouldn't need this cast

tools/data-handler/src/commands/show.ts
```diff
@@ -626,6 +630,125 @@ export class Show {
...
+      const globallyEnabled = new Set(
+        (
+          await this.project.calculationEngine.runQuery('globalSkills', context)
+        ).map((row) => row.key),
+      );
+      const summaries: SkillSummary[] = [];
+      for (const { key } of enabled) {
+        const resource = this.project.resources.byType(key, 'skills');
+        if (!resource) continue;
+        const skill = resource.show() as Skill;
```
samutoljamo: Same here

samutoljamo (approved):
Seems good!

PR #1476, opened 2026-06-30
INTDEV-1336 Static bootstrap skill & install-skills command

tminkkinen:
## Summary

Adds a `cyberismo install-skills [path] [--target claude] [--url <url>]` command that installs the dynamic skill discovery bootstrap skill and MCP server configuration for an AI agent, idempotently. This lets a user point an agent (e.g. Claude Code) at a Cyberismo project and have it discover and apply the project's skills via the MCP server.

## Changes

- **Bootstrap skill** (`tools/assets/src/static/agentSkills/dynamic-skill-discovery/SKILL.md`): a static skill (Claude agent-skill format) instructing agents to call `list_skills` / `get_skill` before acting, shipped via `@cyberismo/assets`.
- **`install-skills` command** (`tools/data-handler/src/commands/install-skills.ts`): for the `claude` target, writes — idempotently — into the install directory:
  - `.claude/skills/dynamic-skill-discovery/SKILL.md` (the bootstrap skill),
  - `.mcp.json` with the `cyberismo` MCP server (merge that preserves any other servers),
  - a marker-delimited pointer block in `CLAUDE.md` (always-loaded, so the agent knows to discover skills).
- Wired via `Cmd.installSkills` (project-not-required, like `create project`) + command-handler dispatch; `InstallSkillsCommandOptions` for `--target` / `--url`.
- **CLI** (`tools/cli/src/index.ts`): the `install-skills` command. Defaults the install dir to the current directory (the directory the AI agent runs from — usually the repo root) and prints the resolved absolute path; `--target` is validated (only `claude` initially).
- Idempotent: re-running yields one `cyberismo` server entry and one `CLAUDE.md` block; existing `.mcp.json` / `CLAUDE.md` content is preserved.

### DIFF
--- a/tools/data-handler/src/commands/install-skills.ts
+++ b/tools/data-handler/src/commands/install-skills.ts
@@ -1,0 +1,130 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation. This program is distributed in the hope that it
+  will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+  of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Affero General Public License for more details.
+  You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { mkdir, readFile } from 'node:fs/promises';
+import { join } from 'node:path';
+
+import { getStaticDirectoryPath } from '@cyberismo/assets';
+
+import {
+  copyDir,
+  pathExists,
+  resolveTilde,
+  writeFileSafe,
+} from '../utils/file-utils.js';
+import { readJsonFile, writeJsonFile } from '../utils/json.js';
+
+import type { InstallSkillsCommandOptions } from '../interfaces/command-options.js';
+
+const SUPPORTED_TARGETS = ['claude'] as const;
+type InstallTarget = (typeof SUPPORTED_TARGETS)[number];
+
+const DEFAULT_MCP_URL = 'http://localhost:3000/mcp';
+
+// Markers delimiting the Cyberismo-managed block in CLAUDE.md, so the block can
+// be replaced idempotently without touching the user's own content.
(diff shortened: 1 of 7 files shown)


tools/data-handler/src/interfaces/command-options.ts
```diff
@@ -120,6 +120,12 @@ export type UpdateModulesCommandOptions = BaseCommandOptions;
 // Options for 'validate' command
 export type ValidateCommandOptions = BaseCommandOptions;
 
+// Options for 'installSkills' command
+export interface InstallSkillsCommandOptions extends BaseCommandOptions {
```
samutoljamo, line 124: It doesn't seem to extend `BaseCommandOptions`, since it does not reuse `projectPath` but gives the install location as argument, which is probably fine, since it would allow doing the setup outside of the project dir

tools/data-handler/src/commands/install-skills.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const mcpPath = join(root, '.mcp.json');
+    const config = pathExists(mcpPath)
+      ? ((await readJsonFile(mcpPath)) as McpConfig)
+      : {};
+    config.mcpServers = {
+      ...config.mcpServers,
+      cyberismo: { type: 'http', url },
+    };
+    await writeJsonFile(mcpPath, config);
```
samutoljamo: The function ignores errors and returns false if it fails, but this passes it to `written` as if it was always successful. At minimum this could not silently push the failure as success, though long term it would be nice to throw these errors to user, since the most likely cause of failure is likely wrong permissions.

samutoljamo (approved):
I think it would be good to merge this as is so that at least Cyberismo can initially try this. But here's a few thoughts for the future:
* Would consider distributing the skill instead using the ways supported by the agents(upside is that we get installing, updating and removing "free" and users control whether the skill is installed for the project or even globally. Though this is per-user, while this PR represents the per project installation so it might be fine that these would live side by side)
* I think hooks could be used instead of injecting the section to CLAUDE.md. Hooks can be also configured if we setup a "marketplace" repo, but it has same difference as the skills: Hooks are then installed by Claude and they are not in the project itself. At least in Claude I think hooks also allow calling the mcp server, so we might not even need the skill in that case
* Also the automated installation assumes cyberismo will be available at port 3000. Maybe we need to re-enable the stdin mcp for local use so it would less likely break. Thought then again that risks multiple processes modifying the same files --> race conditions

PR #1489, opened 2026-07-23
INTDEV-1335: MCP prompts for skill discovery

tminkkinen:
## Summary

Exposes each currently-enabled Cyberismo skill as an **MCP prompt**, so clients can list and inject relevant skill context before an agent starts reasoning.

## What's included

- **`prompts/list`** — returns every enabled skill (`name`, `title`, `description`), recomputed from project state on each call.
- **`prompts/get`** — returns the fully rendered skill (Handlebars applied for templates): a Markdown metadata header (name, category, related tools) followed by the `skill.md` body.
- **Arguments** — card-scoped skills declare an optional `cardKey`; global skills declare none.
- **Graceful responses** — a not-enabled skill returns an explanatory message (not a 404); a card-scoped skill requested without a `cardKey` returns a *needs-card* prompt.

### DIFF
--- a/tools/mcp/src/prompts/index.ts
+++ b/tools/mcp/src/prompts/index.ts
@@ -1,0 +1,117 @@
+/**
+  Cyberismo
+  Copyright © Cyberismo Ltd and contributors 2026
+
+  This program is free software: you can redistribute it and/or modify it under
+  the terms of the GNU Affero General Public License version 3 as published by
+  the Free Software Foundation.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+  details. You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import {
+  GetPromptRequestSchema,
+  ListPromptsRequestSchema,
+  type GetPromptResult,
+  type ListPromptsResult,
+} from '@modelcontextprotocol/sdk/types.js';
+
+import type { ProjectProvider } from '../lib/resolve-project.js';
+
+// A minimal single-message prompt result (used for both rendered skills and the
+// not-enabled / needs-card explanations — never an error, per the spec).
+function textResult(text: string, description?: string): GetPromptResult {
+  return {
+    description,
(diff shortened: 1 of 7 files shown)


tools/mcp/src/prompts/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  );
+
+  server.server.setRequestHandler(
+    GetPromptRequestSchema,
+    async (request): Promise<GetPromptResult> => {
+      const { name } = request.params;
+      const cardKey = request.params.arguments?.cardKey;
+      // Skill names are prefix-qualified (e.g. `decision/skills/foo`); the first
+      // path segment is the project prefix used to resolve the owning project.
+      const prefix = name.split('/')[0];
```
samutoljamo: This won't work if skills come from modules. Provider maps by the projects prefix

samutoljamo: Also it doesn't cost anything to use zod here

samutoljamo (approved):
Tried it and it does work, though it's a bit hard to use. I think we'd need to consider some kind of support from the data model to control which skill is meant to be used automatically / which are these human executed ones. The prompts here are simply prompt templates called by the user and currently calling a card specific skill of the project requires putting the exact card key. The point of that feature is to try providing a more guided approach to making changes in the project without loading all skills, but it seems unintuitive to prevent people from using them. Anyway, this was just future consideration mostly

tools/mcp/src/prompts/index.ts
```diff
@@ -0,0 +1,117 @@
...
+      const commands = provider.get(prefix);
+      if (!commands) {
+        return textResult(`Skill '${name}' is not currently enabled.`);
+      }
+
+      const result = await commands.showCmd.getSkill(name, { cardKey });
+      if (result.status === 'not-enabled') {
```
samutoljamo: This error message is a bit misleading if you put a non-existing key

tools/mcp/src/prompts/index.ts
```diff
@@ -0,0 +1,117 @@
...
+      const cardKey = request.params.arguments?.cardKey;
+      // Skill names are prefix-qualified (e.g. `decision/skills/foo`); the first
+      // path segment is the project prefix used to resolve the owning project.
+      const prefix = name.split('/')[0];
+      const commands = provider.get(prefix);
+      if (!commands) {
+        return textResult(`Skill '${name}' is not currently enabled.`);
+      }
+
+      const result = await commands.showCmd.getSkill(name, { cardKey });
```
samutoljamo: Also `getSkill` does not make sure `cardKey` is a valid key. Might as well fix that in this PR

PR #1491, opened 2026-07-23
INTDEV-1276 Enable override for calculated fields

samutoljamo:
Adds support for an overridable field. This introduces a slightly breaking change, because to be able to derive the value of an overridable field, calculations can not directly specify value of the field and instead they should specify `fieldCalculated` fact, which is derived to a `field` by base.lp. While this is not a requirement for fields that cannot be overridden, it would be confusing that users have many ways to specify the calculated values and thus docs frame it as deprecating using `field` directly for defining a value of an calculated field. The old calculations continue to work and if someone specifies a field value directly for a normal or overrideable field, a rule in base.lp activates that shows up as a notification in the card saying that a field has multiple values. 

Separately we've discussed relaxing validation for metadata. For that reason, I did not want to implement any cascades here, but implemented this as a special case, which will generalize on a follow up PR.


tools/data-handler/src/commands/validate.ts
```diff
@@ -783,12 +783,19 @@ export class Validate {
         continue;
       }
       if (field.isCalculated) {
-        if (card.metadata[field.name] !== undefined) {
+        const value = card.metadata[field.name];
+        const hasStoredValue = value !== undefined && value !== null;
```
samutoljamo, line 798: While this looks initially like a hack, this will be generalized to other fields in a different story

tools/data-handler/src/containers/project.ts
```diff
@@ -1164,7 +1181,19 @@ export class Project extends CardContainer {
     newValue: MetadataContent,
   ) {
     const card = this.findCard(cardKey);
-    if (!card.metadata || card.metadata[changedKey] === newValue) {
+    if (!card.metadata) {
+      return;
+    }
+
+    // Clearing a calculated field removes the stored override entirely, so
```
samutoljamo, line 1194: Same here

henryhaverinen:
I checked out this in module-test and it seems to work great!

I think this would be more usable, if we would show Automatic value and override explicitly only when the field is being edited. When the field is just shown, we could show the actual value (overridden, if exists, otherwise the automatic value). This would mean that the field becomes a lot taller when it is being edited, but that happens already now with all fields. Common use cases for the override would be the owner field, and the information classification field (internal, public etc.), and various review periods,  and at least in these use cases the unused automatic value is interesting only when editing the field.

tools/data-handler/src/utils/clingo-facts.ts
```diff
@@ -267,8 +300,18 @@ export const createCardFacts = async (card: Card, project: Project) => {
         if (value == null) {
           continue;
         }
+        // A stored value on a calculated, non-overridable field is stale: leave
+        // it dormant (emit nothing) so the calculation stays authoritative
+        // instead of colliding with a plain field() fact.
+        if (lockedCalculatedFields.has(field)) {
+          continue;
+        }
```
tminkkinen, line 308: Now disabling field override preserves the already saved override value, right? -> Card will give validation error due to value on non-overridable calculated field. Probably not an issue at all, but minor inconvenience if user disables override and ends up with invalid project.

tools/data-handler/src/commands/edit.ts
```diff
@@ -109,13 +116,46 @@ export class Edit {
     }
 
     // check for editing rights
-    if (this.project.findCard(cardKey)) {
+    const card = this.project.findCard(cardKey);
+    if (card) {
+      this.assertFieldIsEditable(card, changedKey);
```
tminkkinen: Should this guard be used for template cards also?

tminkkinen (approved):

tools/data-handler/src/utils/clingo-facts.ts
samutoljamo: Yup. But with the relaxed validation coming with the rest of stacked PRs, this won't be an error anymore. Later this will likely be part of the "warnings" that can be automatically fixed

tools/data-handler/src/commands/edit.ts
samutoljamo: Yeah. Forgot to think of templates at all. Templates do not have any values for calculated values so we have two options:
* Do not support them at all
* Allow specifying an override allthough a calculated value does not exist

I think the second one is better as then the logic is the same, allthough it's probably a bit niche use-case.

PR #1494, opened 2026-07-30
INTDEV-1305 Add app clean prompt, backend clean endpoint, and docs

samutoljamo:
Add app support for `clean`. This is the simply versions: triggers(remove field from cardtype, disable override on a calculated field and update module) call the dry run to see if there are **any** fields anywhere that could be cleaned and prompts the user if he wants to do that.

Remove field type is added as a trigger in the next PR of the stack.


tools/app/src/components/CleanPrompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function CleanPrompt() {
+  const { t } = useTranslation();
+  const dispatch = useAppDispatch();
+  const findings = useAppSelector((state) => state.cleanPrompt.findings);
+  const [cleaning, setCleaning] = useState(false);
+
+  const confirmClean = async () => {
+    setCleaning(true);
+    try {
+      const { failedCards } = await cleanProject(false);
```
CyberJML: I believe this cleans the project that is in the URL at the time of clicking the confirm? I tested this out with 2 projects and navigated back to other project when the confirm was open and the intended project was not cleaned. Minor problem IMO, quite unlikely that would happen in real world, but might want to take a look at it though

CyberJML (approved):

tools/app/src/components/CleanPrompt.tsx
samutoljamo: I guess we should scope the redux with the project id. Then we would only show it if the project id matches

PR #1502, opened 2026-07-31
INTDEV-1337 Configuration editor: edit and preview for skills

tminkkinen:
## Summary

Makes skills editable in the configuration editor (INTDEV-1337). Skills already existed in the data-handler, CLI and MCP, but the app knew nothing about them — a skill created with `cyberismo create skill` was invisible in the UI. Skills now appear in the configuration tree like any other resource, their metadata and content files are editable, and `skill.md` has an Edit/Preview toggle that renders the unsaved buffer server-side.

## Changes

- **`relatedTools` is now editable** (`data-handler/src/resources/folder-resource.ts`, `skill-resource.ts`): the mutation registry already routed `skills/relatedTools`, but `FolderResource.update` threw on any non-base property, so the field was uneditable through every path including the CLI. Adds a protected `updateAdditionalProperty` hook that subclasses override; `SkillResource` claims `relatedTools` via `handleArray`.
- **`Show.renderSkill`** (`data-handler/src/commands/show.ts`): extracted from `getSkill`, which gates on the `enabledSkills` query — the wrong behaviour while drafting, since a new skill isn't enabled by any logic program yet. `renderSkill` skips the gate and accepts `skillContent` / `skillQuery` overrides so unsaved editor content can be rendered. `getSkill` delegates to it, so there's one renderer.
- **Preview endpoint** (`backend/src/domain/resources/`): `POST /resources/:prefix/skills/:id/preview` → rendered Markdown. Admin-only, since the body carries templates and a logic program the server executes. A broken query returns 400 with the message so the editor shows it in the preview pane rather than a generic 500.
- **Create route and tree** (`backend/src/domain/skills/`, `domain/resources/service.ts`): `POST /skills` mirroring reports, and `skills` added to `resourceTypes` — the file children (`skillContent`, `skillQuery`) come for free from the existing folder-resource handling.
- **Configuration editor** (`app/src/components/config-editors/`): `SkillPreview` + `MarkdownContent` (marked → DOMPurify → html-react-parser), a self-saving `RelatedToolsEditor`, the Edit/Preview toggle in `TextEditor`, and the create modal. `RelatedToolsEditor` bypasses `FieldRow`, whose global Enter shortcut and save-on-blur handler conflict with text entry — matching `EnumValuesEditor`.
- **`cardKey` in preview**: optional, but not cosmetic. A card-scoped skill previewed without one doesn't error — it renders silently wrong, because Handlebars strict mode accepts a present-but-undefined key and clingo accepts the resulting zero-arity atom.

Preview is hidden in static export mode, which rejects POST. Adds one dependency: `marked` (MIT).


henryhaverinen (approved):
I tested the configuration editor, including the preview with a query, and I also tested a query that used {{cardKey}}. Everything worked great.

A small thing, which we don't need to fix in this PR necessarily: when the query uses {{cardKey}}, and before you enter something as the card key, the query template will be yielding invalid ASP, so you get an error message such as `Parsing failed when processing program 'main program' with errors: <string>:3:50-51: error: syntax error, unexpected )`. This goes a way as soon as you provide a card key. This may feel a bit ugly. So as a small enhancement, maybe there could be some text such as "Executing the query failed" always before the error message. If the card key is empty, then the text could be a bit longer: "Executing the query failed. Does it require a card key?"

samutoljamo:
Looking at this, then merging

samutoljamo (approved):

PR #1518, opened 2026-08-05
Spike: migrate toolchain to Vite+ (oxlint/oxfmt) and TypeScript 7

samutoljamo:
Spike, for discussion — not a merge candidate as one lump. Commits are ordered so they can be split into separate PRs.

Replaces ESLint + Prettier with Vite+ (`vp`): oxlint (type-aware), oxfmt, and a cached workspace task runner. Bumps TypeScript 6 → 7, which typescript-eslint was the only thing blocking. Root script names are unchanged, so CI needed no rewiring.

## Commits

Mechanical churn is separated from the decisions, so the three `Mechanical:` commits can be skimmed and the rest read properly.

| | files | commit |
|---|---|---|
| 1 | 21 | **Replace ESLint and Prettier with oxlint and oxfmt** — the actual configuration change |
| 2 | 162 | `Mechanical:` rewrite test imports to `vite-plus/test` |
| 3 | 24 | `Mechanical:` rewrite `eslint-disable` → `oxlint-disable` |
| 4 | 21 | `Mechanical:` reformat with oxfmt |
| 5 | 2 | **Bump TypeScript to 7** — independent; merge candidate on its own |
| 6 | 3 | **Move build/test/dev orchestration to vp run** |
| 7 | 7 | **Make build tasks cacheable by declaring input/output** |
| 8 | 1 | **Persist the Vite+ task cache in CI** |
| 9 | 1 | **Pin Node with `.node-version`** |
| 10 | 2 | **Document the toolchain in AGENTS.md and dependabot** |

Commits 2–4 are reproducible by a single command each; the commit messages say which.

## Results

Re-measured on an idle machine, truly cold (every `dist/` and `.tsbuildinfo` deleted between runs), median of 3.

| | before | after |
|---|---|---|
| lint | 39.8s | **1.9s** |
| format check | 8.6s | **0.43s** |
| build (cold) | 17.6s | **11.9s** |
| build (warm) | ~10.0s | **1.5s** |
| dev prelude (warm) | 0.71s | **0.19s** |

Tests: 2345 pass. The only failures are the two pre-existing `asciidoctor-pdf` env cases.

**Attribution:** the cold-build gain is TypeScript 7, not Vite+ — TS7 under the old `pnpm -r` measures 11.9s, statistically identical to `vp run`. Vite+'s runner pays off on *warm* builds only, and only after commit 4. TS7 needed zero tsconfig changes.

## Caching needed config work

`vp run` won't cache a task that reads a file it also writes, which blocked three. `input`/`output` are only accepted on tasks defined in `vite.config.ts`, so app/assets/backend builds moved out of `package.json`:

- app, assets: tsc reads and rewrites its own `.tsbuildinfo`
- assets: `src/schemas.ts` is generated, then formatted
- assets, backend: `shx cp` probes each destination with `existsSync` before writing, which counts as reading its own output

Verified a cache hit restores `app/dist` byte-complete (130/130 files, including the 255KB `THIRD-PARTY.txt` from the license plugin).

## Runtime management (verified on a real system install)

- `vp` provisions Node — `~/.vite-plus/js_runtime/node/` — selected from `.node-version`. Confirmed serving v22.23.2 inside the repo vs 24.19.0 outside.
- It puts itself first on `PATH` (position 1, ahead of nvm at 10), so it takes precedence without uninstalling nvm.
- The **system install and the devDependency are both needed**: the system install provides Node/pnpm/`vp`, while the `vite-plus` devDependency pins the `vite`/`vitest`/`oxlint`/`oxfmt`/`tsdown` versions. Removing the devDependency would let lint and format versions drift per machine and away from CI.

Net effect on setup: `nvm use` and `corepack enable pnpm` are no longer needed. Per worktree it becomes submodule init + `vp install`.

## Pre-existing bug found (unrelated to Vite+)

`tools/migrations/scripts/generate-version.mjs` writes `tools/assets/src/schema/version.json` — one package mutating another's tracked source. `migrations` depends on `@cyberismo/assets`, so topological order **guarantees** assets builds *before* that rewrite, and a reverse edge would be a cycle. `generateSchemaImports` bakes the value into `SCHEMA_VERSION`, which esbuild bundles.

Latent today (the file is committed and in sync at 5). Failure mode: a PR adding migration N without a committed version bump builds against N-1, so migrations silently don't apply — and it self-heals on a second local build, which makes it hard to spot. It also causes spurious assets rebuilds in `pnpm dev`, since the write lands inside assets' chokidar glob.

Worth fixing regardless of what happens to this PR.

## Open items

- Vite+ is beta (0.2.8). Four `vp migrate` bugs needed hand-fixing: flag order (`--cache` after the task name leaks into the child), fmt running before the import rewrite, dropped promise/react/globals lint config, and `spawn vp ENOENT`.
- Bare `vp test` at the workspace root ignores per-package `vitest.config.ts`. `pnpm test` is correct; noted in AGENTS.md.
- 190 new type-aware warnings, including 68 real floating promises in `tools/app` that ESLint never covered. Non-blocking; needs triage.
- `prettier-plugin-sort-json` has no oxfmt equivalent. No churn, but no future enforcement.
- `pnpm check-licenses` would stop covering the ~360 Rust crates now compiled into the binaries. Probably the one thing to resolve before merging.
- `typeCheck` is off: enabling it surfaces ~60 latent type errors in test dirs that `tsc -p tsconfig.build.json` never covered.
- CI still uses `setup-node` + `pnpm/action-setup`; switching to `voidzero-dev/setup-vp` is a follow-up, unverified on the Windows/macOS matrix.
- app's `tsc -b` is `noEmit: true`, a pure type gate. It could be dropped for speed, but only if `vp check` becomes a required CI gate — today `typeCheck: false` and the type-aware overrides exclude `tools/app`, so dropping it would leave the app with no type checking at all.


samutoljamo:
Closing, since not adopting yet

PR #1519, opened 2026-08-05
INTDEV-1390 Fix inline editors during navigation

CyberJML:


tools/app/src/components/card/CardBody.tsx
```diff
@@ -99,6 +99,18 @@ export const CardBody = forwardRef<CardBodyHandle, CardBodyProps>(
 
     const canEdit = !preview && !!onContentSave && canEditRole;
 
+    // Navigating to a cached card re-renders this component instead of
+    // remounting it, so drop the open editor and its draft — otherwise a save
```
samutoljamo: Might be easier to force a remount by adding `key={card.key}` to components, which consume this component

tools/app/src/lib/hooks/redux.ts
```diff
@@ -45,6 +44,20 @@ export const useAppRouter = (): AppRouter => {
 
   const dialogMsg = t('navigationDialogMsg');
 
+  // useBlocker intercepts ALL navigation — including the browser back/forward
+  // buttons — unlike the safePush/safeBack wrappers which only guard
+  // programmatic calls.
+  const blocker = useBlocker(isEdited);
+  useEffect(() => {
+    if (blocker.state === 'blocked') {
+      if (window.confirm(dialogMsg)) {
```
samutoljamo: A bit ugly, but partially unrelated

tools/app/src/lib/hooks/redux.ts
```diff
@@ -14,9 +14,8 @@ import { useDispatch, useSelector, useStore } from 'react-redux';
 import type { RootState, AppDispatch, AppStore } from '../store';
 import { setIsUpdating } from '../slices/swr';
 import { addNotification } from '../slices/notifications';
-import { useNavigate } from 'react-router';
+import { useBlocker, useNavigate } from 'react-router';
 import { useEffect, useCallback } from 'react';
-import { createFunctionGuard } from './utils';
```
samutoljamo, line 19: If not used anywhere, could remove it

tools/app/src/lib/hooks/redux.ts
```diff
@@ -63,21 +76,10 @@ export const useAppRouter = (): AppRouter => {
...
-          dialogMsg,
-        )
-      : (path: string) => navigate(path, { replace: true }),
-    safeBack: isEdited
-      ? createFunctionGuard(() => navigate(-1), dialogMsg)
-      : () => navigate(-1),
-    safeForward: isEdited
-      ? createFunctionGuard(() => navigate(1), dialogMsg)
-      : () => navigate(1),
+    safePush: (path: string) => navigate(path),
```
samutoljamo: Now that these are just wrappers, maybe just return the router or its methods directly? This hook is still useful since it handles the global `isEdited`, but there's no reason to keep the `safeX` and `X` methods, since it's both confusing and limitng now

samutoljamo (approved):

PR #1522, opened 2026-08-06
INTDEV-1363 Attribute validate errors to the offending file or card

samutoljamo:
Validation produces nicer errors: problematic resource names or macros in content do not throw and stop the validation, but are added to the validation problems list that lists all problems. Also generating the logic programs caused problems for validation as they could crash before validating so they now just log failures and continue. Proper fix for that problem would require a larger refactor, so it was not done here.


henryhaverinen (approved):
Works as expected. I tested with secdevb module

PR #1528, opened 2026-08-06
INTDEV-1293 Add urlPath calculated field for cards

samutoljamo:
Adds a built-in calculated field urlPath. It's a calculated field in the sense that it does not use `fieldCalculated`, but there is no benefit to it being an "official" calculated field.


henryhaverinen (approved):
Looks good to me. I didn't test in a graph, but the included test does exactly what we need this to do

PR #1529, opened 2026-08-06
INTDEV-1344 Refetch card data when presence changes

samutoljamo:
Leaving edit mode causes a refresh for other viewers. Additionally added some mock users so that it's possible to test this. It still works via a cookie, so a separate profile / browser / incognito required. THe role readers doesn't have a precense so currently so won't get the update. Any reason why we couldn't add presence for readers


tools/app/src/lib/api/presence.ts
```diff
@@ -84,3 +86,42 @@ export function usePresence(
...
+  projectPrefix?: string,
+): void {
+  const { user } = useUser();
+  const othersEditing = presence.some(
+    (entry) => entry.userId !== user?.id && entry.mode === 'editing',
+  );
+  const othersWereEditingRef = useRef(false);
+
+  useEffect(() => {
+    if (othersWereEditingRef.current && !othersEditing && cardKey) {
```
CyberJML, line 120: I would change this to mutate when somebody stops editing, not when everybody stops editing. This would make it better for the viewer. For example, A is actively editing, but B is away from the computer with edit mode open, this would lead to case where C would not see the changes until B is back, unless they refresh the page

CyberJML (approved):

PR #1537, opened 2026-08-13
Update the documentation about module updates and versioning

henryhaverinen:


samutoljamo (approved):

docs/.cards/local/cardsConfig.json
```diff
@@ -1,12 +1,14 @@
 {
-    "schemaVersion": 4,
+    "schemaVersion": 5,
```
samutoljamo, line 2: Docs pipeline has `env: CLI_VERSION: 0.0.26`, it should be changed to `1.1.0`. It's in `.github/workflows/docs.yml`

henryhaverinen: I fixed this

PR #1538, opened 2026-08-14
INTDEV-1399 Generate the calculated field bridge per field name

samutoljamo:
Optimize by removing the generic rule from `base.lp`, which caused performance problems


henryhaverinen (approved):
This fixes the performance issue and calculated fields with override still work functionally

CyberJML (approved):

PR #1541, opened 2026-08-18
INTDEV-1403 Add better SaaS git support

CyberJML:


tools/data-handler/src/utils/git-manager.ts
```diff
@@ -172,7 +180,79 @@ export class GitManager {
...
+   *
+   * Everything here is a precondition rather than an error: a project with no
+   * remote, no branch, or nothing new to say is a normal state, not a failure.
+   * A remote that has moved on independently is reported as `diverged` and
+   * left alone — resolving that is a person's job, never a force push.
+   *
+   * @param remoteName Remote to push to.
+   * @returns what happened, for the caller to log or act on.
+   */
+  async pushIfAhead(
```
samutoljamo: This function is quite complex and it also introduces a few races so there is no full guarantee it does not fail. While that is unlikely, I'd consider just using `push` that already exists and handling the exceptions if you need to report them. `push` could be extended to provide the reason. I believe this would cover the error cases and be much simpler

tools/data-handler/src/containers/project.ts
```diff
@@ -145,6 +148,10 @@ export class Project extends CardContainer {
       await this.calculationEngine.generate();
     });
 
+    if (this.options.autopush) {
+      this.autoPush = new AutoPush(this.gitManager);
```
samutoljamo: Would be better to not gate behaviour by not instantiating the class.

tools/data-handler/src/utils/auto-push.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Scheduling only — every decision about whether a push is possible lives in
+ * {@link GitManager.pushIfAhead}. There is one of these per project, so none
+ * of the state needs keying.
+ *
+ * {@link schedule} is deliberately synchronous: it is called from an
+ * after-write hook that still holds the write lock, and a push is network I/O.
+ * The timer moves the work outside the lock, so a slow remote cannot make a
+ * card save wait on it.
+ */
+export class AutoPush {
```
samutoljamo: I'd consider renaming to `GitSync` or `AutoSync` or something like that. I'd assume the behaviour is not separated when it's implemented

tools/data-handler/src/utils/auto-push.ts
```diff
@@ -0,0 +1,162 @@
...
+  constructor(
+    private readonly git: GitManager,
+    private readonly debounceMs: number = DEBOUNCE_MS,
+  ) {}
+
+  /**
+   * Note that there is something to push. Returns immediately; the push runs
+   * later, off the write path.
+   */
+  schedule(): void {
```
samutoljamo: Most of the complexities come from the debounce here, do we really need it? What if this used its own lock and just do the git push inside it? Only negative side is that we'd do a couple of extra pushes at worst, no? Could extend the RWLock to limit the queue: 
Try to acquire lock:
* If gets lock, do the push
* If someone has the lock but nobody is waiting: Go to queue
* If someone has the lock and at least 1 is waiting: throw error or just dont run the function

Now if someone pushes 200 changes that are quick to do, they are guaranteed to pushed, but don't cause 200 pushes

tools/data-handler/src/utils/auto-push.ts
```diff
@@ -0,0 +1,162 @@
...
+    );
+  }
+
+  /**
+   * Routine pushes are silent: with autocommit there is one per card edit.
+   * Only a change of state is worth a line, so a failure is reported once and
+   * the recovery after it is visible.
+   */
+  private report(outcome: PushOutcome): void {
+    switch (outcome.status) {
```
samutoljamo: Bit confused why the outcome is received and parsed here to create the messages. We could just add the logs to git manager. Also I wouldn't mind removing the status check filtering, these shouldn't very spammy and you can use lower log levels if needed

tools/data-handler/src/utils/egress.ts
```diff
@@ -0,0 +1,60 @@
...
+  You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import type { Dispatcher } from 'undici';
+
+/**
+ * Outbound HTTP for deployments with no direct route to the internet.
+ *
+ * Applied per call, not via `setGlobalDispatcher`, and read from a
```
samutoljamo: Quite heavy comments in general, could trim them

tools/data-handler/src/utils/auto-push.ts
samutoljamo: Asked this from claude and a simpler idea came that is very close to the current implementation:
```ts
private running = false;
private pending = false;

schedule(): void {
  if (this.running) { this.pending = true; return; }
  void this.loop();
}

private async loop(): Promise<void> {
  this.running = true;
  try {
    do {
      this.pending = false;
      await this.runOnce(); 
    } while (this.pending);
  } finally {
    this.running = false;
  }
}
```

tools/data-handler/src/containers/project.ts
```diff
@@ -140,6 +143,10 @@ export class Project extends CardContainer {
       await this.calculationEngine.generate();
     });
 
+    this.gitSync = new GitSync(this.gitManager, {
+      enabled: this.options.autopush ?? false,
```
samutoljamo: Prefer guarding the function calls at the caller instead of having to set a flag to `GitSync`. `this.options` is available anyway

tools/data-handler/src/utils/git-manager.ts
```diff
@@ -172,7 +175,51 @@ export class GitManager {
...
+    }
+
+    try {
+      // `-u` covers the first push too; re-setting an upstream is a no-op.
+      await this.git.push(['-u', remoteName, branch]);
+      // Debug, not info: with autocommit there is a push per card edit.
+      this.logger.debug({ branch, remoteName }, 'Pushed to remote');
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      if (GitManager.REJECTED_BY_REMOTE.test(message)) {
```
samutoljamo: I'd still remove this whole func. I'd say it's a configuration error if there is no remote. And this error wrapper saves a few git push calls, but I think it's just as informative to just let the git errors flow to logs and if there's a few extra failures during retries it's fine. Then we can remove all git-try-push tests too.

tools/data-handler/src/utils/git-sync.ts
```diff
@@ -0,0 +1,88 @@
...
+  License along with this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import { getChildLogger } from './log-utils.js';
+import type { GitManager } from './git-manager.js';
+
+const MAX_RETRIES = 2;
+const RETRY_DELAY_MS = 5000;
+
+function sleep(ms: number): Promise<void> {
```
samutoljamo: I'd put this to utils, there are already a few places that use this

tools/data-handler/src/utils/git-manager.ts
```diff
@@ -12,18 +12,17 @@
...
 export class GitManager {
-  private git: SimpleGit;
+  private git: ReturnType<typeof createGit>;
   private logger = getChildLogger({ module: 'GitManager' });
 
   constructor(projectPath: string) {
-    this.git = simpleGit(projectPath, {
+    this.git = createGit({
+      baseDir: projectPath,
       config: ['user.name=Cyberismo Bot', 'user.email=bot@cyberismo.com'],
```
samutoljamo, line 27: Could add the timeout

tools/data-handler/src/containers/project.ts
```diff
@@ -948,6 +956,7 @@ export class Project extends CardContainer {
     if (this.options.autocommit) {
       await this.gitManager.initialize(getCommitContext().author);
     }
+    void this.gitSync.push();
```
samutoljamo: CLI doesn't prevent autocommit from being false when autopush is true. At least this push is run even if there is no autocommit mode enabled

tools/data-handler/src/utils/git-sync.ts
```diff
@@ -0,0 +1,88 @@
...
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms).unref?.();
```
samutoljamo: If we unref here, the retry logic doesn't work for CLI

tools/data-handler/test/git-sync.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  it('pushes again for work that arrived mid-push', async () => {
+    // The second edit must not be lost just because a push was running.
+    const auto = new GitSync(managerPushing(), { enabled: true });
+    const first = auto.push();
+    void auto.push();
+    await first;
+    await settle();
+    expect(true).toBe(true);
```
samutoljamo: ?

tools/data-handler/test/git-sync.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    const sync = new GitSync(git, { enabled: true });
+
+    for (let i = 0; i < 200; i += 1) void sync.push();
+    await settle();
+
+    expect(git.tryPush).toHaveBeenCalled();
+    expect(git.tryPush.mock.calls.length).toBeLessThanOrEqual(2);
+  });
+
+  it('keeps pushing while edits keep arriving', async () => {
```
samutoljamo: It doesn't test what it claims to test. I'd just remove it

tools/data-handler/test/git-sync.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    // The second edit must not be lost just because a push was running.
+    const auto = new GitSync(managerPushing());
+    const first = auto.push();
+    void auto.push();
+    await first;
+    await settle();
+    expect(true).toBe(true);
+  });
+
+  it('pushes without waiting for a commit', async () => {
```
samutoljamo: Why would wait for a commit?

tools/data-handler/test/git-sync.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    const git = managerPushing();
+    await new GitSync(git).push();
+    expect(git.tryPush).toHaveBeenCalledTimes(1);
+  });
+
+  it('retries a failed push, then gives up quietly', async () => {
```
samutoljamo: Doesn't test what it claims to test

tools/data-handler/test/command-create-clone.test.ts
```diff
@@ -3,19 +3,12 @@ import { mkdtemp, mkdir, rm } from 'node:fs/promises';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 
-const cloneFromGitServiceMock = vi.fn();
-const isGitServiceEnabledMock = vi.fn();
 const resolveGitServiceClonePathMock = vi.fn();
```
samutoljamo: git service mock

tools/backend/src/utils.ts
```diff
@@ -61,3 +61,22 @@ function testPort(port: number) {
     }, 2000);
   });
 }
+
+/**
+ * Git write behaviour, from the environment.
+ *
+ * Read in one place because the server start-up path and the test-reset route
+ * both build CommandManagers, and a difference between them would mean a reset
+ * silently ran with different behaviour than the server it replaced.
```
samutoljamo: I'd leave these kind of comments out, which document behaviour of distant module. Especially agents tend to trust the comments and thus makes them less effective.

tools/backend/src/utils.ts
```diff
@@ -61,3 +61,22 @@ function testPort(port: number) {
...
+ */
+export function gitOptionsFromEnv(): {
+  autocommit: boolean;
+  autopush: boolean;
+} {
+  return {
+    autocommit: process.env.CYBERISMO_AUTOCOMMIT === 'true',
+    // Autopush requires autocommit; Project normalises that, so this reports
```
samutoljamo: This too

tools/data-handler/src/utils/git-sync.ts
```diff
@@ -0,0 +1,95 @@
...
+        await this.pushWithRetries();
+      } while (this.requestedAgain);
+    } finally {
+      this.running = false;
+    }
+  }
+
+  private async pushWithRetries(): Promise<void> {
+    // A project without a remote is a supported state, not a misconfiguration:
+    // the URL can be set later from the project settings, while a silo turns
```
samutoljamo: I'd remove

samutoljamo (approved):

PR #1567, opened 2026-09-01
INTDEV-733: Consolidate card storage into a CardTree

samutoljamo:
Adds the new card storage, but does not use it directly yet, instead keeps the old system that uses the new card tree. The idea here was that this would create a bit smaller PR(though still large) and the diff would be more readable. This is why most methods now become wrappers. Implementation side simplifies still, the added changes are mostly test(and of course card tree, which has many previous project methods and card cache functionality index etc.


tools/data-handler/test/creation-side-effects.test.ts
```diff
@@ -40,3 +42,53 @@ describe('creation side effects', () => {
...
+
+    expect(created.length).toBeGreaterThan(1);
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect([...spy.mock.calls[0][0]].sort()).toEqual(
+      created.map((card) => card.key).sort(),
+    );
+    spy.mockRestore();
+  });
+
+  it('runs the creation query once for a batch of added template cards', async () => {
```
samutoljamo, line 79: This does not make sense to me, but it is a bug in main too, so I'll create a separate ticket about it

CyberJML (approved):

PR #1569, opened 2026-09-02
INTDEV-733: One card tree per container

samutoljamo:
This brings what the refactor promised: Now each template resource will have a card tree and project will have its own card tree. They are also used directly so this means that the project gets somewhat smaller. Gets rid of the inheritance between project/template. Template still depends on project.

Not fully happy with this, but I think the direction / architecture is better and certain things(like getting rid of project as a mediator of trees) requires api refactor which should be separate. Confident that the code itself works though


tools/data-handler/src/containers/project/card-tree.ts
```diff
@@ -622,80 +678,118 @@ export class CardTree {
...
-    for (const cardKey of this.locationIndex.get(location) ?? []) {
-      const card = this.cards.get(cardKey);
-      if (card) {
-        cards.push(card);
+  /**
+   * Takes a card and its descendants out of the tree, without touching the
+   * filesystem. The caller hands the result to the destination tree's graft().
+   * @returns the subtree's cards, parents before children.
+   */
+  public uproot(cardKey: string): StoredCard[] {
```
samutoljamo, line 748: Move is still out of card tree, follow up PR fixes this

CyberJML (approved):

PR #1574, opened 2026-09-03
INTDEV-1445 Fix Undici usage after update

CyberJML:


tools/data-handler/src/commands/fetch.ts
```diff
@@ -149,7 +163,7 @@ export class Fetch {
         throw new Error('JSON content too large after parsing');
       }
 
-      return json;
+      return json as HubDocument;
```
samutoljamo: DH has zod --> prefer it over the manual `as`

tools/data-handler/test/egress.test.ts
```diff
@@ -56,12 +59,25 @@ describe('egressFetch', () => {
...
-    const spy = fetchSpy();
-    vi.stubGlobal('fetch', spy);
+    fetchMock.mockResolvedValue({ ok: true });
+
+    await egressFetch('https://hub.example/moduleList.json');
+
+    expect(fetchMock.mock.calls[0][1]?.dispatcher).toBeUndefined();
+  });
+
+  it('does not fall back to the global fetch', async () => {
```
samutoljamo: Assuming that the test works, it very much tests the very specific scenario with lots of mocks, a test that would actually catch a failing proxy request would be better, but not sure how easy would that be

samutoljamo (approved):

PR #1575, opened 2026-09-03
INTDEV-1344 Refetch card views on card-updated presence events

samutoljamo:
Before spending more time on this, I'll show a demo next monday of this.


samutoljamo:
I close this, since the stack replaces this

PR #1576, opened 2026-09-03
INTDEV-733: Pull card facts before a solve instead of pushing after a write

samutoljamo:
The calculation engine stops being told about card writes and starts asking for them. Instead of every write pushing its facts into clingo as a side effect, each write records what changed and the next solve pulls it — so a write that is never queried costs nothing, and every path into clingo is serialised through one `pull()`.

What the engine has not yet projected lives in one `FactLog`: the cards whose facts need rebuilding, the cards whose facts must go, and whether the resource programs are stale. It replaces fact tracking that ran in three layers — per-tree sets, a merge in `Project`, a drain callback on the engine — and because the log outlives any individual tree, a tree that is dropped no longer takes its pending removals with it.

Fixes a bug the pull model would otherwise have introduced: an invalidation arriving while a rebuild was running was cleared by that rebuild's own completion, leaving the engine serving stale programs until something else invalidated them. The log carries a revision, and a rebuild only marks itself clean if nothing invalidated it while it ran.

`generate()` keeps its name and its callers, but is now an invalidation followed by a pull rather than a rebuild of its own — one entry point into clingo, not two.

### DIFF
--- a/tools/data-handler/src/command-manager.ts
+++ b/tools/data-handler/src/command-manager.ts
@@ -143,7 +143,6 @@ export class CommandManager {
     this.project.resources.changedModules();
     await this.project.populateCaches();
     await this.project.initializeGit();
-    await this.project.calculationEngine.generate();
   }
 
   /**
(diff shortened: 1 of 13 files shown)


samutoljamo:
This would be a nice addition, but it's independent from the current refactor and I think it's colliding with a few upcoming changes so I'll drop it from this. Generate calls, especially in the long running app, are not a huge problem, the biggest payoff is for CLI startup / if Clingo ever was a separate service

PR #1583, opened 2026-09-10
Render card AsciiDoc in the browser instead of the backend

samutoljamo:
Improves safety + performance. Taking the syncronous convert operation from backend frees up pressure from it and the secure mode prevents access to file system.

### DIFF
--- a/tools/app/src/lib/hooks/useAdocHtml.ts
+++ b/tools/app/src/lib/hooks/useAdocHtml.ts
@@ -1,0 +1,54 @@
+/**
+    Cyberismo
+    Copyright © Cyberismo Ltd and contributors 2026
+
+    This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License version 3 as published by the Free Software Foundation.
+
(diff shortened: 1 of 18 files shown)


tools/app/src/lib/hooks/useAdocHtml.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const token = `${cardKey}\n${adoc}`;
+  const cached = cachedAdocHtml(adoc, cardKey);
+
+  useEffect(() => {
+    if (cached !== undefined) {
+      return;
+    }
+    let current = true;
+    adocToHtml(adoc, cardKey).then((html) => {
```
CyberJML, line 40: Should there be .catch also?

CyberJML (approved):

tools/app/src/lib/hooks/useAdocHtml.ts
samutoljamo: Not needed, it will just get logged to the console

PR #1586, opened 2026-09-15
INTDEV-1344 Replace per-card presence with one project event stream

samutoljamo:
Stack 2/3, re-derived from #1575. Base: ship-1.

## Summary
- One SSE stream per tab: `GET /api/projects/:prefix/events` emits `ready` and `presence.updated`; readers can subscribe. Static generation skips it via `disableSSG()`.
- Presence is a lease: `PUT /api/projects/:prefix/events/presence` with a monotonic `sequence`, renewed every 30 s, expires after 90 s. Gated by `APP_PRESENCE_ENABLED`.
- `ProjectEventsProvider` owns the connection; `usePresence` reports through it. Old per-card `/cards/:key/presence` route removed.
- README carries the contract.

## Test plan
- Backend `test/events.test.ts`: 8 tests (HTTP stream, lease bookkeeping, flag, SSG skip, disposal).
- App `__tests__/projectEvents.test.tsx`: 6 tests (single connection, snapshots, project switch, static mode, bfcache, no project-wide invalidation).


samutoljamo:
So instead of relying on the SSE stream itself for the precense, this complicates it by having the client itself have to broadcast its state and thus having to implement the lease mechanism. I think this complexity is fine, as it enables us to make SSE stream long living, it can be used for any changes we want to push to the clients. No need to open separate SSE streams for each card.

CyberJML (approved):

PR #1594, opened 2026-09-17
INTDEV-1344 Bound stream lifetime and rotate connections without flicker

samutoljamo:
Authorization was evaluated once when a stream opened and never again, so logout, token revocation, a role downgrade and offboarding had no effect on an open stream — it lived until the tab closed, which could be days. Streams now close at `min(exp - 30s, 30min jittered)` and the reconnect re-runs the auth middleware, so revocation lag becomes the token lifetime the IdP already configures; jitter only ever shortens, so it cannot push a stream past `exp`, and a token already inside the margin rotates at expiry rather than looping. Rotation is invisible to others because a retiring connection's presence entry outlives its socket and the reconnect re-declares the same card, leaving the deduped list unchanged — and `presence.updated` is now suppressed per recipient rather than per card, so a reconnecting or second tab still receives its own current list.


samutoljamo:
Made `exp` required @CyberJML . We can assume and I don't a non expiring access token would ever be required

tools/backend/src/domain/events/project-events.ts
```diff
@@ -29,11 +29,44 @@ export interface Capabilities {
   canSeePresence: boolean;
 }
 
+/** What one connection's client is sent through, and what was last sent to
+ * it. Retiring turns send/close/terminate into no-ops together, so a stale
+ * reference can never reach a socket the connection has given up. */
+class Delivery {
```
samutoljamo, line 36: ship 8 removes this wrapper

PR #1595, opened 2026-09-17
INTDEV-1344 Cap per-user connections, detect stale streams, bound resources

samutoljamo:
A per-user cap of 30 concurrent streams evicts the oldest non-`editing` connection with a `capped` terminal event, reusing ship-7's terminate-then-close shape; the client honours it with a `terminal` flag that also stops `pageshow` from resurrecting a stream closed after a terminal auth failure. Presence changes are throttled to one per second per connection on a trailing edge — a change inside the window is applied when it elapses rather than discarded — and outstanding SSE writes are bounded per connection, dropping only `presence.updated` and never marking a dropped one as delivered. Ship-6's `hb` event now drives a client watchdog that reconnects and surfaces a "may be showing stale content" banner; the copy deliberately promises nothing, since MCP, CLI and direct git writes emit no `card.updated` at all.


samutoljamo:
Ended up removing the limit, I think it complicates this more than it serves us

